PR #586, opened 2018-02-18
Fix Issue 527: MultivaluedHashMap is not deserializable

### LINKED ISSUE #527: MultivaluedHashMap is not deserializable
Despite the MultivaluedHashMap implementing java.io.Serializable and having a public no arg constructor, it is impossible to deserialize a previously serialized instance of MultivaluedHashMap.

Attempting to do so results in an InvalidClassException being thrown.

A code snippet to reproduce the bug is available here:
[http://stackoverflow.com/questions/34732363/unable-to-deserialize-multivaluedhashmapstring-string-in-java](http://stackoverflow.com/questions/34732363/unable-to-deserialize-multivaluedhashmapstring-string-in-java)
#### Affected Versions
[2.0.1]

NiklasMehner:
- Added Serializable interface to AbstractMultivaluedMap
- Renamed MultivaluedMapTest -> MultivaluedHashMapTest to match tested class

This should fix #527

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/AbstractMultivaluedMap.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/AbstractMultivaluedMap.java
@@ -16,6 +16,7 @@
 
 package javax.ws.rs.core;
 
+import java.io.Serializable;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
@@ -30,7 +31,7 @@ import java.util.Set;
  * @param <V> the type of mapped values.
  * @author Marek Potociar
  */
-public abstract class AbstractMultivaluedMap<K, V> implements MultivaluedMap<K, V> {
+public abstract class AbstractMultivaluedMap<K, V> implements MultivaluedMap<K, V>, Serializable {
 
     /**
      * Backing store for the [key, multi-value] pairs.
--- a/jaxrs-api/src/test/java/javax/ws/rs/core/MultivaluedHashMapTest.java
+++ b/jaxrs-api/src/test/java/javax/ws/rs/core/MultivaluedHashMapTest.java
@@ -22,10 +22,13 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.io.*;
+
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.assertFalse;
 
-public class MultivaluedMapTest {
+public class MultivaluedHashMapTest {
 
     @Before
     public void setUp() throws Exception {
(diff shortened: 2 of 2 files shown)


jaxrs-api/src/main/java/javax/ws/rs/core/AbstractMultivaluedMap.java
```diff
@@ -30,7 +31,7 @@
  * @param <V> the type of mapped values.
  * @author Marek Potociar
  */
-public abstract class AbstractMultivaluedMap<K, V> implements MultivaluedMap<K, V> {
+public abstract class AbstractMultivaluedMap<K, V> implements MultivaluedMap<K, V>, Serializable {
```
mkarg: @chkal I am uncertain whether adding the `Serializable` _marker_ interface could imply any backwards compatibility issues with existing third-party code: _In theory_ somebody could have extended this abstract class so his code could fail now due to missing implementation of the methods of `Serializable`, on the other hand `Serializable` is a pure marker interface with *no* methods at all. I personally think this is too risky for `2.1.1-SNAPSHOT` and we should move it to `2.2-SNAPSHOT` instead? What do you think?

chkal: That's a really good question. I agree that adding it to `2.1.1-SNAPSHOT` could be too risky. But I'm not very familiar with the technical details of serialization to give a qualified answer here.

spericas: Yes, this is 2.2-SNAPSHOT.

NiklasMehner: I have updated the PR to 2.2-SNAPSHOT.

mkarg:
@chkal Right, you are not a committer, but as Santiago did not explicitly vote minus one when asking for 2.2-SNAPSHOT target, I assume his vote is at least zero. Also you are a committer to MVC API, which is so tightly related, that I think we can count you. ;-)

spericas:
@mkarg I actually voted +1 but was a bit shy about it :)

mkarg:
@spericas No need to be shy. ;-) I looked at the Github review results and yours said "commented" (i. e. "zero"), so I assumed this is your vote. Question: Do you want us to write comments saying "+1" or do you want us using the Github Review tool? I always used the tool so far, maybe this is wrong?

chkal:
I actually really like the GitHub code review tooling. Works great for the MVC spec. It makes votes very explicit. But maybe this should be discussed on the mailing list?

mkarg:
Good idea. I will start a thread.

spericas:
@mkarg Sorry I was inconsistent between using the tool and adding comments. Let's use the tool as suggested.

Issue #587, opened 2018-02-19
[INFO] "There are 984 checkstyle errors"


mkarg:
```
[INFO] --- maven-checkstyle-plugin:2.12.1:checkstyle (default) @ javax.ws.rs-api ---
C:\Users\Markus\git\jaxrs-api\jaxrs-api\src\main\java\javax\ws\rs\client\FactoryFinder.java:48:76: unexpected token: )
C:\Users\Markus\git\jaxrs-api\jaxrs-api\src\main\java\javax\ws\rs\core\Response.java:1230:9: unexpected token: public
C:\Users\Markus\git\jaxrs-api\jaxrs-api\src\main\java\javax\ws\rs\ext\FactoryFinder.java:48:76: unexpected token: )
C:\Users\Markus\git\jaxrs-api\jaxrs-api\src\main\java\javax\ws\rs\ext\MessageBodyWriter.java:83:5: unexpected token: public
C:\Users\Markus\git\jaxrs-api\jaxrs-api\src\main\java\javax\ws\rs\sse\FactoryFinder.java:48:76: unexpected token: )
C:\Users\Markus\git\jaxrs-api\jaxrs-api\src\main\java\javax\ws\rs\sse\Sse.java:47:5: expecting RCURLY, found 'default'
C:\Users\Markus\git\jaxrs-api\jaxrs-api\src\main\java\javax\ws\rs\sse\SseEventSource.java:239:5: unexpected token: default
[INFO]
[INFO] There are 984 checkstyle errors.
```

mkarg:
It would be great if somebody would find the time to fix this **984** checkstyle errors! :-)

chkal:
Maybe we could just adjust the checkstyle config. Usually the default configuration is a bit "strict". :laughing:

spericas:
+1 on adjusting the config

mkarg:
@chkal Any proposal for a _wise_ config?

chkal:
My guess is that there are some rules which are violated VERY often which causes this high number of violations. I'll try to find some time in the next days to have a deeper look at it.

Can you assign this issue to me although I'm not a committer? Not sure if GitHub allows this.

spericas:
@chkal We should nominate you as a committer.

chkal:
@spericas: Thanks a lot! I would be happy to join!

mkarg:
@chkal I tried to assign the issue to you but Githab does not accept that.

[...]

PR #588, opened 2018-02-23
Updated Checkstyle and improved configuration (#587)

chkal:
This pull requests addresses the problem described in #587. Merging this pull request will reduce the number of Checkstyle violations from 984 to 88.

The pull request contains the following changes:

* Updated of the `maven-checkstyle-plugin` to the latest version
* Ignore `module-info.java`, because it breaks the analysis for some reason.
* Increased the maximum allows line length to 160.
* Don't enforce whitespaces around `<` and `>` tokens, as this would fail even for simple generics usage like `List<String>`.
* Deleted `RedundantThrows` check which was removed from Checkstyle (see https://github.com/checkstyle/checkstyle/issues/473).
* Don't enforce all method parameters to be `final`.

Please let me know if you have any objections regarding these changes.

There are 3 other error categories which cause about 90% of the remaining violations. But I think we should discuss them in #587 to decide whether to remove the check or to adjust the source code.

### DIFF
--- a/etc/config/checkstyle.xml
+++ b/etc/config/checkstyle.xml
@@ -135,7 +135,7 @@
         <!-- Checks for Size Violations.                    -->
         <!-- See http://checkstyle.sf.net/config_sizes.html -->
         <module name="LineLength">
-            <property name="max" value="120"/>
+            <property name="max" value="160"/>
             <property name="ignorePattern" value="@version|@see|@todo|TODO"/>
         </module>
         <module name="MethodLength"/>
@@ -163,7 +163,7 @@
                 LITERAL_TRY, LITERAL_WHILE, LOR, LT, MINUS, MINUS_ASSIGN, MOD,
                 MOD_ASSIGN, NOT_EQUAL, PLUS, PLUS_ASSIGN, QUESTION, SL,
                 SLIST, SL_ASSIGN, SR, SR_ASSIGN, STAR, STAR_ASSIGN,
-                TYPE_EXTENSION_AND, GENERIC_START, GENERIC_END" /> <!-- RCURLY removed -->
+                TYPE_EXTENSION_AND" /> <!-- RCURLY removed -->
         </module>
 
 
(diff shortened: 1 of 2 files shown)


chkal:
Hey @mkarg. Thanks for the review. Sure, every comment is welcome. And I must confess that I also had a hard time deciding which checks to disable and which checks to keep. And many of those checks are about personal preference.

Basically my assumption while creating this patch was:

> If there is a huge number of violations of the same type, then most likely these  rules weren't considered important back then.

Regarding your comments:
* I agree that 160 feels very huge. I would be fine with reducing it. But there are quite some lines which may be difficult to format correctly with 140 or 120 character line width. See [this example](https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/core/Response.java#L1296).
* I could split the changes into multiple commits if you want. No problem.
* I also love `final`. But introducing it for every method/constructor parameter is a huge change. And actually not binary compatible, or am I mistaken.

However, I could also send a pull request just for the Checkstyle update as a first step if you prefer it this way. Then we could discuss further changes separately.

chkal:
I forgot to mention one thing. The remaining 88 issues are most likely more interesting to address as they point to real issues. Like missing `@throws` or malformed HTML (see #589).

mkarg:
@chkal Maybe it would be best if we open a thread on the actually wanted formatting style in the dev mailing list before we further invest time into the PR itself? I don't want to decide things that other commiter might see different. About `final`, I found some helping hands (a list of them) waiting for simple work to start with. I could ask them to hack it in for us?

chkal:
Agreed. But perhaps we should at least update Checkstyle to the latest version? I think this should be save and it would be a good starting point for discussing the actual rule set.

mkarg:
upgrading checkstyle sounds good to me

chkal:
Ok, I'll adjust this pull request tomorrow or start a new one...

chkal:
I just submitted #595, which contains the requested subset of the pull request. So I'm closing this one now.

Issue #426, opened 2013-08-28
EntityTag.equals(): Remove unnecessary call to super.equals()


glassfishrobot:
javax.ws.rs.core.EntityTag invokes:

```
if (!(obj instanceof EntityTag))
  return super.equals(obj);
```

The second line should return false.

1.  This is the first time I've ever seen equals() delegate to super in case of a type mismatch. I find it hard to imagine when this would be useful.
2.  EntityTag extends Object, so in this case we are guaranteed super.equals(obj) will return false. We might as well return false for clarity, if nothing else.
#### Affected Versions
[2.0]

NiklasMehner:
Create PR https://github.com/eclipse-ee4j/jaxrs-api/pull/593 for this issue.

PR #593, opened 2018-02-25
Cleanup EntityTag.equals() / hashcode() (#426)

NiklasMehner:
This PR fixes issue #426 and simplifies equals/hashcode in EntityTag.

As correctly stated in the ticket the call to super.equals() can never return true, since the super is a simple instance comparison.

I have also used the Objects.equals()/hash() methods to simplify the code.

I am not particulary happy with the test since it has the side effect of initializing the EntityType.HEADER_DELEGATE to null. But I do not see any way around this as long as the HEADER_DELEGATE field is is initialized on class initialization. Perhaps this field should be "inlined" in the toString() method (but I am not sure on the performance implications of this)

The test rule could later be extended to add more functionality (if needed in other tests)

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -585,6 +585,12 @@
             <version>4.11</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>2.16.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <properties>
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
@@ -18,6 +18,7 @@ package javax.ws.rs.core;
 
 import javax.ws.rs.ext.RuntimeDelegate;
 import javax.ws.rs.ext.RuntimeDelegate.HeaderDelegate;
+import java.util.Objects;
 
 /**
  * An abstraction for the value of a HTTP Entity Tag, used as the value
@@ -99,17 +100,12 @@ public class EntityTag {
      */
     @Override
     public boolean equals(final Object obj) {
-        if (obj == null) {
-            return false;
-        }
         if (!(obj instanceof EntityTag)) {
             return false;
         }
+        
         EntityTag other = (EntityTag) obj;
-        if (value.equals(other.getValue()) && weak == other.isWeak()) {
-            return true;
-        }
-        return false;
+        return Objects.equals(value, other.getValue()) && weak == other.isWeak();
     }
 
     /**
@@ -119,10 +115,7 @@ public class EntityTag {
      */
     @Override
     public int hashCode() {
-        int hash = 3;
-        hash = 17 * hash + (this.value != null ? this.value.hashCode() : 0);
-        hash = 17 * hash + (this.weak ? 1 : 0);
-        return hash;
+        return Objects.hash(this.value, this.weak);
     }
 
     /**
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
@@ -103,7 +103,7 @@ public class EntityTag {
             return false;
         }
         if (!(obj instanceof EntityTag)) {
-            return super.equals(obj);
+            return false;
         }
         EntityTag other = (EntityTag) obj;
         if (value.equals(other.getValue()) && weak == other.isWeak()) {
--- a/jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
+++ b/jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
@@ -1,0 +1,74 @@
+package javax.ws.rs.core;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.*;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.mockito.Mockito.*;
+
+import javax.ws.rs.ext.RuntimeDelegate;
+
+public class EntityTagTest {
+
+
+    @Before
+    public void setUp() {
+        RuntimeDelegate.setInstance(mock(RuntimeDelegate.class));
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        RuntimeDelegate.setInstance(null);
+    }
+
+    @Test
+    public void shouldBeEqualToTheSameInstance() {
+        EntityTag entityTag = new EntityTag("value", true);
+
+        assertThat(entityTag, equalTo(entityTag));
+        assertThat(entityTag.hashCode(), equalTo(entityTag.hashCode()));
+    }
+
+    @Test
+    public void shouldBeEqualsForSameFieldValues() {
+        EntityTag entityTag = new EntityTag("value", true);
+        EntityTag entityTagWithSameValues = new EntityTag("value", true);
+        assertThat(entityTag, equalTo(entityTagWithSameValues));
+        assertThat(entityTag.hashCode(), equalTo(entityTagWithSameValues.hashCode()));
+    }
+
+    @Test
+    public void shouldNotBeEqualIfValueFieldDiffers() {
+        EntityTag entityTag = new EntityTag("value", true);
+        EntityTag entityTagWithDifferentValue = new EntityTag("differentValue", true);
+
+        assertThat(entityTag, not(equalTo(entityTagWithDifferentValue)));
+        assertThat(entityTag.hashCode(), not(equalTo(entityTagWithDifferentValue.hashCode())));
+    }
+
+    @Test
+    public void shouldNotBeEqualIfWeekSettingDiffers() {
+        EntityTag entityTag = new EntityTag("value", true);
+        EntityTag entityTagWithDifferentWeakSetting = new EntityTag("value", false);
+
+        assertThat(entityTag, not(equalTo(entityTagWithDifferentWeakSetting)));
+        assertThat(entityTag.hashCode(), not(equalTo(entityTagWithDifferentWeakSetting.hashCode())));
+    }
+
+    @Test
+    public void shouldNeverEqualsNull() {
+        EntityTag entityTag = new EntityTag("value", true);
+
+        assertThat(entityTag, not(equalTo(null)));
+    }
+
+    @Test
+    public void shouldNotBeEqualToObjectFromDifferentClass() {
+        EntityTag entityTag = new EntityTag("value", true);
+
+        assertThat(entityTag, not(equalTo(new Object())));
+    }
+}


NiklasMehner:
One additional remark: The refactoring of the hashcode() method is only safe, if the hashcodes of EntityTags are not persisted in any way.

mkarg:
@NiklasMehner Can you please check Signed-Off-By?

jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
```diff
@@ -99,17 +100,13 @@ public String getValue() {
      */
     @Override
     public boolean equals(final Object obj) {
-        if (obj == null) {
+        if (obj == null || !(obj instanceof EntityTag)) {
```
mkarg: `null instanceof <class>` is guaranteed to be `false` by definition, so you should omit the redundant check.

jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
```diff
@@ -119,10 +116,7 @@ public boolean equals(final Object obj) {
      */
     @Override
     public int hashCode() {
-        int hash = 3;
-        hash = 17 * hash + (this.value != null ? this.value.hashCode() : 0);
-        hash = 17 * hash + (this.weak ? 1 : 0);
-        return hash;
+        return Objects.hash(this.value, this.weak);
```
mkarg: One should never mix bug fixes with enhancements in the same commit. It prevents committers from picking bug fixes into lts branches.

jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Before
+    public void setUp() {
+        tag1 = new EntityTag("value1", true);
+        tag2 = new EntityTag("value1", true);
+        tag3 = new EntityTag("value2", true);
+        tag4 = new EntityTag("value1", false);
+    }
+
+    @Test
+    public void testEqualsAndHashCode() {
```
mkarg: Would be great if the test would be more self-describing (like BDD style), so reviewers can more easily understand what the target of the actual test is. Currently this test just says it does test methods.

jaxrs-api/src/test/java/javax/ws/rs/core/testrules/DummyRuntimeDelegate.java
```diff
@@ -0,0 +1,41 @@
+package javax.ws.rs.core.testrules;
+
+import javax.ws.rs.core.*;
+import javax.ws.rs.ext.RuntimeDelegate;
+
+/**
+ * Dummy implementation of the RuntimeDelegate. This is needed to test classes that require a RuntimeDelegate.
+ * Currently all methods just return null values.
+ */
+public class DummyRuntimeDelegate extends RuntimeDelegate {
```
mkarg: Open discussion: Do we want to collect such dummies? Personally I love Mockito... ;-)

mkarg:
@spericas Why did you close this PR?

spericas:
@mkarg Was closed when branch was deleted. Reopened now.

mkarg:
@spericas Github behaves rather strange. You deleted `3-SNAPSHOT` but this PR is against `master`. Weird.

spericas:
@mkarg It wasn't targeted at master before, I changed that when I saw your comment and re-opened it.

jaxrs-api/src/test/java/javax/ws/rs/core/testrules/DummyRuntimeDelegate.java
NiklasMehner: I have created #604 for adding the mockito dependency. I guess a change like that warrants its own PR. I am not sure what has to be checked before adding a depedency (allowed licenses etc.).

jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
NiklasMehner: ok, I understand the issue, but I think in this case this is not really a bug-fix, but only a simplification of the code (since no outside observable behavior is changed). So for me doing a PR "simplify hashcode/equals" instead of only removing one "if"-statement would be ok, but I can split the PR if you want me to.

jaxrs-api/src/test/java/javax/ws/rs/core/testrules/DummyRuntimeDelegate.java
NiklasMehner: I also just noticed there is already a similar class javax.ws.rs.core.RuntimeDelegateStub that is used in other tests.

jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
NiklasMehner: I am happy do change this. But looking at the other tests the test style seems to be the "traditional" junit test methods. So I didn't want to introduce a new style for only one test class.

If we want to use a "BDD style" we probably should refactor all tests.

jaxrs-api/src/test/java/javax/ws/rs/core/testrules/DummyRuntimeDelegate.java
mkarg: Waiting for approval of the CQ I filed (https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15982).

jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
mkarg: No need to split PRs. Just split commits within the same PR.

jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
mkarg: Refactoring sounds good since BDD is much more readable, but as it is a very heavy change, one should discuss the idea on the dev list. For now, I would be happy with writing all _new_ code in BDD style.

jaxrs-api/src/test/java/javax/ws/rs/core/testrules/DummyRuntimeDelegate.java
chkal: This may be an unpopular opinion, but I'm very conservative about adding new dependencies. In case of specifications like JAX-RS we should be very careful about dependencies, even for the `test` scope. JAX-RS should last for at least the next decade, and maintenance is usually simpler with fewer dependencies. Even if we have to create mocks our self.

However, as mentioned above, I'm very (maybe too) conservative about this. So if I'm the only one having doubts about add Mockito, just ignore my concern. :wink:

mkarg: I share your doubts in general, but Mockito in particular is way past its hype and has become a de-facto standard in lots of projects I contribute to. For me, JUnit-Hamcrest-Mockito meanwhile has became an indivisible default for Java projects.

mkarg: CQ passed.

chkal: @mkarg I agree that Mockito is the most successful mocking library for Java out there. Therefore, I would be OK with adding it. However, I just wanted to note that we should take special care when adding new dependencies. Even for the test scope.

NiklasMehner:
I have pushed an updated implementation. This should fix all issues that have come up in the review.

It also includes the changes from #604, so #604 should probably be merged first.

jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    public void shouldNotBeEqualIfValueFieldDiffers() {
+        assertThat(tag1, not(equalTo(tag3)));
+        assertThat(tag1.hashCode(), not(equalTo(tag3.hashCode())));
+    }
+
+    @Test
+    public void shouldNotBeEqualIfWeekSettingDiffers() {
```
mkarg: I wonder if we really want to have negative tests? I mean, there could happen a lot of things we do not want to happen, but do we have to test them?

jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+
+    @Before
+    public void setUp() {
+        RuntimeDelegate.setInstance(mock(RuntimeDelegate.class));
+
+        tag1 = new EntityTag("value1", true);
+        tag2 = new EntityTag("value1", true);
+        tag3 = new EntityTag("value2", true);
+        tag4 = new EntityTag("value1", false);
```
chkal: I don't want to be too niggling here, but what do you think about giving these fields more descriptive names? To understand the tests, you currently have to scroll up to the `setUp()` method to see how these `EntityTag` instances have been created, because the field names don't help much.

I personally would even go one step further and define these instances in each test method. Of course this means more code, but tests are much easier to understand.

jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
chkal: IMO this test makes sense. So let's keep it. Or what kind of downside do you see if we keep it?

jaxrs-api/pom.xml
```diff
@@ -585,6 +585,12 @@
             <version>4.11</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
chkal: As already mentioned in #604, I'm fine with adding Mockito. But I agree that we should wait for #604 to be merged before merging this one. This way the history will be much cleaner. I know this defers the merge of this PR, but I think it is worth the effort.

jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
mkarg: I think this is a question of personal taste. My person choice would be BDD style, with the typical //given section holding pretty clearly named variables instead of a setUp method at all.

mkarg: @chkal I do not see a downside. It is more a philosopical question: How to select that particular "false matches" that we want to cover with tests? Like "does not match a string", "does not match a subclass", and so on? We have to draw a line somewhere. OTOH I understand this particular test, but it more looks like "This class correctly implements 'equals'". Maybe we could find a generic solution for that test case, as it is independent of this particular class?

chkal: > I think this is a question of personal taste. My personal choice would be BDD style, with the typical //given section holding pretty clearly named variables instead of a setUp method at all.

I agree. However, I'm not too strong on this one. We could also improve the overall test structure later.

@NiklasMehner: Sorry that all these basic discussions are delaying your pull request. But there is still much very basic stuff to discuss. And primarily working on GitHub is a pretty new concept for the spec.

NiklasMehner: No, problem at all. 

Regarding the test selection:
Most of the time I'm trying to achieve 100% branch coverage. So if there is a branch like in this case:

if (!(obj instanceof EntityTag)) [...]

Both cases (expression evaluating to true and false) should be tested. I treat && and || as branches as well (they are in the bytecode).
If you then add some boder cases (null, invalid input) you imho get a reasonable set of tests.

A generic test for equals is not that easy to do. https://github.com/jqno/equalsverifier is a nice library for this, but I guess it is not worth adding a dependency just for testing a few equals-methods.

Also probably these classes can become (valhalla) value classes once we get to that kind of language level and then we won't need any equals methods any more.

I think the given/when/then is a bit overblown for tests this simple, but I will inline the fields and try to come up with some more descriptive names.

jaxrs-api/src/test/java/javax/ws/rs/core/EntityTagTest.java
NiklasMehner: See comment below: They are actually there to get to 100% branch coverage.

chkal:
@NiklasMehner Thanks a lot for cleaning up the tests. IMO this structure is much better. Awesome!

So I'm find with adding my +1 as soon as #604 is merged! :+1:

jaxrs-api/pom.xml
mkarg: I merged #604 meanwhile.

mkarg:
@NiklasMehner Can you please fixup 7fde500 into 41b3c55? It seems it is only a fix or that one, and I'd like to keep the other commits _unsquashed_ when I merge the PR.

As soon as one more committer is willing to approve this PR, I will merge it. :-)

Issue #472, opened 2014-06-18
javax.ws.rs.* packages should implement java.lang.AutoCloseable where applicable


glassfishrobot:
Currently the javax.ws.rs.core.Response and javax.ws.rs.client.Client implement close(), but not the AutoCloseable interface. This would be a nice feature so that developers can leverage the try-with-resources syntax.
#### Environment
All
#### Affected Versions
[2.0]

mkarg:
@spericas @NiklasMehner As PR #594 was merged by Santiago already, can we now close this issue or are the classes still open that someone wants to make AutoCloseable?

mkarg:
Closed by #594.

oliviercailloux:
Has this been reverted, as suggested by those commits of end-Dec-2019 and by the fact that [Client](https://javadoc.io/doc/jakarta.ws.rs/jakarta.ws.rs-api/3.0.0-M1/jakarta/ws/rs/client/Client.html) does not implement AutoCloseable in the [Maven artifact](https://search.maven.org/artifact/jakarta.ws.rs/jakarta.ws.rs-api) 3.0.0-M1 (out in Feb 2020)?

It seems that this was a valuable addition, wanted since 2014.

spericas:
@oliviercailloux Should be in the 3.1-SNAPSHOT branch. Version 3.0.0 will only include the necessary package renaming for Jakarta EE 9.

PR #594, opened 2018-02-25
Classes should implement java.lang.AutoCloseable where applicable (#472)

NiklasMehner:
Looking at issue #472 I found that Response, SseBroadcaster, SseEventSink, SseEventSource already implement AutoCloseable.

The only class that has a "void close()" method and that does not implement AutoCloseable is Client.

This PR adds the AutoCloseable interface to Client.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/client/Client.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/client/Client.java
@@ -40,7 +40,7 @@ import javax.net.ssl.SSLContext;
  * @see javax.ws.rs.core.Configurable
  * @since 2.0
  */
-public interface Client extends Configurable<Client> {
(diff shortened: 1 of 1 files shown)


NiklasMehner:
Yes, you are right, added @Override.

ggam:
Great!

chkal:
Thanks a lot! This looks great! So +1 for merging.

But could you please squash the 3 commits into a single one? Just to get a clean history.

NiklasMehner:
@chkal Squashed commits.

johnament:
What about `Response`?

NiklasMehner:
Response is already AutoClosable:
https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/core/Response.java

PR #600, opened 2018-03-17
Adjusted Checkstyle's WhitespaceAround rule for generics (#587)

chkal:
Currently Checkstyle reports 2516 (!) violations in the source code. The number increased dramatically after updating Checkstyle as part of #595.

A majority of the violations is caused by the `WhitespaceAround` rule, which is currently configured to enforce whitespaces around `<` and `>`. Therefore, this rule is violated for all generic declarations like `List<String>`.

This pull requests updates the Checkstyle configuration to not check for whitespaces around `<` and `>` anymore, which reduces the total number of violations to 674.

### DIFF
--- a/etc/config/checkstyle.xml
+++ b/etc/config/checkstyle.xml
@@ -163,7 +163,7 @@
                 LITERAL_TRY, LITERAL_WHILE, LOR, LT, MINUS, MINUS_ASSIGN, MOD,
                 MOD_ASSIGN, NOT_EQUAL, PLUS, PLUS_ASSIGN, QUESTION, SL,
                 SLIST, SL_ASSIGN, SR, SR_ASSIGN, STAR, STAR_ASSIGN,
-                TYPE_EXTENSION_AND, GENERIC_START, GENERIC_END" /> <!-- RCURLY removed -->
+                TYPE_EXTENSION_AND" /> <!-- RCURLY removed -->
         </module>
 
 


chkal:
Well, actually this rule is rather generic an allows to check for whitespaces around various tokens. The tokens `GENERIC_START` and `GENERIC_END` (which I removed in this PR) are actually not enabled by default. So I guess this is more an issue with our configuration.

mkarg:
Reading the javadocs of `GENERIC_START` to me it seems as if Checkstyle simply makes no difference _where_ the angle is used. At least I could not find different tags for _at type declaration_ to _at typification_.

spericas:
Even after the last change, the number of violations is still quite large. Tweaking/removing 3 or 4 rules can easily get this down to around 50 violations which probably need real attention. 

Here is the list that needs tweaking: LineLength, MethodLength, ModifierOrder and FinalParameters. Any reason we shouldn't do this? I could create another issue and submit a PR.

mkarg:
I personally cannot tell any reason against tweaking the ruleset. In the end, I don't know who set the ruleset up and what his intention was. I would just not remove rules that we actually want to have. For example, FinalParameters will be actually solved by a PR of mine which I am preparing at the moment, so there is no need to remove that rule.

spericas:
@mkarg Ok, I will wait for your PR and then create another issue to address this.

mkarg:
@spericas Great! :-) But it really will need a while as I am offline for the rest of this week due to business travel. :-(

spericas:
@mkarg No worries, I'm out most of the week too :)

mkarg:
@spericas Announed PR #614 is ready to get merged, just as we agreed upon in this thread (see above).

Issue #597, opened 2018-03-08
Provide api to allow SseBroadcaster.close without cascading effect  


jimma:
When there is a sink that is used via a broadcaster but also outside of it, close broadcaster will close this sink now. We need an api to do close without  a cascading effect

spericas:
Created PR https://github.com/eclipse-ee4j/jaxrs-api/pull/603

mkarg:
Closing this issue as I merged Santiago's PR #603 into master already.

PR #603, opened 2018-03-19
New close method to control cascading effect on registered SseEventSi…

spericas:
…nks. The existing close method will always close registered SseEventSinks. Issue #597.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/sse/SseBroadcaster.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/sse/SseBroadcaster.java
@@ -76,14 +76,27 @@ public interface SseBroadcaster extends AutoCloseable {
     CompletionStage<?> broadcast(final OutboundSseEvent event);
 
     /**
-     * Close the broadcaster and all registered {@link SseEventSink} instances.
-     * <p>
-     * Any other resources associated with the {@link SseBroadcaster} should be released.
+     * Close the broadcaster and all registered {@link SseEventSink} instances. Any other resources
+     * associated with the {@link SseBroadcaster} should be released. This method is equivalent to
+     * calling {@code close(true)}.
      * <p>
      * Subsequent calls have no effect and are ignored. Once the {@link SseBroadcaster} is closed,
-     * invoking any other method on the broadcaster instance would result in an {@link IllegalStateException}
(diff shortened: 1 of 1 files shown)


jaxrs-api/src/main/java/javax/ws/rs/sse/SseBroadcaster.java
```diff
@@ -76,14 +76,27 @@
...
-     * <p>
-     * Any other resources associated with the {@link SseBroadcaster} should be released.
+     * Close the broadcaster and all registered {@link SseEventSink} instances. Any other resources
+     * associated with the {@link SseBroadcaster} should be released. This method is equivalent to
+     * calling {@code close(true)}.
      * <p>
      * Subsequent calls have no effect and are ignored. Once the {@link SseBroadcaster} is closed,
-     * invoking any other method on the broadcaster instance would result in an {@link IllegalStateException}
-     * being thrown.
+     * invoking any other method on the broadcaster instance would result in an
```
mkarg: This diff shows why we need IDE formatting rules. ;-)

mkarg:
@deki Will CXF implement this if we approve the API change now?

deki:
Yes, that's not a big deal.

spericas:
@andymc12 That would have been ideal, but I don't know what a useful default would look like. Implementation of this method is trivial, though.

mkarg:
@spericas I think what @andymc12 meant is something like this...
```
default void close(boolean cascading) {
    close();
}
```
...so no vendor has to actually implement it in the first step.

The problem with this idea is that applications cannot rely on the parameter having any impact at all, rendering the new method risky to use.

andymc12:
@mkarg correct, that is what I was thinking.  I'm fine with it as-is, just wanted to cover all bases.  Thanks.

spericas:
Exactly, that is actually incorrect according to the semantics described in the Javadoc.

chkal:
I don't think that a default method would make sense here. Default methods are great if we need to extend interfaces which are implemented by users. But for interfaces implemented by the JAX-RS implementations this won't bring much benefit. Especially in this case, because there isn't a reasonable default implementation we could provide.

mkarg:
Closes #597

PR #604, opened 2018-03-22
Add mockito test dependency (2.16)

NiklasMehner:
Since adding a new dependency is probably a change that should be discussed on its own I have created this PR instead of adding the dependency as part of #593.

I think by now mock testing is a standard in using testing, so we should use mock testing. Especially since we are only creating an API, mocking the implementation part will help in a lot of cases.

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -585,6 +585,12 @@
             <version>4.11</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>2.16.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <properties>


jansupol:
A [Contribution questionnaire (CQ)](https://wiki.eclipse.org/Development_Resources/Contribution_Questionnaire) needs to be filed for mockito then?

NiklasMehner:
http://download.eclipse.org/tools/orbit/downloads/drops/R20180206163158/ already lists
* mockito 1.8.5 https://dev.eclipse.org/ipzilla/show_bug.cgi?id=3930
* mockito 1.9.5 https://dev.eclipse.org/ipzilla/show_bug.cgi?id=7600
unfortunatly I cannot open the issues. Perhaps we can piggyback on one of those?

Does anybody have any experience with this process?

NiklasMehner:
According to https://wiki.eclipse.org/images/6/60/Third_party_package_requirement.pdf we have to discuss this https://www.eclipse.org/org/documents/Eclipse_Policy_and_Procedure_for_3rd_Party_Dependencies_Final.pdf with the PMC, since we do not plan to distribute mockito from within eclipse.

mkarg:
Yes, we need CQ for any EF-external depencency, to guarantee legal consistency. CQ will be filed _by me_ as part of my review as I am _the merging commiter_, according to [Committer Due Diligence](https://www.eclipse.org/legal/committerguidelines.php)

mkarg:
Filed [CQ 15982] (https://dev.eclipse.org/ipzilla/show_bug.cgi?id=15982), waiting for EF's decision now.

mkarg:
@spericas EMO and PMC are OK with Mockito 2.16.0. What is your opinion?

Issue #607, opened 2018-03-27
Remove HEADER_DELEGATE from header classes


NiklasMehner:
The classes CacheControl, Cookie, EntityTag, NewCookie in package javax.ws.rs.core all have a static variable HEADER_DELEGATE that is initialized when the class is initalized.

The HEADER_DELEGATE is used to convert to and from the string representation in valueOf/toString.

The valueOf/toString do not seem to be used by at least RestEasy and Jersey and I do not see a use case for an "end user".

Since the static variables are not a very nice design, conflict with our push to CDI and make testing the code harder we should deprecate them for 2.2 and remove them in 3.0 (no-one comes up with a reason why they have to exist).

See also:
https://dev.eclipse.org/mhonarc/lists/jaxrs-dev/msg00221.html

TonyWeston:
Aghhhh!!

I need HEADER_DELEGATE, to be able to parse the "Cache-Control" string passed as part of a response.

Something like

Response rsp = client.getDataUsingJaxrsClient();	
long maxAge = jakarta.ws.rs.core.CacheControl(rsp.getHeaderString("Cache-Control")).getMaxAge();

Etc.

Is there an alternative way, other tan parsing the header in my own code?

mkarg:
> Is there an alternative way, other tan parsing the header in my own code?

See https://github.com/jakartaee/rest/issues/1291.

PR #608, opened 2018-03-28
Deprecate use of HEADER_DELEGATE #607

NiklasMehner:
See ticket #607

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/CacheControl.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/CacheControl.java
@@ -35,6 +35,10 @@ import javax.ws.rs.ext.RuntimeDelegate.HeaderDelegate;
  */
 public class CacheControl {
 
+    /**
+     * @deprecated This field will be removed in version 3.0. See https://github.com/eclipse-ee4j/jaxrs-api/issues/607
+     */
+    @Deprecated
     private static final HeaderDelegate<CacheControl> HEADER_DELEGATE =
             RuntimeDelegate.getInstance().createHeaderDelegate(CacheControl.class);
     private List<String> privateFields;
@@ -83,7 +87,10 @@ public class CacheControl {
      *
      * @throws IllegalArgumentException if the supplied string cannot be parsed
      *                                  or is null
+     * @deprecated This method will be removed in version 3.0. Please use
+     *   RuntimeDelegate.getInstance().createHeaderDelegate(CacheControl.class).fromString(value) instead.
      */
+    @Deprecated
     public static CacheControl valueOf(final String value) {
         return HEADER_DELEGATE.fromString(value);
     }
@@ -329,8 +336,12 @@ public class CacheControl {
      * corresponding HTTP header.
      *
      * @return a stringified cache control
+     * @deprecated The format of the toString() method is subject to change in version 3.0. Please use
+     * RuntimeDelegate.getInstance().createHeaderDelegate(CacheControl.class).toString(value) instead if you rely on
+     * the format of this method.
      */
     @Override
+    @Deprecated
     public String toString() {
         return HEADER_DELEGATE.toString(this);
     }
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/Cookie.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/Cookie.java
@@ -35,6 +35,10 @@ public class Cookie {
      * Cookies using the default version correspond to RFC 2109.
      */
     public static final int DEFAULT_VERSION = 1;
+    /**
+     * @deprecated This field will be removed in version 3.0. See https://github.com/eclipse-ee4j/jaxrs-api/issues/607
+     */
+    @Deprecated
     private static final HeaderDelegate<Cookie> HEADER_DELEGATE =
             RuntimeDelegate.getInstance().createHeaderDelegate(Cookie.class);
     private final String name;
@@ -98,7 +102,10 @@ public class Cookie {
      * @return the newly created {@code Cookie}.
      * @throws IllegalArgumentException if the supplied string cannot be parsed
      *                                  or is {@code null}.
+     * @deprecated This method will be removed in version 3.0. Please use
+     *   RuntimeDelegate.getInstance().createHeaderDelegate(Cookie.class).fromString(value) instead.
      */
+    @Deprecated
     public static Cookie valueOf(final String value) {
         return HEADER_DELEGATE.fromString(value);
     }
@@ -153,8 +160,12 @@ public class Cookie {
      * corresponding HTTP header.
      *
      * @return a stringified cookie.
+     * @deprecated The format of the toString() method is subject to change in version 3.0. Please use
+     * RuntimeDelegate.getInstance().createHeaderDelegate(Cookie.class).toString(value) instead if you rely on
+     * the format of this method.
      */
     @Override
+    @Deprecated
     public String toString() {
         return HEADER_DELEGATE.toString(this);
     }
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/EntityTag.java
@@ -29,7 +29,10 @@ import javax.ws.rs.ext.RuntimeDelegate.HeaderDelegate;
  * @since 1.0
  */
 public class EntityTag {
-
+    /**
+     * @deprecated This field will be removed in version 3.0. See https://github.com/eclipse-ee4j/jaxrs-api/issues/607
+     */
+    @Deprecated
     private static final HeaderDelegate<EntityTag> HEADER_DELEGATE =
             RuntimeDelegate.getInstance().createHeaderDelegate(EntityTag.class);
     private String value;
@@ -67,7 +70,10 @@ public class EntityTag {
      * @return the newly created entity tag.
      * @throws IllegalArgumentException if the supplied string cannot be parsed
      *                                  or is {@code null}.
+     * @deprecated This method will be removed in version 3.0. Please use
+     *   RuntimeDelegate.getInstance().createHeaderDelegate(EntityTag.class).fromString(value) instead.
      */
+    @Deprecated
     public static EntityTag valueOf(final String value) {
         return HEADER_DELEGATE.fromString(value);
     }
@@ -130,8 +136,12 @@ public class EntityTag {
      * corresponding HTTP header.
      *
      * @return a string version of the entity tag.
+     * @deprecated The format of the toString() method is subject to change in version 3.0. Please use
+     * RuntimeDelegate.getInstance().createHeaderDelegate(EntityTag.class).toString(value) instead if you rely on
+     * the format of this method.
      */
     @Override
+    @Deprecated
     public String toString() {
         return HEADER_DELEGATE.toString(this);
     }
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/NewCookie.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/NewCookie.java
@@ -317,7 +317,7 @@ public class NewCookie extends Cookie {
      *
      * @return a stringified cookie.
      * @deprecated The format of the toString() method is subject to change in a future version. Please use
-     * RuntimeDelegate.getInstance().createHeaderDelegate(CacheControl.class).toString(value) instead if you rely on
+     * RuntimeDelegate.getInstance().createHeaderDelegate(NewCookie.class).toString(value) instead if you rely on
      * the format of this method.
      */
     @Override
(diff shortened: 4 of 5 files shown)


jaxrs-api/src/main/java/javax/ws/rs/core/CacheControl.java
```diff
@@ -83,7 +87,10 @@ public CacheControl() {
      *
      * @throws IllegalArgumentException if the supplied string cannot be parsed
      *                                  or is null
+     * @deprecated This method will be removed in version 3.0. Please use
+     *   RuntimeDelegate.getInstance().createHeaderDelegate(CacheControl.class).fromString(value) instead.
      */
+    @Deprecated
```
mkarg: @NiklasMehner While I do support the removal of ```HEADER_DELEGATE``` I wonder why you want to get rid of ```valueOf``` and ```toString```?

mkarg:
@asoldano Can you please provide an "official" RESTeasy statement on this issue?

jaxrs-api/src/main/java/javax/ws/rs/core/CacheControl.java
NiklasMehner: They only complicate things: 
* I don't see a use-case for an end-user of the jax-rs api
* For a jax rs implementation this is a case of the implementation calling an api method that does nothing but look up an implementation class to do the work. Why tunnel these calls through the api? It does not add any functionality but only complicates things (also the implementations I looked at will not call these methods).

mkarg: Getting rid of methods is a breaking change and definitively needs to go into 3.0, while getting rid of internal delegates does not break anything and can be fixed by implementors by simply overwriting the methods. Hence it is essential to understand your drivers.
* Getting rid of useless methods is something else than implementing them in a default way (which is non-breaking).
* There is a difference between getting rid of the delegation, and getting rid of the methods -even without delegation- at all. So you might want to split the PR?

NiklasMehner: This PR is not breaking since it does not yet change anything but only deprecates the methods so the breaking change can be done in 3.0. 

I agree that removing the delegation is a different issue and can be done in a separate PR. But since this change is not included in this PR this would be an additional PR and not a split.

Overwriting the methods by an implementor is not possible, since the classes will be instantiated by an end-user of the api. An implementor has no chance to use an extended Version of
e.g. CacheControl:  Response.ok().cacheControl(new CacheControl());

mkarg: Agreed. :-)

asoldano:
@mkarg , well, the  valueOf/toString methods in the classes mentioned here are actually used in RESTEasy. Clearly, we could replace them with an explicit invocation to RuntimeDelegate.getInstance().createHeaderDelegate(...).fromString/toString .
To me the question is really if whether we think those methods are really hurting or not, or as exposing something the user would really never need.
As for the static HEADER_DELEGATE... that looks like an attempt to improve performances / have a consistent behavior once the class is loaded.

mkarg:
@spericas I think it is absolutely correct that the contributor wrote "3.0" as we said that we will add all breaking changes into that version. I do not see why we should not do that in exactly 3.0.

spericas:
@mkarg It's very likely that it will happen in 3.0, I just thought it was sufficient to say in a future version. Not a showstopper I suppose.

chkal:
I agree with @spericas that writing "future version" may be betten then mentioning a concrete version. But I'm not strong on this one.

mkarg:
I need to say that I made very bad experiences with inexact versions in real life (customers complained that it was not clearly told that "future" means "exactly this one release") but never experienced any bad with saying "it is defintively off in version x". We should train ourselfs to stand to our plans instead of keeping more and more things open.

mkarg:
@NiklasMehner As meanwhile two committers rejected this PR I need to ask you to modify it according to their comments.

mkarg:
@spericas @andymc12 The phrase "in version 3.0" was replaced by "in a future version" meanwhile. You did not request other changes. I will merge this PR as soon as you reviewed it again and revoke your -1 votes.

andymc12:
Hi @mkarg  - I don't think anybody has addressed @asoldano's and my other concern - that both RESTEasy and CXF depend on these methods, and it is not clear why we want to remove them.

spericas:
I was waiting for others to vote. Based on @andymc12 latest comment, this could potentially create problems for some implementations. Seems like the risk-benefit ratio is not worth it here.

mkarg:
But why did you explicitly ask to replace "3.0" by "future version" if you do not plan to deprecate these methods at all?

andymc12:
Hi @mkarg - my comments about replacing "3.0" with "future version" was intended to be a general rule.  In my initial comment, I mentioned that CXF depends on these methods - and that I did not see a compelling reason to remove.

> CXF also makes use of these methods, and while we could change them to create the HeaderDelegate from the RuntimeDelegate, I would rather not unless there is a really compelling reason to do so. I agree with @asoldano that there doesn't appear to be a really good reason to do it.

I could be persuaded (maybe I don't fully understand the benefits of this proposal), but at this point, I don't see the value of requiring implementors to change something that is working today.

Hope this helps - and if I'm missing something please let me know.  Thanks, Andy

spericas:
I agree with @andymc12. The version comments would apply to this or any other issue that fixes a version number like that. With more information about the impact on implementations, this does not seem worth to me. Perhaps someone can make a better case for this change (something I see people asking over and over in the discussion above).

arjantijms:
@spericas I personally haven't looked very deep into where the various implementations exactly use this for, but the rationale in the initial issue was this:

> Since the static variables are not a very nice design, conflict with our push to CDI and make testing the code harder we should deprecate them for 2.2 and remove them in 3.0 (no-one comes up with a reason why they have to exist).

So summarised:

1. Not a nice design
2. There is no reason why they should exist

The assumption was that both RestEasy and Jersey don't use them, and nothing in the API seems to use them either.

But in this discussion it came to light that RestEasy DOES use them, and that there's another JAX-RS implementation (CXF) that was previously forgotten, and this one appears to be using them as well.

mkarg:
I think it is always a good idea to keep an API as lean as possible, so I understand @NiklasMehner's intention to get rid of stuff which simply is standing in the way. From a user's perspective there is no need for this stuff, and from a vendor's perspective it seems to be possible to work around it easily. So let's try a second attempt to find a concensus:

@NiklasMehner I assume there are good reasons to get rid of these methods. Maybe they are a showstopper for you? I think it would be a good idea to follow @andymc12's advice to elaborate here why you want to get rid of them.

- Jersey: @jansupol IIUC Jersey doesn't need this at all.
- CXF: @andymc12 Would it be acceptable that another CXF committer, e. g. @deki, provides the needed changes to CXF? Or is your rejection not effort-driven but rather risk-driven?
- RESTeasy: @asoldano Would it be acceptable that another CXF committer (TBA) provides the needed changes to RESTeasy? Or is your rejection not effort-driven but rather risk-driven?

NiklasMehner:
Sorry for taken so long to react. I had some busy weeks.

The main reason for getting rid of these methods is that if we go all the way to embracing CDI, RuntimeDelegate will probably be a CDI managed been while CacheControl will probably still be a non-managed bean. Once the RuntimeDelegate.getInstance() goes away the only way to get the instance would be to do something like CDI.current().select(RuntimeDelegate.class).get() which is neither fast nor nice nor testable.

So removing these methods is one step in this direction. If CDI support will be limited to being able to inject some classes into an endpoint or if you have completly different opinions on how CDI support will work, removing the methods is probably not worth the effort.

mkarg:
@eclipsewebmaster ip-validation hangs. Can you please trigger the process again? Thanks.

asoldano:
@mkarg to answer your question, I have the same feeling  @deki expressed for CXF.So let's say I'm neutral on this, if it's merged, RESTEasy will adapt to it.

mkarg:
As neither @deki (for CXF) nor @asoldano (for RESTeasy) voted -1 (both are 0) I want to kindly ask @spericas and @andymc12 to overthink their voting. Basing on @NiklasMehner's last comments I think it would be a good thing to adopt his PR.

mkarg:
@eclipsewebmaster ip-validation hangs. Can you please trigger the process again? Thanks.

andymc12:
I'm still not sure about this.  I can understand the notion that `RuntimeDelegate`'s lifecycle may at some point be managed by CDI - and in that case we would need to use CDI's APIs to obtain the RuntimeDelegate's instance.  But these code changes are telling users to use the `RuntimeDelegate.getInstance()...` method as a replacement to the existing methods - if the ultimate goal is to get the RuntimeDelegate from CDI, then we shouldn't be telling users to get it from the `getInstance()` method.

Maybe we should remove the javadoc text saying to use `RuntimeDelegate.getInstance()...` for now - but make sure that implementors like CXF and RestEasy know to do that for the time being.  But at some point in the future - this approach won't work.  We would either need to update the `getInstance` method to lookup the instance using CDI or implementors would need to do that themselves.  If we choose to do the latter, then I would also recommend deprecating the `getInstance` method.  

What do you guys think about either of those (long term) options ( (1) modifying `RuntimeDelegate.getInstance` to use CDI to lookup and return the instance or (2) deprecating the `getInstance` method in favor of having callers lookup the instance via CDI ?

NiklasMehner:
@andymc12 I am all for deprecating RuntimeDelegate.getInstance(). The problem I see is that we need a working replacement for it. So in that case CDI injection of RuntimeDelegate has to be possible in JAX-RS 2.2. Is this an option? 
A deprecated method without a migration path would not help since people cannot react before JAX-RS 3.0 (or whatever version supports CDI injection for RuntimeDelegate) becomes available.

andymc12:
@NiklasMehner good point.  I would be in favor of bringing in CDI in the JAX-RS 2.2 release - but in order for it not to be breaking, we would need to support both (i.e. `@Context` and `@Inject` would both need to work - likewise, finding the `RuntimeDelegate` instance would need to work using `ServiceLoader` _and_ CDI).  If people are ok with that in JAX-RS 2.2, then I would be all for it.

Otherwise, we could put these changes (in addition to deprecating `RuntimeDelegate.getInstance()`) into the JAX-RS 3.0 stream.

spericas:
@andymc12 @NiklasMehner I would not rely on any CDI wizardry until 3.0. I find this too big of a change for 2.2.

NiklasMehner:
@spericas @mkarg ok, but this means that we have to keep RuntimeDelegate.getInstance() at least until 4.0 since removing it as a breaking change would not be possible from 3.0 -> 3.1.

Depending on how fast JAX-RS will move on we will have to support it for a long time.

Since there currently is no 3.0 branch: What should I do with this PR? Add the deprecation to RuntimeDelegate.getInstance() and keep it open until a 3.0 branch exists?

andymc12:
@NiklasMehner @mkarg @spericas This might be a little off topic (we can discuss on the mail list if it is more appropriate), but what if we made CDI optional in JAX-RS 2.2?

Basically, we could say that ‘@Inject’ could be used anywhere that ‘@Context’ would be used - and maybe some of the lifecycle annotations too...  

But the main idea is that users would not need to depend on any of the CDI APIs. Vendors would only be required to support the CDI APIs if a CDI implementation is available.  JAX-RS 2.1 has a similar provision for JSONB integration.  

If we did this, then I think we could go ahead with deprecating these methods (as well as ‘RuntimeDelegate.getInstance()’) in 2.2 and remove them in 3.0.

What do you think?  (Apologies if the format is off - I’m commenting from my phone)

mkarg:
@NiklasMehner I think what you mean is 3.x not 4.x,, as the committers agreed earlier this year to introduce CDI in 3.x (but not 2.x). Yes, there is no 3.0 branch as the majority didn't want to have such a branch. For sake of collecting such issues, we agreed upon me collecting them in the 3.0 _milestone_.

mkarg:
@andymc12 CDI is already optional since JAX-RS 2.1 (see chapter 11.2.3 of the spec). So basically you like to extend that chapter with a few sentences? Sounds feasible to me. @jansupol Do you think Andy's proposal could be easily implemented in Jersey in the next months or is this too complicated / too much change?

BTW, yes, Andy I think it is worth opening an issue for that proposal. :-)

NiklasMehner:
@mkarg No, I was talking about 4.0. We need:

- One release that supports both non-CDI and CDI and deprecates the non-CDI part
- One release that removes the deprecated parts

The first release can be a minor release since only features are added. The second release has to be a major release, since it breaks things. So if 3.0 is the first release, the second release has to be 4.0 .

mkarg:
@NiklasMehner If the one with _optional_ CDI does not break anything it could be 2.2, as @andymc12 already pointed out. I am only against _mandatory_ CDI in 2.2 as it would break existing applications.

spericas:
@andymc12 @Inject and @Context have different targets. @Inject cannot be used for params and @Context cannot be used in constructors. Moreover, a very popular use case for @Context is in param position. I'm not sure this suggestion will help a smoother transition to an @Context-free version of JAX-RS, it may actually create even more confusion among users in the meantime. The move to CDI should be an all or nothing step IMO.

PR #610, opened 2018-03-28
Removed references to java.net and updated some URLs.

spericas:
A few basic updates to the pom. More changes may be necessary to clean up the profiles, but it is a start.

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -20,19 +20,13 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>net.java</groupId>
-        <artifactId>jvnet-parent</artifactId>
-        <version>5</version>
-    </parent>
-
     <groupId>javax.ws.rs</groupId>
     <artifactId>javax.ws.rs-api</artifactId>
     <version>2.2-SNAPSHOT</version>
     <packaging>${packaging.type}</packaging>
     <name>javax.ws.rs-api</name>
 
-    <url>https://github.com/jax-rs/api</url>
+    <url>https://github.com/eclipse-ee4j/jaxrs-api</url>
 
     <organization>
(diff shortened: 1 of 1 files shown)


jansupol:
Currently there seems to be a discussion on creating parent pom for all the projects with java.net parent. Should one be created, it should have been referenced. Erasing the parent seems to be premature right now.

spericas:
@jansupol OK, I will wait on the merge then. Otherwise, refactor it to keep that parent.

mkarg:
@jansupol Can you please share a link to that discussion?

m0mus:
Markus, discussion is Oracle internal now. The idea is to create a parent pom prototype as a starting point for discussion with PMC. If you want to take a task of creating a prototype you are welcome.

chkal:
I would be fine with merging the PR now, even if there **may** be a common parent in the future. Keeping the java.net parent feels wrong. ;-)

mkarg:
@m0mus I can provide a prototype pom.xml if somebody would be willing to discuss this outside of Oracle so I know what actually shall be the idea of that pom: Do you really want a global EE4J super pom and why? I do not see the use of that. In real life (approx. 100 projects under one common hood) we made rather experience with that. In the end it only contained the organization and the licence because every sub project had to much changes. Would be good if not the PMC or Oracle discusses that but if you just wait for the subproject committers to ASK for such a pom. The EF way is to bubble things up.

m0mus:
@mkarg, what you are saying makes sense. On the other hand, all EE4J projects are a part of EE4J umbrella, hosted by Eclipse Foundation and using Eclipse build infra. In parent .pom we can have repositories management, configuration of plugins recommended for all projects, such as license check plugin, etc. And it worked well in java.net times.

mkarg:
@m0mus Yes that makes sense but still I think we should follow the EF's typical way of letting committers bubble up solutions to the PMC instead of Oracle/PMC coming up with final solutions that are forced down into the projects.

chkal:
I agree with @mkarg that such "global" parent poms don't bring much benefit. However, IF there is some parent pom for EE4J containing common configuration like the release repositories, we should of course adopt it. But waiting for this parent pom shouldn't block this PR from being merged.

spericas:
I just merged it. We can review this later on if a parent pom is created that we can use.

— Santiago

mkarg:
I don't want to be smart-alec but you merged before two weeks of review had been over. ;-)

(Just kidding, as we agreed that it is up to the assignee to decide differently)

PR #611, opened 2018-03-29
First version of .travis.yml

chkal:
The Eclipse Webmaster team has already enabled Travis for our repository.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=533020

This pull request contains a first simple `.travis.yml` file. It will build both the API and the example project with JDK8 and JDK9. I think that's a good start. 

Unfortunately JDK10 isn't available on Travis by default, but there are ways to install it manually during the build. I'll provide a separate pull request for an enhanced setup with more JDK versions in the next days.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+dist: trusty
+language: java
+sudo: false
+
+jdk:
+  - oraclejdk8
+  - oraclejdk9
+
+install:
+  - cd $TRAVIS_BUILD_DIR/jaxrs-api
+  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+  - cd $TRAVIS_BUILD_DIR/examples
+  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+
+script:
+  - cd $TRAVIS_BUILD_DIR/jaxrs-api
+  - mvn test -B
+  - cd $TRAVIS_BUILD_DIR/examples
+  - mvn test -B
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 
 ---
 
-# JAX-RS API
+# JAX-RS API [![Build Status](https://travis-ci.org/eclipse-ee4j/jaxrs-api.svg?branch=master)](https://travis-ci.org/eclipse-ee4j/jaxrs-api)
 
 This repository contains JAX-RS API source code.
 


chkal:
Please click on "Show all checks" below to see the Travis integration in action...

mkarg:
This PR was merged after just two (not three) positive reviews after just two days of review period. I love this contribution, but the question is whether we should do that, looking at what Santiago wrote in the wiki recently?
- Is it ok to assign my own PRs to myself?
- Is it ok to review my own PRs?
- Is it ok to merge after just two days of review period?

Because, in the end others might have no chance to review or vote -1, or one could say that committers apparently need only two reviews (as their own +1 is implied).

Maybe we should discuss this in a new thread on the mailing list?

chkal:
Hi @mkarg, 

thanks for your comment. Let me respond on a few things:

> This PR was merged after just two (not three) positive reviews

Correct, there are only two reviews. But merging requires three committers agreeing on a change. In this case the PR author is also a committer and therefore I think that creating a pull request for a change can be defined as an implicit +1 vote by the author. Therefore, if a PR is created by a committer, only two additional reviews are required. Does this make sense?

> after just two days of review period.

I agree that merging was performed too early. Sorry about that. I didn't check the date and actually thought the PR was open for review longer than just two days. Seems like my mind tricked me. Maybe because I started the first discussion about this about a week ago.

However, we talked about voting times back then on the mailing list and didn't agree on any specific period. Some people wrote that one week is fine, others preferred two weeks. I guess that was also the reason for @spericas to not include it in the wiki page. But we should definitely define something for that.

> Is it ok to assign my own PRs to myself?

Good question. IMO this is fine. We defined the assignee as a committer who is "in charge". So my impression was that in case of a PR which was created by a committer, it would be fine if the author assigns the PR to himself. However, I agree that there are also good arguments for separating author and assignee. Maybe we should really discuss this on the list.

> Is it ok to review my own PRs?

You cannot review your own pull request using GitHub. But if a committer creates a PR, we could count that as a review. At least in my opinion.

To sum up. Perhaps we should really discuss some of these things on the list.

mkarg:
@chkal Started discussion on the list. Let's wait for feedback. While I would have done it the same way you did, it might be better if we raise the bar a bit for the future. I tend to be overhastily with my own PRs. ;-)

chkal:
@mkarg Sure! I agree that we should discuss such topics on the list and find a general consensus. :+1:

PR #614, opened 2018-04-01
Added 'final' keyword where appropriate (#587)

mkarg:
Reduces checkstyle issues from 674 to 271.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/BadRequestException.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/BadRequestException.java
@@ -43,7 +43,7 @@ public class BadRequestException extends ClientErrorException {
      * @param message the detail message (which is saved for later retrieval
      *                by the {@link #getMessage()} method).
      */
-    public BadRequestException(String message) {
+    public BadRequestException(final String message) {
         super(message, Response.Status.BAD_REQUEST);
     }
 
@@ -54,7 +54,7 @@ public class BadRequestException extends ClientErrorException {
      * @throws IllegalArgumentException in case the status code set in the response
      *                                  is not HTTP {@code 400}.
      */
-    public BadRequestException(Response response) {
+    public BadRequestException(final Response response) {
         super(validate(response, Response.Status.BAD_REQUEST));
     }
 
@@ -67,7 +67,7 @@ public class BadRequestException extends ClientErrorException {
      * @throws IllegalArgumentException in case the status code set in the response
      *                                  is not HTTP {@code 400}.
      */
-    public BadRequestException(String message, Response response) {
+    public BadRequestException(final String message, final Response response) {
         super(message, validate(response, Response.Status.BAD_REQUEST));
     }
 
@@ -76,7 +76,7 @@ public class BadRequestException extends ClientErrorException {
      *
      * @param cause the underlying cause of the exception.
      */
-    public BadRequestException(Throwable cause) {
+    public BadRequestException(final Throwable cause) {
         super(Response.Status.BAD_REQUEST, cause);
     }
 
@@ -87,7 +87,7 @@ public class BadRequestException extends ClientErrorException {
      *                by the {@link #getMessage()} method).
      * @param cause   the underlying cause of the exception.
      */
-    public BadRequestException(String message, Throwable cause) {
+    public BadRequestException(final String message, final Throwable cause) {
         super(message, Response.Status.BAD_REQUEST, cause);
     }
 
@@ -99,7 +99,7 @@ public class BadRequestException extends ClientErrorException {
      * @throws IllegalArgumentException in case the status code set in the response
      *                                  is not HTTP {@code 400}.
      */
-    public BadRequestException(Response response, Throwable cause) {
+    public BadRequestException(final Response response, final Throwable cause) {
         super(validate(response, Response.Status.BAD_REQUEST), cause);
     }
 
@@ -113,7 +113,7 @@ public class BadRequestException extends ClientErrorException {
      * @throws IllegalArgumentException in case the status code set in the response
      *                                  is not HTTP {@code 400}.
      */
-    public BadRequestException(String message, Response response, Throwable cause) {
+    public BadRequestException(final String message, final Response response, final Throwable cause) {
         super(message, validate(response, Response.Status.BAD_REQUEST), cause);
     }
 }
--- a/jaxrs-api/src/main/java/javax/ws/rs/ClientErrorException.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/ClientErrorException.java
@@ -37,7 +37,7 @@ public class ClientErrorException extends WebApplicationException {
      *                                  {@link javax.ws.rs.core.Response.Status.Family#CLIENT_ERROR} status code
      *                                  family.
      */
-    public ClientErrorException(Response.Status status) {
+    public ClientErrorException(final Response.Status status) {
         super((Throwable) null, validate(Response.status(status).build(), Response.Status.Family.CLIENT_ERROR));
     }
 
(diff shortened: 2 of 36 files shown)


ggam:
El jue., 5 abr. 2018 16:53, Christian Kaltepoth <notifications@github.com>
escribió:

> *@chkal* commented on this pull request.
>
> I'm -0 for this, which means something like "I won't get in the way, but
> would prefer to not merge it"
>
> The reasons:
>
>    - Although adding final enforces immutability for variables and
>    parameters, it adds MUCH boilerplate to the code.
>
> That's true :)

>
>    - Even if parameters are modified or variables are reused, this isn't
>    actually a bug but just bad practice. Maybe there are ways (like
>    Findbugs/Checkstyle) to check for such bad practices without adding the
>    boilerplate to the code?
>
> The API can be checked but not the user code. Preventing bad practices
from people extending those classes is a nice benefit in my opinion. Lots
of noise on the API though, yes.

I like the idea anyway, but what do you all think about asking on a broader
list what other projects think? IMO this kind of things should be better
determined by an umbrella project.

spericas:
Markus,

 Could I demand a change to remove all those final’s? ;-)

 I sort of share Christian’s concern about the verbosity of this pattern (and wish final was the default in the language!) but at this point, we should move on as Markus says. We could revisit this later if needed.

— Santiago

arjantijms:
I do share the opinion of the others here. While I do see some small theoretical benefit of marking everything as final, the verbosity, which already plagues Java a bit too much, isn't really worth it.

In fact, one of the first things that I typically do when I inherit a code base is removing all those verbose final keywords.

One way or the other, a umbrella group / architecture board should define a coding style that works for all. We all have our little own things that we lobe, but which may not work for everyone. If you take a look at the OmniFaces source code, you'll find I'm particularly a fan of static imports, but I also know not everyone shares this.

As I have this before in a role as lead developer I'm willing to devote some time to set up an initial code convention. In my experience, things that already have broad support typically work best, which means the slightly updated Sun code conventions (Sun, but with slightly longer lines, emphasis on only spaces for indenting, builder style for functional / stream coding), unix line endings, UTF-8 encoding, no trailing white space, and a selection of the most accepted practices recommended by Robert C. Martin in Clean Code.

Organising imports would be good to add to this as well, but indeed a check whether Eclipse and NetBeans and IntelliJ don't do this differently would be needed.

mkarg:
-1 for removing all finals

-1 for disucssing this _general_ issue within this specific bugfix-PR

+1 for opening your own issue wrt to removing all finals and then submitting your own PR effectively removing all existing finals

arjantijms:
See the edited reply (was edited immediately after typing it):

>In fact, one of the first things that I typically do when I inherit a code base is removing all those verbose final keywords **(unless of course where they actually are quite important, such as for final classes and final instance variables that actually must be final)**.

I would of course not argue to remove *every* final, but only those where there's barely any point of having them. Like those here:

```java
	1	    static Builder newBuilder() {
	2	         try {
	3	             final Object delegate = FactoryFinder.find(JAXRS_DEFAULT_SSE_BUILDER_PROPERTY,
	4	                     JAXRS_DEFAULT_SSE_BUILDER, SseEventSource.Builder.class);
	5	             if (!(delegate instanceof Builder)) {
	6	                 final Class pClass = Builder.class;
	7	                 final String classnameAsResource = pClass.getName().replace('.', '/') + ".class";
	8	                 ClassLoader loader = pClass.getClassLoader();
	9	                 if (loader == null) {
	10	                     loader = ClassLoader.getSystemClassLoader();
	11	                 }
	12	                 final URL targetTypeURL = loader.getResource(classnameAsResource);
	13	                 throw new LinkageError("ClassCastException: attempting to cast"
	14	                                        + delegate.getClass().getClassLoader().getResource(classnameAsResource)
	15	                                        + " to " + targetTypeURL);
	16	             }
	17	             return (Builder) delegate;
	18	         } catch (final Exception ex) {
	19	             throw new RuntimeException(ex);
	20	       
```

mkarg:
I understand your point, but see, this is simply personal taste, so it is completely off-topic to _this_ PR, as these finals are needed to match the _current_ checkstyle settings. And modifying the checkstyle settings is also off-topic to _this_ PR. You need to start a different issue if you want a different solution. _This_ PR solely is good for fixing the current project _as-is_.

arjantijms:
Sorry, I may be thick, but didn't Christian already update the checkstyle rules and mentioned it brought the violations down to 88?

I'm referring to this comment:

https://github.com/eclipse-ee4j/jaxrs-api/issues/587#issuecomment-367918270

The reported bug is "There are 984 checkstyle errors",  and the discussion was to fix the config / rules. I might be missing something, so my apologies if I do, but where exactly does it say in #587 that the fix would be to add final?

chkal:
Hey all,

I don't think that we agreed on adding finals everywhere. But maybe I missed something in the conversation? Let me try to summarize the discussion of the last weeks.

 * @mkarg opened #587 which is about the large number of Checkstyle violations. 
 * After that I created pull request #588 which updated Checkstyle and adjusted a few rules, including removal of the FinalParameters rule. @mkarg raised concerns about the rules I adjusted and proposed to discuss the code conventions on the list. I agreed and closed #588.
 * We started a discussion about various code conventions on the mailing list. [I proposed](https://dev.eclipse.org/mhonarc/lists/jaxrs-dev/msg00056.html) to drop the FinalParameters rule. @mkarg and @ggam voted for keeping the rule. Both @spericas (see [here](https://dev.eclipse.org/mhonarc/lists/jaxrs-dev/msg00079.html)) and @arjantijms (see [here](https://dev.eclipse.org/mhonarc/lists/jaxrs-dev/msg00058.html)) mentioned that final parameters are very noisy and that they dislike them. @arjantijms even explicitly voted against them.
  * I created #600 to fix the WhitespaceAround rule. In this pull request @spericas [proposed again](https://github.com/eclipse-ee4j/jaxrs-api/pull/600#issuecomment-374313950) to drop the FinalParameters rule. @mkarg replied that he is preparing a pull request for adding the final parameters.

So to sum up: I don't think that we agreed on keeping the rule and adjusting the code. Instead, the topic was rather controversial. Or did I miss some discussion or vote in my summary?

However, I don't want to block this PR from being merged. Therefore, I voted -0 for the reasons already mentioned before.

ggam:
> A small addition is that the PR doesn't just add final parameters, but adds final to local variables as well.

That puzzled me a bit too. Adding it to local variables adds very little value in my opinion.

mkarg:
Thanks @chkal for your recapitulation of the several threads and confirming that in the end we *already agreed* that I create this PR. As I did what we agreed upon upfront (i. e. fixing the existing code so it *matches* the existing rules) I do not see any need for further discussion unless I _incidently_ did too much as @arjantijms points out. In this case, please use the *Github Review Tool* to show me the code line (just as we agreed upon earlier this year in a separate thread) instead of simply posting lengthy comments about likes and dislikes regarding code style; I will fix that ASAP certainly by adding another commit ontop, as this helps to have an *initial* alignment between what the code liks like *at the moment* and what checkstyle wants it to look like *at the moment*.

*After that* I would be more than glad to discuss *format and rule changes* in a _separate_ issue, as it is not a *fix* anymore but a simple personal *wish* of a group of committers to explicitly change *existing* rules for the *future*.

mkarg:
With my recent commit of today, I followed your requests and have reduced the number of additional 'finals' to really only match the cases actually complained by checkstyle. Thanks @arjantijms for pointing this out. In particular, this PR _only_ adds `final` to method parameters as checkstyle wants, but does _not_ make private fields or local variables final anymore, as apparently checkstyle is satisfied already wrt.

As this is what the project was configured to be like before, I hope you will accept my PR now and provide one more +1 so I can finally fix this project bug.

mkarg:
@ggam @arjantijms +1 now?

ggam:
+1 from me, although I'm not a commiter

Issue #601, opened 2018-03-17
Formating Rulesets for IDEs


mkarg:
It would be great it one would find the time to donate some formatting ruleset files for IDEs, matching the currently applied formatting style of the existing code. This makes it easier for hacking the code with the IDE of choice. :-)

spericas:
@mkarg I don't believe I have those exact rules, but let me poke around and see if I can find something close.

mkarg:
@spericas I assume you simply applied the original Sun Java rules? Or did Oracle have a different company style?

rogergl:
So far I only know of editoroconfig to accomplish something like this.  Any other ideas ?

mkarg:
@rogergl Eclipse for example can export the complete set of formatting rules as a single file. Possiby other IDEs can do the same. The actual problem infact is that nobody knows what the settings have to look like. So to solve _this_ issue, one has to go through all the settings of his IDE and tune them as long as the formatting outcome produces a more or less empty git diff to master's HEAD. Once that is done, others can try to apply the same rules to other IDEs. So in the end, we will have one PR for Eclipse, one PR for NetBeans, etc.

rogergl:
@mkarg Ok I could test for IntelliJ. Nevertheless I think it would make sense to provide a simple .editorconfig. That would at least help some users of non IDEs (Visual Code, Vim) and would provide a common basic config across different IDEs.

mkarg:
@rogergl A PR with a ```.editorconfig``` would be terrific, but please take care that it will match the _existing_ code style as perfect as possible. I want to split up the creation of the config file from the discussion about _changing_ the style.

arjantijms:
> but please take care that it will match the existing code style as perfect as possible.

I don't think there really is an existing code style yet, is there?

So let's establish that style first, and then create a formatting ruleset for it that at least partially matches that style. Partially, since most IDEs probably don't have formatting that applies all rules (i.e. I think Eclipse has no formatting to automatically apply the modifier like `private static final).

mkarg:
@arjantijms The plan is to _fix existing_ things first and change them later _if needed_. So if you simply _want_ to change the style then please start a separate discussion in a separate issue but do not stop contributors from working on this one. Thanks.

There definitively *is* an existing code style as there is existing code and a short check using Eclipse tells us that it seems to be a rather simple deviation of the original Sun style.

arjantijms:
[...]

PR #615, opened 2018-04-01
Organized imports

mkarg:
Corrected lexical order of imports, to stop checkstyle from telling me that I should organize my imports... ;-)

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/NotAuthorizedException.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/NotAuthorizedException.java
@@ -16,14 +16,15 @@
 
 package javax.ws.rs;
 
+import static javax.ws.rs.core.HttpHeaders.WWW_AUTHENTICATE;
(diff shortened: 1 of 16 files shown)


chkal:
I'm basically +1 for this. Especially because this PR removes the wildcard imports.

However, the PR also changes the order of import statements. While I don't have a problem with changing that order, it may be a good idea to have a matching code format profile first which defines the desired import order. AFAIK both Eclipse and IntelliJ support to specify such ordering in the code format profile. Without that there is a high chance that the next person performing a "organize imports" changes the order back again.

mkarg:
@chkal See #601. While I see your point, I hate that at every second discussion we end up with "maybe we should wait until...". Let's just accept my initial change and discuss afterwards, it improves the project, it does not impose any real harm. It might lead to a change once later, but chances are like zero as all those other things we should wait for simply never seem to happen. The project stucks more and more due to all that "maybe" discussions.

PR #619, opened 2018-04-09
Java SE Bootstrapping API

mkarg:
This PR implements a vendor-neutral way to bootstrap JAX-RS applications in pure Java SE environments (i. e. making the application self-contained, without being powered by an application server product). It just provides a common API for starting up the application and runtime, but does not make any contstraints about supported features or technology.

Signed-off-by: Markus KARG <markus@headcrashing.eu>

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
@@ -1,0 +1,697 @@
+/*
+ * Copyright (c) 2018 Markus KARG. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package javax.ws.rs;
+
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+import java.util.function.BiFunction;
+
+import javax.net.ssl.SSLContext;
+import javax.ws.rs.core.Application;
+import javax.ws.rs.ext.RuntimeDelegate;
+
+/**
+ * Bootstrap class used to startup a JAX-RS application in Java SE environments.
+ * <p>
+ * The {@code JAXRS} class is available in a Jakarta EE container environment as
+ * well; however, support for the Java SE bootstrapping APIs is <em>not
+ * required</em> in container environments.
+ * </p>
+ * <p>
+ * In a Java SE environment an application is getting started by the following
+ * command using default configuration values (i. e. mounting application at
+ * {@code http://localhost:80/} <em>or a different port</em> (there is <em>no
+ * particular default port</em> mandated by this specification). As the JAX-RS
+ * implementation is free to choose any port by default, the caller will not
+ * know the actual port unless explicitly checking the actual configuration of
+ * the instance started:
+ * </p>
+ *
+ * <pre>
+ * Application app = new MyApplication();
+ * JAXRS.Configuration config = JAXRS.Configuration.builder().build();
+ * JAXRS.start(app, config).thenAccept(instance -&gt; instance.configuration().port());
+ * </pre>
+ *
+ * <p>
+ * Running instances can be instructed to stop serving the application:
+ * </p>
+ *
+ * <pre>
+ * JAXRS.start(app, config).thenAccept(instance -&gt; { ... instance.stop(); } );
+ * </pre>
+ *
+ * <p>
+ * A shutdown callback can be registered which will get invoked once the
+ * implementation stops serving the application:
+ * </p>
+ *
+ * <pre>
+ * instance.stop().thenAccept(stopResult -&gt; ...));
+ * </pre>
+ *
+ * {@code stopResult} is not further defined but solely acts as a wrapper around
+ * a native result provided by the particular JAX-RS implementation. Portable
+ * applications should not assume any particular data type or value.
+ *
+ * <p>
+ * Protocol, host address, port and root path can be overridden explicitly. As
+ * the JAX-RS implementation is bound to that values, no querying of the actual
+ * configuration is needed in that case:
+ * </p>
+ *
+ * <pre>
+ * JAXRS.Configuration.builder().protocol("HTTPS").host("0.0.0.0").port(8443).rootPath("api").build();
+ * </pre>
+ *
+ * <p>
+ * TLS can be configured by explicitly passing a customized {@link SSLContext}:
+ * </p>
+ *
+ * <pre>
+ * SSLContext tls = SSLContext.getInstance("TLSv1.2");
+ * // ...further initialize context here (see JSSE API)...
+ * JAXRS.Configuration.builder().protocol("HTTPS").sslContext(tls).build();
+ * </pre>
+ *
+ * <p>
+ * In case of HTTPS, client authentication can be enforced to ensure that only
+ * <em>trustworthy</em> clients can connect:
+ * </p>
+ *
+ * <pre>
+ * JAXRS.Configuration.builder().protocol("HTTPS").sslClientAuthentication(SSLClientAuthentication.MANDATORY).build();
+ * </pre>
+ *
+ * <p>
+ * Implementations are free to support more use cases by native properties,
+ * which effectively render the application non-portable:
+ * </p>
+ *
+ * <pre>
+ * JAXRS.Configuration.builder().property("productname.foo", "bar").build()
+ * </pre>
+ *
+ * <p>
+ * Bulk-loading allows to attach configuration storages easily without the need
+ * to write down all properties to be transferred. Hence, even properties
+ * unknown to the application author will get channeled into the implementation.
+ * This can be done both, explicitly (hence portable) and implicitly (hence
+ * <em>not necessarily</em> portable as no particular configuration mechanics
+ * are required to be supported by compliant implementations):
+ * </p>
+ *
+ * <pre>
+ * // Explicit use of particular configuration mechanics is portable
+ * JAXRS.Configuration.builder().from((name, type) -&gt; externalConfigurationSystem.getValue(name, type)).build();
+ *
+ * // Implicitly relying on the support of particular configuration mechanics by
+ * // the actual JAX-RS implementation is not necessarily portable
+ * JAXRS.Configuration.builder().from(externalConfigurationSystem).build();
+ * </pre>
+ *
+ * @author Markus KARG (markus@headcrashing.eu)
+ * @since 2.2
+ */
+public interface JAXRS {
+
+    /**
+     * Starts the provided application using the specified configuration.
+     *
+     * <p>
+     * This method is intended to be used in Java SE environments only. The outcome
+     * of invocations in Jakarta EE container environments is undefined.
+     * </p>
+     *
+     * @param application
+     *            The application to start up.
+     * @param configuration
+     *            Provides information needed for bootstrapping the application.
+     * @return {@code CompletionStage} (possibly asynchronously) producing handle of
+     *         the running application {@link JAXRS.Instance instance}.
+     * @see Configuration
+     * @since 2.2
+     */
+    static CompletionStage<Instance> start(final Application application, final Configuration configuration) {
+        return RuntimeDelegate.getInstance().bootstrap(application, configuration);
+    }
+
+    /**
+     * Provides information needed by the JAX-RS implementation for bootstrapping an
+     * application.
+     * <p>
+     * The configuration essentially consists of a set of parameters. While the set
+     * of actually effective keys is product specific, the key constants defined by
+     * the {@link JAXRS.Configuration} interface MUST be effective on all compliant
+     * products. Any unknown key MUST be silently ignored.
+     * </p>
+     *
+     * @author Markus KARG (markus@headcrashing.eu)
+     * @since 2.2
+     */
+    public static interface Configuration {
+
+        /**
+         * Configuration key for the protocol an application is bound to.
+         * <p>
+         * A compliant implementation at least MUST accept the strings {@code "HTTP"}
+         * and {@code "HTTPS"} if these protocols are supported.
+         * </p>
+         * <p>
+         * The default value is {@code "HTTP"}.
+         * </p>
+         *
+         * @since 2.2
+         */
+        static final String PROTOCOL = "javax.ws.rs.JAXRS.Protocol";
+
+        /**
+         * Configuration key for the hostname or IP address an application is bound to.
+         * <p>
+         * A compliant implementation at least MUST accept string values bearing
+         * hostnames, IP4 address text representations, and IP6 address text
+         * representations. If a hostname string, the special IP4 address string
+         * {@code "0.0.0.0"} or {@code "::"} for IP6 is provided, the application MUST
+         * be bound to <em>all</em> IP addresses assigned to that hostname. If the
+         * hostname string is {@code "localhost"} the application MUST be bound to the
+         * local host's loopback adapter <em>only</em>.
+         * </p>
+         * <p>
+         * The default value is {@code "localhost"}.
+         * </p>
+         *
+         * @since 2.2
+         */
+        static final String HOST = "javax.ws.rs.JAXRS.Host";
+
+        /**
+         * Configuration key for the TCP port an application is bound to.
+         *
+         * <p>
+         * A compliant implementation MUST accept {@code java.lang.Integer} values.
+         * </p>
+         * <p>
+         * There is no default <em>port</em> mandated by this specification, but the
+         * default <em>value</em> of this property is {@link #DEFAULT_PORT} (i. e.
+         * <code>-1</code>). A compliant implementation MUST use its own default
+         * <em>port</em> when the <em>value</em> <code>-1</code> is provided, and MAY
+         * apply (but is not obligated to) auto-selection and range-scanning algorithms.
+         * </p>
+         *
+         * @since 2.2
+         */
+        static final String PORT = "javax.ws.rs.JAXRS.Port";
+
+        /**
+         * Configuration key for the root path an application is bound to.
+         * <p>
+         * The default value is {@code "/"}.
+         * </p>
+         *
+         * @since 2.2
+         */
+        static final String ROOT_PATH = "javax.ws.rs.JAXRS.RootPath";
+
+        /**
+         * Configuration key for the secure socket configuration to be used.
+         * <p>
+         * The default value is {@link SSLContext#getDefault()}.
+         * </p>
+         *
+         * @since 2.2
+         */
+        static final String SSL_CONTEXT = "javax.ws.rs.JAXRS.SSLContext";
+
+        /**
+         * Configuration key for the secure socket client authentication policy.
+         *
+         * <p>
+         * A compliant implementation MUST accept {@link SSLClientAuthentication} enums.
+         * </p>
+         * <p>
+         * The default value is {@code SSLClientAuthentication#NONE}.
+         * </p>
+         *
+         * @since 2.2
+         */
+        static final String SSL_CLIENT_AUTHENTICATION = "javax.ws.rs.JAXRS.SSLClientAuthentication";
+
+        /**
+         * Secure socket client authentication policy
+         *
+         * <p>
+         * This policy is used in secure socket handshake to control whether the server
+         * <em>requests</em> client authentication, and whether <em>successful</em>
+         * client authentication is <em>mandatory</em> (i. e. connection attempt will
+         * fail for invalid clients).
+         * </p>
+         *
+         * @author Markus KARG (markus@headcrashing.eu)
+         * @since 2.2
+         */
+        public enum SSLClientAuthentication {
+
+            /**
+             * Server will <em>not request</em> client authentication.
+             *
+             * @since 2.2
+             */
+            NONE,
+
+            /**
+             * Client authentication is performed, but invalid clients are
+             * <em>accepted</em>.
+             *
+             * @since 2.2
+             */
+            OPTIONAL,
+
+            /**
+             * Client authentication is performed, and invalid clients are
+             * <em>rejected</em>.
+             *
+             * @since 2.2
+             */
+            MANDATORY
+        }
+
+        /**
+         * Special value for {@link #PORT} property indicating that the implementation
+         * MUST use its default port.
+         *
+         * @since 2.2
+         */
+        static final int DEFAULT_PORT = -1;
+
+        /**
+         * Returns the value of the property with the given name, or {@code null} if
+         * there is no property of that name.
+         *
+         * @param name
+         *            a {@code String} specifying the name of the property.
+         * @return an {@code Object} containing the value of the property, or
+         *         {@code null} if no property exists matching the given name.
+         * @since 2.2
+         */
+        Object property(String name);
+
+        /**
+         * Convenience method to get the {@code protocol} to be used.
+         * <p>
+         * Same as if calling {@link #property(String) (String) property(PROTOCOL)}.
+         * </p>
+         *
+         * @return protocol to be used (e. g. {@code "HTTP")}.
+         * @throws ClassCastException
+         *             if protocol is not a {@link String}.
+         * @see JAXRS.Configuration#PROTOCOL
+         * @since 2.2
+         */
+        default String protocol() {
+            return (String) property(PROTOCOL);
+        }
+
+        /**
+         * Convenience method to get the {@code host} to be used.
+         * <p>
+         * Same as if calling {@link #property(String) (String) property(HOST)}.
+         * </p>
+         *
+         * @return host name or IP address to be used (e. g. {@code "localhost"} or
+         *         {@code "0.0.0.0"}).
+         * @throws ClassCastException
+         *             if host is not a {@link String}.
+         * @see JAXRS.Configuration#HOST
+         * @since 2.2
+         */
+        default String host() {
+            return (String) property(HOST);
+        }
+
+        /**
+         * Convenience method to get the actually used {@code port}.
+         * <p>
+         * Same as if calling {@link #property(String) (int) property(PORT)}.
+         * </p>
+         * <p>
+         * If the port was <em>not explicitly</em> given, this will return the port
+         * chosen implicitly by the JAX-RS implementation.
+         * </p>
+         *
+         * @return port number <em>actually</em> used (e. g. {@code 8080}).
+         * @throws ClassCastException
+         *             if port is not an {@code Integer}.
+         * @see JAXRS.Configuration#PORT
+         * @since 2.2
+         */
+        default int port() {
+            return (int) property(PORT);
+        }
+
+        /**
+         * Convenience method to get the {@code rootPath} to be used.
+         * <p>
+         * Same as if calling {@link #property(String) (String) property(ROOT_PATH)}.
+         * </p>
+         *
+         * @return root path to be used, e. g. {@code "/"}.
+         * @throws ClassCastException
+         *             if root path is not a {@link String}.
+         * @see JAXRS.Configuration#ROOT_PATH
+         * @since 2.2
+         */
+        default String rootPath() {
+            return (String) property(ROOT_PATH);
+        }
+
+        /**
+         * Convenience method to get the {@code sslContext} to be used.
+         * <p>
+         * Same as if calling {@link #property(String) (SSLContext)
+         * property(SSL_CONTEXT)}.
+         * </p>
+         *
+         * @return root path to be used, e. g. {@code "/"}.
+         * @throws ClassCastException
+         *             if sslContext is not a {@link SSLContext}.
+         * @see JAXRS.Configuration#SSL_CONTEXT
+         * @since 2.2
+         */
+        default SSLContext sslContext() {
+            return (SSLContext) property(SSL_CONTEXT);
+        }
+
+        /**
+         * Convenience method to get the secure socket client authentication policy.
+         * <p>
+         * Same as if calling {@link #property(String) (SSLClientAuthentication)
+         * property(SSL_CLIENT_AUTHENTICATION)}.
+         * </p>
+         *
+         * @return client authentication mode, e. g. {@code NONE}.
+         * @throws ClassCastException
+         *             if sslClientAuthentication is not a
+         *             {@link SSLClientAuthentication}.
+         * @see JAXRS.Configuration#SSL_CLIENT_AUTHENTICATION
+         * @since 2.2
+         */
+        default SSLClientAuthentication sslClientAuthentication() {
+            return (SSLClientAuthentication) property(SSL_CLIENT_AUTHENTICATION);
+        }
+
+        /**
+         * Creates a new bootstrap configuration builder instance.
+         *
+         * @return {@link Builder} for bootstrap configuration.
+         * @since 2.2
+         */
+        static Builder builder() {
+            return RuntimeDelegate.getInstance().createConfigurationBuilder();
+        };
+
+        /**
+         * Builder for bootstrap {@link Configuration}.
+         *
+         * @author Markus KARG (markus@headcrashing.eu)
+         * @since 2.2
+         */
+        static interface Builder {
+
+            /**
+             * Builds a bootstrap configuration instance from the provided property values.
+             *
+             * @return {@link Configuration} built from provided property values.
+             * @since 2.2
+             */
+            Configuration build();
+
+            /**
+             * Sets the property {@code name} to the provided {@code value}.
+             * <p>
+             * This method does not check the validity, type or syntax of the provided
+             * value.
+             * </p>
+             *
+             * @param name
+             *            name of the parameter to set.
+             * @param value
+             *            value to set, or {@code null} to use the default value.
+             * @return the updated builder.
+             * @since 2.2
+             */
+            Builder property(String name, Object value);
+
+            /**
+             * Convenience method to set the {@code protocol} to be used.
+             * <p>
+             * Same as if calling {@link #property(String, Object) property(PROTOCOL,
+             * value)}.
+             * </p>
+             *
+             * @param protocol
+             *            protocol parameter of this configuration, or {@code null} to use
+             *            the default value.
+             * @return the updated builder.
+             * @see JAXRS.Configuration#PROTOCOL
+             * @since 2.2
+             */
+            default Builder protocol(String protocol) {
+                return property(PROTOCOL, protocol);
+            }
+
+            /**
+             * Convenience method to set the {@code host} to be used.
+             * <p>
+             * Same as if calling {@link #property(String, Object) property(HOST, value)}.
+             * </p>
+             *
+             * @param host
+             *            host parameter (IP address or hostname) of this configuration, or
+             *            {@code null} to use the default value.
+             * @return the updated builder.
+             * @see JAXRS.Configuration#HOST
+             * @since 2.2
+             */
+            default Builder host(String host) {
+                return property(HOST, host);
+            }
+
+            /**
+             * Convenience method to set the {@code port} to be used.
+             * <p>
+             * Same as if calling {@link #property(String, Object) property(PORT, value)}.
+             * </p>
+             *
+             * @param port
+             *            port parameter of this configuration, or {@code null} to use the
+             *            default value.
+             * @return the updated builder.
+             * @see JAXRS.Configuration#PORT
+             * @since 2.2
+             */
+            default Builder port(Integer port) {
+                return property(PORT, port);
+            }
+
+            /**
+             * Convenience method to set the {@code rootPath} to be used.
+             * <p>
+             * Same as if calling {@link #property(String, Object) property(ROOT_PATH,
+             * value)}.
+             * </p>
+             *
+             * @param rootPath
+             *            rootPath parameter of this configuration, or {@code null} to use
+             *            the default value.
+             * @return the updated builder.
+             * @see JAXRS.Configuration#ROOT_PATH
+             * @since 2.2
+             */
+            default Builder rootPath(String rootPath) {
+                return property(ROOT_PATH, rootPath);
+            }
+
+            /**
+             * Convenience method to set the {@code sslContext} to be used.
+             * <p>
+             * Same as if calling {@link #property(String, Object) property(SSL_CONTEXT,
+             * value)}.
+             * </p>
+             *
+             * @param sslContext
+             *            sslContext parameter of this configuration, or {@code null} to use
+             *            the default value.
+             * @return the updated builder.
+             * @see JAXRS.Configuration#SSL_CONTEXT
+             * @since 2.2
+             */
+            default Builder sslContext(SSLContext sslContext) {
+                return property(SSL_CONTEXT, sslContext);
+            }
+
+            /**
+             * Convenience method to set SSL client authentication policy.
+             * <p>
+             * Same as if calling {@link #property(String, Object)
+             * property(SSL_CLIENT_AUTHENTICATION, value)}.
+             * </p>
+             *
+             * @param sslClientAuthentication
+             *            SSL client authentication mode of this configuration
+             * @return the updated builder.
+             * @see JAXRS.Configuration#SSL_CLIENT_AUTHENTICATION
+             * @since 2.2
+             */
+            default Builder sslClientAuthentication(SSLClientAuthentication sslClientAuthentication) {
+                return property(SSL_CLIENT_AUTHENTICATION, sslClientAuthentication);
+            }
+
+            /**
+             * Convenience method for bulk-loading configuration from a property supplier.
+             * <p>
+             * Implementations ask the passed provider function for the actual values of all
+             * their supported properties, before returning from this configuration method.
+             * For each single request the implementation provides the name of the property
+             * and the expected data type of the value. If no such property exists (i. e.
+             * either the name is unknown or misspelled, or the type does not exactly
+             * match), the {@link Optional} is {@link Optional#empty() empty}.
+             * </p>
+             *
+             * @param <T>
+             *            Type of the requested property value.
+             * @param propertiesProvider
+             *            Retrieval function of externally managed properties. MUST NOT
+             *            return {@code null}.
+             * @return the updated builder.
+             * @since 2.2
+             */
+            <T> Builder from(BiFunction<String, Class<T>, Optional<T>> propertiesProvider);
+
+            /**
+             * Optional convenience method to bulk-load external configuration.
+             * <p>
+             * Implementations are free to support any external configuration mechanics, or
+             * none at all. It is completely up to the implementation what set of properties
+             * is effectively loaded from the provided external configuration, possibly none
+             * at all.
+             * </p>
+             * <p>
+             * If the passed external configuration mechanics is unsupported, this method
+             * MUST simply do nothing.
+             * </p>
+             * <p>
+             * Portable applications should not call this method, as the outcome is
+             * completely implementation-specific.
+             * </p>
+             *
+             * @param externalConfig
+             *            source of externally managed properties
+             * @return the updated builder.
+             * @since 2.2
+             */
+            default Builder from(Object externalConfig) {
+                return this;
+            }
+
+        }
+    }
+
+    /**
+     * Handle of the running application instance.
+     *
+     * @author Markus KARG (markus@headcrashing.eu)
+     * @since 2.2
+     */
+    public interface Instance {
+
+        /**
+         * Provides access to the configuration <em>actually</em> used by the
+         * implementation used to create this instance.
+         * <p>
+         * This may, or may not, be the same instance passed to
+         * {@link JAXRS#start(Application, Configuration)}, not even an equal instance,
+         * as implementations MAY create a new intance and MUST update at least the
+         * {@code PORT} property with the actually used value. Portable applications
+         * should not make any assumptions but always explicitly read the actual values
+         * from the configuration returned from this method.
+         * </p>
+         *
+         * @return The configuration actually used to create this instance.
+         * @since 2.2
+         */
+        public Configuration configuration();
+
+        /**
+         * Initiate immediate shutdown of running application instance.
+         *
+         * @return {@code CompletionStage} asynchronously shutting down this application
+         *         instance.
+         * @since 2.2
+         */
+        public CompletionStage<StopResult> stop();
+
+        /**
+         * Result of stopping the application instance.
+         *
+         * @author Markus KARG (markus@headcrashing.eu)
+         * @since 2.2
+         */
+        public interface StopResult {
+
+            /**
+             * Provides access to the wrapped native shutdown result.
+             * <p>
+             * Implementations may, or may not, have native shutdown results. Portable
+             * applications should not invoke this method, as the outcome is undefined.
+             * </p>
+             *
+             * @param <T>
+             *            Requested type of the native result to return.
+             * @param nativeClass
+             *            Requested type of the native result to return.
+             * @return Native result of shutting down the running application instance or
+             *         {@code null} if the implementation has no native result.
+             * @throws ClassCastException
+             *             if the result is not {@code null} or is not assignable to the
+             *             type {@code T}.
+             * @since 2.2
+             */
+            public <T> T unwrap(Class<T> nativeClass);
+        }
+
+        /**
+         * Provides access to the wrapped native handle of the application instance.
+         * <p>
+         * Implementations may, or may not, have native handles. Portable applications
+         * should not invoke this method, as the outcome is undefined.
+         * </p>
+         *
+         * @param <T>
+         *            Requested type of the native handle to return.
+         * @param nativeClass
+         *            Requested type of the native handle to return.
+         * @return Native handle of the running application instance or {@code null} if
+         *         the implementation has no native handle.
+         * @throws ClassCastException
+         *             if the handle is not {@code null} and is not assignable to the
+         *             type {@code T}.
+         * @since 2.2
+         */
+        public <T> T unwrap(Class<T> nativeClass);
+    }
+
+}
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -31,6 +31,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <microprofile.config.version>1.3</microprofile.config.version>
     </properties>
 
     <organization>
@@ -267,6 +268,12 @@
             <groupId>javax</groupId>
             <artifactId>javaee-api</artifactId>
             <version>7.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.eclipse.microprofile.config</groupId>
+            <artifactId>microprofile-config-api</artifactId>
+            <version>${microprofile.config.version}</version>
         </dependency>
     </dependencies>
 
--- a/examples/src/main/java/jaxrs/examples/bootstrap/BasicJavaSeBootstrapExample.java
+++ b/examples/src/main/java/jaxrs/examples/bootstrap/BasicJavaSeBootstrapExample.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2018 Markus KARG. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package jaxrs.examples.bootstrap;
+
+import java.net.URI;
+
+import javax.ws.rs.JAXRS;
+import javax.ws.rs.JAXRS.Configuration;
+import javax.ws.rs.core.Application;
+import javax.ws.rs.core.UriBuilder;
+
+/**
+ * Basic Java SE bootstrap example.
+ * <p>
+ * This example demonstrates bootstrapping on Java SE platforms solely using
+ * defaults. It will effectively startup the {@link HelloWorld} application at
+ * the URL {@code http://localhost/} on an implementation-specific default IP
+ * port (e. g. 80, 8080, or something completely different). The actual
+ * configuration needs to be queried after bootstrapping, otherwise callers
+ * would be unaware of the actual chosen port.
+ * </p>
+ *
+ * @author Markus KARG (markus@headcrashing.eu)
+ * @since 2.2
+ */
+public final class BasicJavaSeBootstrapExample {
+
+    /**
+     * Runs this example.
+     *
+     * @param args
+     *            unused command line arguments
+     * @throws InterruptedException
+     *             when process is killed
+     */
+    public static final void main(final String[] args) throws InterruptedException {
+        final Application application = new HelloWorld();
+
+        final JAXRS.Configuration requestedConfiguration = JAXRS.Configuration.builder().build();
+
+        JAXRS.start(application, requestedConfiguration).thenAccept(instance -> {
+            Runtime.getRuntime()
+                    .addShutdownHook(new Thread(() -> instance.stop()
+                            .thenAccept(stopResult -> System.out.printf("Stop result: %s [Native stop result: %s].%n",
+                                    stopResult, stopResult.unwrap(Object.class)))));
+
+            final Configuration actualConfigurarion = instance.configuration();
+            final URI uri = UriBuilder.newInstance().scheme(actualConfigurarion.protocol().toLowerCase())
+                    .host(actualConfigurarion.host()).port(actualConfigurarion.port())
+                    .path(actualConfigurarion.rootPath()).build();
+            System.out.printf("Instance %s running at %s [Native handle: %s].%n", instance, uri,
+                    instance.unwrap(Object.class));
+            System.out.println("Send SIGKILL to shutdown.");
+        });
+
+        Thread.currentThread().join();
+    }
+
+}
--- a/examples/src/main/java/jaxrs/examples/bootstrap/ClientAuthenticationJavaSeBootstrapExample.java
+++ b/examples/src/main/java/jaxrs/examples/bootstrap/ClientAuthenticationJavaSeBootstrapExample.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2018 Markus KARG. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package jaxrs.examples.bootstrap;
+
+import java.net.URI;
+
+import javax.ws.rs.JAXRS;
+import javax.ws.rs.JAXRS.Configuration;
+import javax.ws.rs.JAXRS.Configuration.SSLClientAuthentication;
+import javax.ws.rs.core.Application;
+import javax.ws.rs.core.UriBuilder;
+
+/**
+ * Java SE Bootstrap Example using TLS Client Authentication.
+ * <p>
+ * This example demonstrates bootstrapping on Java SE platforms using HTTPS with
+ * <em>bidirectional</em> TLS authentication (i. e. only <em>trustworthy</em>
+ * clients may connect). It relies mostly on defaults but explicitly sets the
+ * protocol to HTTPS and mandates TLS client certificate validation. The example
+ * will effectively startup the {@link HelloWorld} application at the URL
+ * {@code https://localhost/} on an implementation-specific default IP port (e.
+ * g. 443, 8443, or someting completely different). The actual configuration
+ * needs to be queried after bootstrapping, otherwise callers would be unaware
+ * of the actual chosen port. If the client's certificate is invalid or cannot
+ * be validated, the server will reject the connection.
+ * </p>
+ * <p>
+ * This example uses some basic <em>external</em> JSSE configuration:
+ * </p>
+ * <ul>
+ * <li>{@code javax.net.ssl.keyStore=~/.keystore} - HTTPS: Path to a keystore
+ * holding an X.509 certificate for {@code CN=localhost}</li>
+ * <li>{@code javax.net.ssl.keyStorePassword=...} - HTTPS: Password of that
+ * keystore</li>
+ * <li>Client Authentication: The default truststore
+ * ({@code $JAVA_HOME/lib/security/cacerts}) must hold the root certificate of
+ * the CA and all intermediate certificates used for signing the client's
+ * certificate.</li>
+ * </ul>
+ * </p>
+ *
+ * @author Markus KARG (markus@headcrashing.eu)
+ * @since 2.2
+ */
+public final class ClientAuthenticationJavaSeBootstrapExample {
+
+    /**
+     * Runs this example.
+     *
+     * @param args
+     *            unused command line arguments
+     * @throws InterruptedException
+     *             when process is killed
+     */
+    public static final void main(final String[] args) throws InterruptedException {
+        final Application application = new HelloWorld();
+
+        final JAXRS.Configuration requestedConfiguration = JAXRS.Configuration.builder().protocol("HTTPS")
+                .sslClientAuthentication(SSLClientAuthentication.MANDATORY).build();
+
+        JAXRS.start(application, requestedConfiguration).thenAccept(instance -> {
+            Runtime.getRuntime()
+                    .addShutdownHook(new Thread(() -> instance.stop()
+                            .thenAccept(stopResult -> System.out.printf("Stop result: %s [Native stop result: %s].%n",
+                                    stopResult, stopResult.unwrap(Object.class)))));
+
+            final Configuration actualConfigurarion = instance.configuration();
+            final URI uri = UriBuilder.newInstance().scheme(actualConfigurarion.protocol().toLowerCase())
+                    .host(actualConfigurarion.host()).port(actualConfigurarion.port())
+                    .path(actualConfigurarion.rootPath()).build();
+            System.out.printf("Instance %s running at %s [Native handle: %s].%n", instance, uri,
+                    instance.unwrap(Object.class));
+            System.out.println("Send SIGKILL to shutdown.");
+        });
+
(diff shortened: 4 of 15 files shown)


jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+interface JAXRS {
+
+    /**
+     * Invoked in Java SE environments to start the provided application at the
+     * specified root URL.
+     *
+     * This method will not return until the JAX-RS application is terminated.
+     *
+     * @param rootURI
```
chkal: Is it actually common to provide an `URI` to configure the endpoint? I'm not against it, I just didn't see this before.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * <em>This method is not intended to be invoked by applications. Call
+     * {@link JAXRS#start(URI, Application)} instead.</em>
+     *
+     * @param rootURI
+     *            The root URI to which the application will be bound.
+     * @param application
+     *            The application to start up.
+     */
+    default void bootstrap(final URI rootURI, final Application application) {
```
chkal: I'm not sure if I like having this on the `JAXRS` interface itself. Maybe we should add some other interface for that?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * The {@code JAXRS} class is available in a Jakarta EE container environment as
+ * well; however, support for the Java SE bootstrapping APIs is <em>not
+ * required</em> in container environments.
+ *
+ * @author Markus KARG (markus@headcrashing.eu)
+ *
+ * @since 2.2
+ */
+interface JAXRS {
```
jansupol: Do we have a better name than JAXRS? A `Bootstrapper`, or an `ApplicationFactory` is more descriptive perhaps? Should this be an Interface? Abstract class could have `#bootstrap` method protected, if do not want to call it outside from `#start`.

I'd slightly prefer a `RuntimeDelegate` method such as `#createApplicationFactory()`, to avoid another static method.

mkarg:
I do not think that this particular PR has to be withhold until more Microservice APIs are discussed. See, the aim is to have a vendor-neutral API for bootstrapping JAX-RS, not more, not less. Yes, it is useful for Microservices. But it also is useful without any that. In fact, we are using Jersey-onley Microservices currently, and this is the single line of code that makes our code Jersey-dependend. I really want to get rid of it, and I do not see why I have to wait just because other useful things are in the queue.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
chkal: No, I would definitely prefer `URI` over `URL`. It's just that most API's I know of don't use `URL`/`URI` but some other way to specify all the required parameters for bootstrapping.

The most interesting question is: Which information does an implementation need for bootstrapping?
* A TCP port number
* Which interface to bind to
* Which protocol to use (HTTP/HTTPS/AJP)
* Socket Timeouts?
* TLS certificate to use for HTTPS? (just an example. Most likely not relevant for most apps)

Most if this can be represented with an `URI`. But maybe implementations would like to support advanced options. In this case the API should provide a way to pass these options.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
chkal: I like `JAXRS`. And I agree with @mkarg that it is consistent with `JAXB` and `Persistence`,

ggam: I think current Java API trend is to only uppercase the first letter (see Jsonb). That would leave as with Jaxrs.

Don't have a preference. Just a thought to discuss.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: Well, the minimum is the root URI and the Application, obvisouly. Certainly we could add more parameters if needed. Which one are essential and supported by all vendors?

BTW, Jersey is using an URI:
```
    public static HttpServer createHttpServer(final URI uri, final ResourceConfig configuration) {
```
(from: https://github.com/jersey/jersey/blob/43c81e3f307e1d364094d0cba57f7618908e8145/containers/jdk-http/src/main/java/org/glassfish/jersey/jdkhttp/JdkHttpServerFactory.java#L89)

mkarg:
@spericas Can you elaborate a bit what actual risk do you see by standardizing the already existing Java SE bootstrapping APIs without further discussing Microservice APIs? I just cannot see what you mean with "flawed" here, as I did not propose additional features, but just wrapping existing features which only deal with the very initial bootstrapping. What risk do you see? What is flawed here?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
chkal: So the root URI would provide us with the protocol, port and host address. And I agree that this is the minimum. Perhaps we just need some kind of map or properties parameter so that implementations can support advanced options?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
chkal: @ggam Good point.

jansupol:
Couple of thoughts: 

1. The name - Persistence class seems more like JAXRS RuntimeDelegate in functionality. Comparing Jsonb and Jaxrs (in names) feels wrong, it is like comparing a car and an airport, while JSON-B is more a standalone do here and now a single thing, JAX-RS is more a server technology that is about to accept multiple threads/requests on possibly multiple endpoints/resources. Good naming should be by the functionality not by remote similarity in names. 
2. What could be a good technology to compare with is JAX-WS. This is where bootstrapping was thought over back then.  Section 5 of JAX-WS spec brings a lot of ideas, what methods should/could be supported (bootstrap vs. teardown, @PostConstruct, ....)
3. Configuration - The VIs should be given an opportunity to configure the server outside of what the Application/ResourceConfig allows for the application. Jersey is capable of starting the app on top of about 5 frameworks which the user would like to choose between them, The configuration might be passed as a file or as at least as a Builder pattern ( ((VIAppFactory)RuntimeDelegate.getInstance().createApplicationFactory).configure()....bootstrap()). This is a point where config JSR/Microprofile config should be explored for integration.
4. JAX-RS 3.0 should be mandatory dependent on CDI. At that point, should we think about CDI bootstraping the application rather than JAX-RS bootstrapping it?
5. I agree that we should provide API that is mature enough and complete rather then what seems like a first draft. The approach that is undertaken in this group seems like a lot of refactoring will be going on. Would it make sense to crate a separate branch where the new features taken from Microprofile (the new JAXRS client, this bootstrap, config, ...are beta tested first before putting them to a 2.2 / 3.0 branch? 
The big picture of putting everything together will show the changes needed.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: While `Properties` sounds like the vehicle of choice to my ears, it actually was invented for _persistent_ properties and has a focus on loading and storing them. `Map<Object,Object>` would be more lightweight.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: I cannot see a _trend_. This sounds more like personal taste.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
chkal: Sure, I'm also -1 for `Properties`, especially because they just support strings. A map would be much better. I just wanted to list all the options. 

Why do you prefer `Map<Object,Object>` over `Map<String,Object>`? I cannot think of a use case  for having object keys. JPA for example uses the latter and it works quite well.

spericas: Using a URI/URL is very strange. Both are more general than necessary here are not the right abstraction. A host/port combination is standard, using 0.0.0.0 to bind all interfaces (again, something that would look very odd in a URL). We need a set of parameters to configure the service. Portability, which is the goal of this, would require defining standard parameters that every implementation must support. As I said, there's more work than meets the eye here to achieve portability and make this API useful.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
spericas: I don't love the use of the JAXRS name, but I understand the rationale for it. However, why isn't the implementation of the methods in this class not using RuntimeDelegate?

spericas:
@mkarg A lot of my points are already in @jansupol last comment. Elaborating on one of them, bootstrapping a JAXRS app requires a significant amount of configuration (SSL, authn/authz, ports, network interfaces, CDI support, health checks, etc.). 

This is typically larger that what people want to specify/hardwire in code. That is the reason for specs like MP Config. Should JAX-RS define an external source for this config? Not sure we should be in that business, we should instead leverage other specs. Also, back to the microservices use case which is a motivation for this work if I understand this correctly, config can be provided externally: e.g., via config maps in kubernetes. 

In summary, sure we can define an API with a simple URI, but I don't see it being useful beyond hello-world use cases.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: @chkal I simply do not see any need to synthetically restrict the key to `String` as it will work pretty well with any kind of Object. Unless the key must be human-readable (i. e. it solely serves the purpose of technically identifying the value in the map), `Object` is pretty sufficient. So every vendor (and the API itself) can define `static Object`s for the keys. This is, BTW, the best performance you can gain with comparisons, as it will never use character comparions.

chkal: @spericas I agree that binding to 0.0.0.0 would be weird when using an `URI`. However, apart from that it feels elegant to use an `URI`, because it naturally includes the protocol, the port and the base path. But as mentioned above I'm not 100% sold on this idea. I never saw an URI being used in this context, so I'm still in the process to make up my mind if I like it or not. ;-)

chkal: @mkarg Understood. I see advantages of using `Object` as you can for example use enumerations as keys.  But one disadvantage of it is that you are **forced** to use some kind of `static Object` as the key if you configure the runtime. So you cannot read values from config files, environment variables, etc anymore.

chkal:
I agree that how to configure the runtime is one of the key aspects of such an API. There are many possible sources for configuration values we could support (like MP Config). However, I'm not sure if we really should add a dependency on MP config.

I just had a crazy idea I would like to bring up. Imagine a bootstrap API like this:

```java
public interface JAXRS {
    static void start( Application app, Map<String,Object> config ) { ... }
}
```

This would allow the developer to provide a configuration map just like when configuring JPA or other frameworks. I think this is really common. But what about if the developer could also provide a function which resolves configuration values:

```java
public interface JAXRS {
    // [...]
    static void start( Application app, Function<String,Object> provider ) { ... }
}
```

This could provide a nice and clean way to integrate with all possible configuration APIs. Just provide a function to delegate to your config source of choice and you are done.

```java
JAXRS.start( myApp, key -> ConfigProvider.getConfig().getValue(key, Object.class) );
```

Just an idea. ;-)

quipsy-karg:
+1 for providing a map; 0 for provider functions: the bootstrap is called exactly once per application, so there is no real benefit of the additional complexity, and the application programmer MUST write a rather complex intermediate lambda anyways to filter which properties he wants to set or not.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
quipsy-karg: @spericas I had chosen `URI` because Jersey uses that, see https://jersey.github.io/apidocs/latest/jersey/index.html. But I am also happy with `Map`.

@chkal The question is whether persisting config is _our_ problem? If someone wants that, he can simply write some `load` method which replaces _any_ text by predefined singletons. This is very beneficial as it would throw an "unknown configuration option" right at the time of parsing the text file, way before the app even tries to bootstrap the runtime.

chkal:
@mkarg:

> he bootstrap is called exactly once per application, so there is no real benefit of the additional complexity

Sure, but having to read each configuration property from an external source just to populate a map which you then pass to the bootstrap method is ugly boilerplate you have to repeat for each app. And also, which complexity? The complexity of having an additional bootstrap method with anther signature?

> and the application programmer MUST write a rather complex intermediate lambda anyways to filter which properties he wants to set or not.

No, you won't need that. The function is called by the JAX-RS implementation only for support configuration keys. So there is no need to do any filtering.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
chkal: @quipsy-karg Sure, persisting isn't our problem. But we we should keep the API simple for real world use cases. And loading config property from external sources is a real world use case.

quipsy-karg:
@chkal I am not against what you propose, I just am not finally convinced that it is really better.

mkarg:
I tried to adopt as many as request changes as I could. Please feel free to comment. I hope you like it better now. :-)

Regarding compatibility with MP Config I'd rather go with @chkal's idea of callbacks to abstain from any direct reference to any particular technology.

Regarding CDI and 3.0 I'd like to have Java SE bootstrapping ASAP, i. e. in 2.2.

spericas:
The use of a config provider in the form of a lambda is definitely a step forward, but the API still feels a bit clanky to me (think about actual implementations of those config providers). If we are going to define properties for compatibility, which we have to, why not use the builder pattern like in the Client API for more natural COC and ease of use?

Here is how I envisioned this bootstrap API:

`
Config c = Config.builder()
    .host("0.0.0.0").port(8080)
    .useCdi(true).build();
Service s = Service.builder()
    .application(MyApplication.class)
    .config(c).build();
s.start().then(Main::postStart);
`
Also, the use of a config builder will allow integration with "other config sources". This feels more modern and also better aligned with recent APIs like those in the MP.

quipsy-karg:
When using the builder pattern, an implementation of JAX-RS runtime cannot add _custom_ properties (like Jersey: the type of HTTP engine to use either JDKHTTP or Grizzly for example) -- he will be limited to the properties we define upfront in the API. Also when someone wants to simply load the config from a text file or from the MP config API he has to write rather complex glue code instead of simply passing key-value pairs.

spericas:
That's not true, you can always have a generic property(key, value) in your builder if necessary (but of course, if you use that you cannot guarantee compatibility which is the original motivation for the API). As for the implementation for loading a config, you're just pushing that to the application by providing a lambda (config provider).

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @since 2.2
+ */
+public interface JAXRS {
+
+    /**
+     * Invoked in Java SE environments to start the provided application at the
+     * specified root URL.
+     *
+     * This method will not return until the JAX-RS application is terminated.
```
andymc12: How is a JAX-RS application terminated?  Would you want to add a new `stop()` or `close()` method to the `Application` class?  Alternatively, maybe the `JAXRS.start` method could return immediately after starting the application asynchronously, and then you could add a `JAXRS.stop` method to terminate the application?

mkarg: My PR only proposed to wrap existing vendor-specific technology under a common API hood. AFAIK not all JAX-RS runtimes do support explicitly stopping an application. So your proposal would be to add a completely new feature. If all vendors support that, I am fine with it. But until then, I cannot unilaterally set up a PR for that.

So question to @jansupol, @asoldano, @deki: Do you like to add support for explicitly stopping a JAX-RS application as proposed by @andymc12? I would be more than happy to add this to my PR in that case.

mkarg:
Ok, if we add `property(key, value)` I'm fine with a builder and I will be happy to provide a third draft with the next commit. OTOH I'd like to ask all committers what they like best, the simple key-value approach of the current second draft, or the more sophisticated (but also more complex) approach of a builder including a property(key, value) method? Just to make sure that the next draft is supported by the majority of committers.

chkal:
+1 for using a builder with a generic `.property(key, value)` method for custom configuration options. I agree with @spericas that this is the best and easiest way for users to configure the runtime.

mkarg:
FYI: Had been on the road last week. Resuming work on 3rd draft including builder pattern now. Stay tuned. :-)

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
andymc12: Hi @mkarg - I'm mainly having trouble understanding how the `start()` method works - or more specifically how the method ever returns.  And possibly my brain is malfunctioning because I've been so deep in the Java EE (server-side) mode for too long that I can't process Java SE patterns.  ;-)  

Here are a couple scenarios where I think a blocking `start` method with no spec-defined way to stop the app could be troublesome:
1) User wants an app to read from a database.  In their `main` method, they set up some database connections then call `start` to load up the JAX-RS front-end.  At some point, they will want to stop the process (perhaps to deploy a new version of the app, or do DB maintenance, etc.).  They could just kill the JVM process, but that won't gracefully disconnect from the DB - not a huge problem, but still messy.

2) User wants to test their app before deploying it to a Java EE server.  They set up mock resources (like WireMock or MockServer, etc.) that run in separate processes.  Once their testing is complete, the user would need to kill not only their Java SE JAX-RS process, but any other processes that it may have started - instead of leveraging the convenient APIs for stopping those processes.

I'm not sure if CXF has a way to stop an application (outside of the runtime it is running in - Tomcat, OpenLiberty, etc.) so it might not have a mechanism to return from this method as it is today.  That's not necessarily a problem - but then we should probably update the spec to indicate that implementations should provide a mechanism for stopping apps.  

Hope this helps...

mkarg: @andymc12 I share your thoughts, but again, I am only trying to wrap the current behaviour of the existing products as-is. The start method will only be implemented IF the existing products CAN be started on Java SE (this is AFAIK not mandatory as of JAX-RS 2.1+). IF an existing product hasn't a way to be stopped BUT supports Java SE, then effectively effects 1 and 2 will happen actually, independent of my proposed API or not -- again, I am not specifying a new feature or behaviour, I only wrap EXISTING products. So again my question to all vendors: Can your existing products be stopped explicitly, yes or no? If yes, I can add a stop mechanism in this API. If no, are you willing to change this?

deki: In think in most cases ShutdownHook is used in this case and CXF is shutdown by closing the Spring context. Should be also possible to implement it for non Spring scenarios, org.apache.cxf.endpoint.Server has stop and destroy methods. Comparing to JAX-WS javax.xml.ws.Endpoint has also a stop method, so I'm open to have something similar in JAX-RS.

mkarg:
Please comment my third draft of the Java SE bootstrap interface. It now contains a builder interface and a shutdown command.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,248 @@
...
```
spericas: The new proposal looks much better Markus. Here are a few comments:

- I believe start/stop should be async so that apps can do other inits during startup on main thread. I would use CompletionStage<Instance> as return value. That way you can also do,

 JAXRS.start(app, conf).thenAccept(instance -> ...)

- I'm not sold on the name "Instance". Is "Server" better?

mkarg: I see your point and in fact had the same vision first, but then I abstained from it, because:
* If I am writing any Java SE program that shall do multiple things at the same time, it is my job to provide the needed parallelization plumbing, as there are virtually no APIs in the Java SE world that return CompletableStages as of today, and I simply could not find a good argument why JAX-RS is so special that it should provide a CompletionStage as the first and only one. If someone wants continuations, he can simply wrap the synchronous start() method by CompletableFuture.supplyAsync. So the sole benefit would be that a JAX-RS implementation would not need to synthetically wait for the end of the bootsrap. That means OTOH if the majority of existing implementations would be async, certainly it would make sense to simply disclose the internal stage.
* "Server" came in to my mind first, but isn't it a bit too broad? I think it would be not concise enough, as with "Server" people might imply "Java EE Server", "Web Server", "Servlet Server", etc. It is just an overstressed term. "Instance" (in the sense of "Application Instance") seems to be more accurate, as in fact, the Java SE programmer does not start a Server, he starts a local application instance.

Anyways, I have no strong feelings about the terms. Committers, what do you think? We simply could vote. Here is mine:

async: 0
Instance: +1
Server: 0

mkarg:
I would be more than happy to receive more comments -- or even review approvals in case you think we're done! :-)

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
spericas: Since nobody else is providing feedback, here are some additional comments ;)

(1) I don't the follow your argument about about CompletionStage. Lots of APIs use it, in fact we use it in JAX-RS already :) This is about your application's main, which will likely need to do other things in addition to starting JAX-RS. Given that the start method is non-trivial, it makes perfect sense to make it async in this context.

(2) I feel that CDI support is such an important use case that I'm wondering whether the builder should support it. I realize this is somewhat orthogonal to JAX-RS, but practically, very useful if I can start my JAX-RS server with CDI enabled.

mkarg: (1) What I meant to say is that Java SE 8 (= the sole target of my proposal) does not contain other APIs that produce CompletableFuture, so ALL other work besides starting JAX-RS has to be manually wrapped by the application programmer anyways. Hence there is no benefit. We just spare him one sole line of code. Anyways, I am not against it.

(2) CDI is optional in JAX-RS 2.1 and not mentioned in Java SE 8 at all, but it might be a good idea to simply add a method like `.cdi(SeContainer javaseCdiContainer)` to pass the Java SE 8 CDI container initialized by the `main()` method of the standalone bootstrap code? The JAX-RS implementations can then further configure that container.

pavelbucek: (1) I like the CompletionStage idea and the start being asynchronous, but .. on which thread it would run? Shouldn't we have a threadpools-related method in Configuration class then? (it's somehow similar to RxInvokerProvider in this context) 

(2) I believe the CDI bootstrap routine would be little more complex than that and single builder method might not be enough.

From what I recall, CDI does require all classes / providers to be registered before the container is initializer (ready to serve beans). That might actually require completely different kind of API if there is a need to make it part of the "jax-rs" initializer.

(Obviously, that depends on how the CDI integration would be defined for this particular usecase, but I believe it should be possible to inject any application-scoped instances which JAX-RS runtime creates to any other "app-scoped" bean).

mkarg: I leave it up to @spericas to answer on that, as it was *him* who wanted both. I just reacted upon his ideas and can live without `CompletionStage` *and* CDI. Again, we do this for 2.2, which does *not* mandate CDI.

OTOH if it is really impossible to add classes / providers to an existing `SeContainer` then we could simply do `.cdi(ContainerInitializer)` which allows that. But before we discuss the actual solution it might be good to ask someone who is really deep into Java SE CDI initialization... @struberg Can you please help us?

spericas: Hi @pavelbucek. On (1) we can add a method to config and possibly use ForkJoinPool.commonPool() by default. I'm not sure I follow the issue on (2), are you thinking about a JAX-RS/CDI issue or a Jersey/CDI issue?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,248 @@
...
+ * environments.
+ * <p>
+ * The {@code JAXRS} class is available in a Jakarta EE container environment as
+ * well; however, support for the Java SE bootstrapping APIs is <em>not
+ * required</em> in container environments.
+ * </p>
+ * <p>
+ * In a Java SE environment an application is getting started by the following
+ * command using default configuration values (i. e. mounting application at
+ * {@code http://localhost:80/}:
```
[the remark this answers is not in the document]
mkarg: On the other hand, virtually all public services run on 80 by default, as it is the default port of HTTP. 8080 is a workaround on Linux that I do not need on Windows. So shall we default on Linux-workarounds or wouldn't it be better to offload that decision to the application programmer? I mean, that's why a default that can be overridden easily.

mkarg:
Pushed draft number four producing `CompletionStage`s.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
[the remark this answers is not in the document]
pavelbucek: Where is the line between what has to and could be provided for the container to be started?

I mean.. host, port.. makes sense, but it doesn't need to be here, right? (host can be localhost by default and port can be any open port from some defined range - shouldn't that be also supported?)

what about thread pool sizes and threading model in general?
what about https, http/2 and related configuration?

Shouldn't we create examples (see https://github.com/eclipse-ee4j/jaxrs-api/tree/master/examples/src/main/java/jaxrs/examples) as it used to be done? (https is a standard requirement and should be part of that example/spec itself if this api should be portable)

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,248 @@
...
+ * JAXRS.Configuration config = JAXRS.Configuration.builder().protocol("HTTP").host("0.0.0.0").port(80).rootPath("api")
+ * 	.build();
+ * JAXRS.start(new MyApplication(), config);
+ * </pre>
+ *
+ * @author Markus KARG (markus@headcrashing.eu)
+ *
+ * @since 2.2
+ */
+public interface JAXRS {
```
pavelbucek: what about thread pool sizes and threading model in general?
what about https, http/2 and related configuration?

Shouldn't we create examples (see https://github.com/eclipse-ee4j/jaxrs-api/tree/master/examples/src/main/java/jaxrs/examples) as it used to be done? (https is a standard requirement and should be part of that example/spec itself if this api should be portable)

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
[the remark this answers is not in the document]
pavelbucek: I agree with @chkal - `80` doesn't seem to be a good default - it won't run on many environments.

shouldn't there be a functionality which would pick any open port from defined range? Also, the instance should have getters for everything (?) which was passed as a configuration - and extendable part for container to talk back to user of this api - it could expose statistics and other advanced runtime info that way. (that would also solve the issue of getting the "randomly" selected port).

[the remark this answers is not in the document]
spericas: * I also think 8080 would be a better default. However is less of an issue today with containers.
* I'd leave dynamic port selection to Docker, K8s, etc. if necessary
* +1 on getters

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
spericas: Examples would be nice :)

mkarg: @pavelbucek *Which* thread pools do you want to configure in a Java SE environment? A general one for JAX-RS (besides common pool) or do you see a need to configure additional pools (which ones)?

Threading models are not covered by JAX-RS specification IMHO so if an implementation supports that, custom config parameters have to be used, e. g. `builder.property("org.eclipse.jersey.threadingmodel", "...")`.

"HTTPS" is enabled using `.protocol("HTTPS");` as mentioned in the description of `JAXRS.Configuration.PROTOCOL`. HTTP/2 is AFAIK not yet standardized in JAX-RS 2.1, so when implemented by a JAX-RS engine, it has to defined a custom value for this property.

Minimum examples are contained in JavaDocs; I will add more examples once we agreed upon the API itself to spare the work to change them every day. ;-)

pavelbucek: ad thread pools - http container can have various threading models and I believe that's what being started under the hood (it's not even hidden, right, since host and port is already specified). Here I thought about accept and worker pools.

ad examples - examples for uses of the new API - please look around in the referenced space, there is plenty of examples of what example mean in this context.

In terms of Java SE, there is no good http container to be used (correct me if I'm wrong) - I would expect this to be pluggable and open for container providers.

https can be magically enabled, but there might be a need to configure certificates - similarly to what can be done in containers, i.e. glassfish if we are staying in java/jakarta ee env. I'm not updated on what is the RI for jakarta ee, but would you expect that this command could start something like gf?

not leaving the https space - this IS touching security, since https can be configured to provide client authentication; should the api allow to set the keystore and truststore? Also, having the POC implementation would help a lot in this case, I'm pretty sure it would reveal potential caveats. 

For me (as an outsider), this seems like a container-like api should be part of this and if that is correct, there might be another set of options which are relevant and important for containers. (The opposite could mean that JAXRS class would be usable only for examples, which wouldn't be enough for being included in the spec if I'd have any saying in that matter).

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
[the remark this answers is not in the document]
pavelbucek: So JAX-RS will state that anything other than docker container env is irrelevant for this API?

Getters would be useful to actually GET the configuration - I can image having some library which would work with an Instance and let's say I want to mention in the log I have received it - what would I get from it? There is no name, id or anything I could use to make the message good (not saying useful).

You could mention that if I want to do that, I should store the builder/configuration I used to create the instance - but why would I do that? Especially when Instance represents running instance, which does have some properties which clearly are managed by that class but are not exposes at this moment.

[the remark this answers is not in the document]
spericas: @pavelbucek I don't think that by not supporting dynamic port ranges we are saying Docker is the only relevant environment. What I think we are saying is that it is and will be a very popular environment by the time we finish with this version of JAX-RS and it already solves that problem. A Bootstrap API like this would be very desirable in that very same environment.

As I said, I like the idea of getters if anything for completeness sake. It is a small and trivial addition too.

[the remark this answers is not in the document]
pavelbucek: Sure, the dynamic port range was mainly point for having getters, I'm not so keen of having something like that defined in the API. On the other hand, it could be nice to still have the possibility to implement something like that - which should be possible even now, since it can be defined via custom properties.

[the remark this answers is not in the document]
chkal: Regarding ports: I don't think that dynamic port ranges make much sense. This shouldn't be the responsibility of JAX-RS. Regarding the default port: Of course port 80 is the default port for HTTP as defined by IANA, but I don't know a single Java EE container which defaults to this port. They all use 8080 and other non-privileged ports by default.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
chkal: I agree with @pavelbucek that TLS typically needs much more configuration. Of course this could be done using the generic `.property(String, Object)` method in an JAX-RS implementation specific way. 

The question is if we want to support detailed TLS configuration in the API. Especially because in many real world scenarios the TLS connection is terminated before the request reaches the container (so TLS termination happens in the load balances or some other component) so that the container itself just needs an HTTP port. Playing devils advocate here. Not sure which way I prefer.

mkarg:
Contributors: If there is someone who likes to contribute several nice examples using this new API or has fun in writing tests, docs or reference implementations, please speak up! My forked repo is open to your PRs! :-)

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: @pavelbucek Thanks for your points. I will update the proposed API this week and try to cover the issues you mentioned. Having said that we have to be careful to only cover features actually supported by current implementations. For example, I do not know if all implementations support pluggable containers as IMHO this is not mentioned by the JAX-RS specification. So if e. g. only Jersey supports that but not RESTeasy then it makes no sense to have an API for that other than `.property("org.eclipse.jersey.container", ...providerclass...)` because the app using it must not rely on this feature. I'd even dare to say that most Java SE application will not care about the actual container engine but trust the JAX-RS implementation to select one automatically.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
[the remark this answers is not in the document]
mkarg: Yes all Java EE containers start on 8080 because they are relicts from an ancient time where you had a full-blown OS underneath and a full-blown EE server. The idea behind this new API is to *not* have both, but to write a small Java SE `main()` that simply pulls up JAX-RS standalone in a micro-environment (such like docker). In such environments you do not have the mentioned restriction, as systems like docker map ports and users IDs automatically. So even when using root in the container you will not be root but a normal user actually, so even when using 80 in the container you will not be on 80 but anything beyond 1K actually.

Anyways, I do not want to stand in the way, but those not wanting 80 must now answer the question: What port to use when protocol is set up HTTPS? Not 443 I assume.

rogergl:
I would like to work on the examples.

mkarg:
>I would like to work on the examples.

@rogergl Great! :-) Please send a PR against my forked repo https://github.com/mkarg/jaxrs-api/tree/mkarg-javase-bootstrap and keep it rebased while this API develops. I think it would be best to start with an at-most example using solely defaults, and then add additional examples describing more complex scenarios like starting in background, shutting down, etc. If you like I can send you some starting point I already have set up but not yet pushed?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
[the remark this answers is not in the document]
chkal: Well, I wouldn't call it ancient to install a Java EE server on a server without wrapping it in a container. Maybe containers are the future, but non-container deployments aren't the past (yet). IMO one of the big advantages of being able start a JAX-RS runtime from a `main()` method is that you can start your server directly from the IDE without the need to set up anything. And this would fail with a default port of `80` (maybe except for Windows?). So defaulting to non-privileged ports still makes a lot of sense. BTW: IANA defines `8080` as an alternative HTTP port. So it is absolutely OK to use it from this perspective.

Regarding TLS: I think that using `8443` is a de-facto standard for non-privileged TLS ports.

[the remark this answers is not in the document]
omnijava: +1 on the use of 8080 and 8443
From an academic perspective there is a long history of recognizing 8080 and 8443 especially on a Windows platform for Java server/services access. We run servers that support IIS/Apache and GlassFish simultaneously for teaching purposes. The same holds true for our Linux servers.

omnijava:
Its rare that I get to look under the hood. I think of myself as a tool user rather than a tool maker. As such there are a number of concepts that are new to me. I have already commented on keeping 8080 and 8443 as the default ports.

Sending and receiving secure messages should be the goal. TLS must be the norm and transparent. At the same time I feel that the code to manage the conversation between a client and service should be machine generated with the option of some customization. I teach my students that constructing a message or reply is a lot more fun then stringing up the wires. 

A question about the examples. Why is the target anything other than localhost:8080? I see jaxrs.examples.org or examples.org without port numbers. Shouldn’t they all be localhost:8080? This will make it easier to search and replace on all examples.

As for the API, here I will defer to developers who use the API to build real applications. In teaching an introduction to web services as I am doing now we touch on only a very small part of the API. I will continue to go over it for anything that strikes me as odd assuming I understand it

rogergl:
@mkarg I think some starting points would be very helpful.

mkarg:
@rogergl I will send you a small example, stay tuned.

mkarg:
@rogergl I have uploaded a minimum example here: https://gist.github.com/mkarg/a38a68f6025f1ef6ddb4916022bd150d.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,269 @@
...
+     */
+    public interface Instance {
+
+	/**
+	 * Shutdown running application instance.
+	 *
+	 * @return {@code CompletionStage} asynchronously shutting down this application
+	 *         instance.
+	 */
+	public CompletionStage<Void> stop();
```
jansupol: I think it would be useful to have`public CompletionStage<? extends Object> stop();`
so that the vendor can return `CompletionStage<Void>` or `CompletionStage<whatever_the_server_stop_returns>`, for instance it can be `CompletionStage<HttpServer>`.

mkarg: The idea behind my `Instance` API object is to be a vendor-independent wrapper around the actual "native server object" (SPI object) of the implementation. For example, with Grizzly-based http servers, `Instance` will internally hold a reference to `HttpServer` (see Jersey-based [*Proof-of-Concept*](https://github.com/mkarg/jersey/commit/d577dee28a598a40e64ca92677b2a17a7da22ca4#diff-8a2241aa5d325442a0b56d0ce85322bbR112)). So my counterproposal for my next iteration of this draft:
* Adding `Instance::unwrap(Class<T> nativeClass): T` which allows to get the wrapped "native server object" at any time (even long before `stop` was called).
* `stop` produces not `Void` anymore but `Instance` again.

But as @spericas asked [in a different thread](https://gist.github.com/mkarg/a38a68f6025f1ef6ddb4916022bd150d#gistcomment-2599500) to *not* return `CompletionStage` at all when stopping ...
> @mkarg Looks good to me. After further review, perhaps we should drop the async behavior of stop(), it seems much less useful that that of start(). Here is the example I had in mind:
JAXRS.start(...).thenAccept(instance -> {
    System.out.println("Instance has started " + instance);
    Runtime.getRuntime().addShutdownHook(new Thread(() ->  instance.stop()));      // sync stop
    try { Thread.currentThread().join(); } catch (Exception e) {};    // blocks for signal
}).toCompletableFuture().get();

...I wonder *which way* we want to finally go? I personally really do like the `Instance::stop(): ComplectionStage<Instance>` solution most, as it looks familar to what we do at starting. @andymc12 you brought up the request for stopping servers, what do you think?

mkarg:
I am currently working on the fifth draft of this PR and plan to push it next week. It will contain *updated* [initial examples](https://gist.github.com/mkarg/a38a68f6025f1ef6ddb4916022bd150d), and there will be an *updated* [Jersey-based Proof-of-Concept](https://github.com/mkarg/jersey/commit/d577dee28a598a40e64ca92677b2a17a7da22ca4#diff-8a2241aa5d325442a0b56d0ce85322bbR112).

As with the first four drafts, I will try to implement as much of your great proposals. But there is one thing I'd like to mention:
* Several of you asked me to change the port default of HTTP from 80 to 8080. Several of you asked me to change from `URI` to distinct parameters. But when authoring the Jersey-based PoC (see above) I noticed that the former JAX-RS RI Jersey actually internally has a default of 80 (*not* 8080) and is using `URI` (*not* distinct parameters). My PR is only a wrapper around existing technology. It would be most straightforward to keep 80 and URI. If we insist on 8080 and distinct parameters, this enforces rewriting of some Jersey code. Is it worth that? Do we really want that change so badly? Or can we live with 80 and URI for the sake of being really only a wrapper around existing code? And are CXF, RESTeasy, etc. already using 8080 and URI *really*? **Everbody's short statement please. :-)**

rogergl:
I’m fine with 80 as the default port as long as the API allows to change the port easily. Since default settings will interfere with the environment no matter what the default is.

omnijava:
Changing the port to 8080 from 80 is not a critical issue. However, it should be a topic for discussion going forward. If we are deferring a change because a particular implementation uses port 80 then why does the API even mention a port. Shouldn’t it be port agnostic and, as Roger wrote, be easy to change. I always cringe when I see literals

mkarg:
@omnijava @rogergl Users can always _explicitly_ set the port using this API. I only plea for keeping the *default* at 80 because we (= the JAX-RS **API** project) should concentrate on finding the common sense of _existing_ products, but abstain from enforcing *new or changed* features in other owner's _products_ without a real _need_. There is existing user code in field that already depends on that existing defaults of that exists product. If we now define a standard interface for these existing products, we should try not to break too much existing code, neither the user code, nor the products. Certainly we can agree to change things, but we must always keep in mind that even a small change in our API produces n times the costs at users and at vendors.

mkarg:
@rogergl (1) The target audience of _that_ particular example you refer to is primarily **you** to demonstrate how you can *actually* run the code *right now*, and secondarily the vendors of Jersey, CXF, RESTeasy, etc so they have a proof-of-concept that the API I proposed actually will be feasible. The target adience of the examples _you_ will provide is vendors of Jersey, CXF, RESTeasy etc. in short term, and users of JAX-RS API in the long term. (2) There had been a (mostly finished) discussion and agreement how the asynchronous parts of the boostrap API shall look like, and we (the people contributing to this API here in this PR discussion) decided weeks ago that it is best to go with the Java SE Standard API for this, which is definitively CompletionStage and nothing else. The reason is that one can combine stages coming from different aspects of the Java SE microservice to run very complex parallel initialization scenarios. The API is definitively not intended for beginners, it is intended to replace product-specific (and far more complex) APIs. Your proposal (just like my initial one -- see the original commits of this PR) is too naive: It won't work well in real-world microservices scenarios. (3) The Jersey-specific properties are contained solely to demonstrate *to you* how you can make your examples actually execute on Java SE. These two custom properties are needed because Jersey _at the moment_ has neither automatic selection of HTTP Server Providers (but supports lots of different providers) nor a common way to shutdown HTTP Servers across different providers (I will soon file a PR on that against the Jersey repo so then both properties will be gone). Hence, for enabling *you* to run the code (you asked for that, remember?) I had to add these two properties _for now_. These two properties _are not_ intended to stay there in the long term.

rogergl:
@mkarg Thanks for the explanation. That helped a lot.

omnijava:
Your explanation is sound.

Ken

rogergl:
I have a question regarding the startup of multiple applications:

Looking at the following example (from JAXRS.java)  I have the question, if it is really intended that two applications can have the same port ? And if so, why would I want to do that ? Different thread pools ?

BTW: CompletableFuture.allOf(firstInstance, secondInstance) is probably missing some toCompletableFuture calls.

```
AXRS.Configuration.Builder configBuilder = JAXRS.Configuration.builder().protocol("HTTP").host("0.0.0.0").port(80);
 JAXRS.Configuration firstConfig = configBuilder.rootPath("firstApp").build();
 JAXRS.Configuration secondConfig = configBuilder.rootPath("secondApp").build();
 CompletionStage<Instance> firstInstance = JAXRS.start(new MyFirstApplication(), firstConfig);
 CompletionStage<Instance> secondInstance = JAXRS.start(new MySecondApplication(), secondConfig);
 CompletableFuture.allOf(firstInstance, secondInstance).join(); // Wait for start of both
```

arjantijms:
Hi,

On Sun, May 27, 2018 at 12:24 PM, Roger Gilliar <notifications@github.com>
wrote:

> The longer I look at the following example the more I ask myself if this
> isn't to complicated ?
>
>  JAXRS.start(new HelloWorld(), JAXRS.Configuration.builder()
>                 .property(ServerProperties.HTTP_SERVER_PROVIDER,
>                         (BiFunction<URI, ResourceConfig, HttpServer>) GrizzlyHttpServerFactory::createHttpServer)
>                 .property(ServerProperties.HTTP_SERVER_ANNIHILATOR, (Consumer<HttpServer>) HttpServer::shutdownNow)
>                 .build()).thenCompose(instance -> {
>             try {
>                 System.out.println("Press any key to shutdown.");
>                 System.in.read();
>                 return instance.stop();
>             } catch (final IOException e) {
>                 throw new CompletionException(e);
>             }
>         }).toCompletableFuture().join();
>
>
It already looks a tad simpler when you write it like this:

JAXRS.start(new HelloWorld(),
            JAXRSbuilder()
                .property(HTTP_SERVER_PROVIDER,
GrizzlyHttpServerFactory::createHttpServer)
                .property(HTTP_SERVER_ANNIHILATOR, HttpServer::shutdownNow)
                .build())
                .thenCompose(instance -> {
                    try {
                        System.out.println("Press any key to shutdown.");
                        System.in.read();
                        return instance.stop();
                    } catch (IOException e) {
                        throw new CompletionException(e);
                    }})
                .toCompletableFuture()
                .join();

;)

Kind regards,
Arjan

rogergl:
@arjantijms Yes that looks better.

rogergl:
One more question. How is the  @ApplicationPath suppose to interact with the rootPath given by the JAXRS configuration ?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
spericas: @andymc12 I'm fine with keeping stop() async, I think I suggested this at some point.

mkarg:
@eclipse-ee4j/eclipse-jaxrs With rootPath = "/root" and @ApplicationPath("/app") and @Path("/res") one would assume that the final path if http://localhost/root/app/res. Agreed? Counter proposals?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
spericas: @mkarg We shouldn't design the API based on what an implementation can or cannot do. IMO we finish up the API exercise at this point and gather feedback from implementors at a later stage. If we find that implementors struggle with a particular design decision, we could revisit it. The proposed "common denominator" isn't pretty IMO.

mkarg: @andymc12 @spericas @pavelbucek I agree that we should not too much restrict ourselfs. I will provide a generic Result type for `stop()` in fifth draft. Regarding an "API first" approach I need to say that this is exactly what Mike Milinkovic told me shall _not_ happen in the API projects. He wants us to do _products first_, and infer common API later -- which is exactly what I tried with this PR. I understand both of you, and I am all for good APIs, but at least I need to _try_ to follow his rules. ;-)

spericas: @mkarg I did mention getting feedback from implementations. But you cannot "infer a common API" by just looking at one implementation.

chkal:
> Several of you asked me to change the port default of HTTP from 80 to 8080. Several of you asked me to change from URI to distinct parameters. But when authoring the Jersey-based PoC (see above) I noticed that the former JAX-RS RI Jersey actually internally has a default of 80 (not 8080) and is using URI (not distinct parameters).

I'm still in favor of port `8080` and distinct parameters, even if Jersey handles it differently. IMO port `8080` is the more widely used default port and distinct parameters are much cleaner and easier to understand.

deki:
Me too.

ggam:
I suspect my opinion won't be popular but what about leavint the default
port unspecified? Users have it pretty easy to customize it, and it leaves
an open door to support port ranges.

The Javadoc should clearly state there's *currently* no default port, but
that *may* change on a future release.

Btw, 8080 is my preferred option and the most common one in my experience,
but even then, Liberty runs by default on 9080 and WebLogic does on 7001.

Tomas-Kraus:
Port 80 or any other port lower than 1024 is a problem on many systems. On UNIX you need root privileges to bind such a port. Or at least configure some exception for user to allow him to do so.
That's why 8080 port is so popular in Java world. Only fool will run java as UID 0.
But as usually using it as default has some pluses and minuses.
+ "+" Having 8080 as default is intuitive. People would expect Java HTTP service to be there.
-  "-" If you run multiple services on the same IP, there is a big chance that this port will be already taken

I still think that having default is a good idea and 8080 is the best option as long as javadoc will tell people that it may not be free.

arjantijms:
Hi,

On Thu, May 31, 2018 at 8:04 PM, Christian Kaltepoth <
notifications@github.com> wrote:

> Several of you asked me to change the port default of HTTP from 80 to
> 8080. Several of you asked me to change from URI to distinct parameters.
> But when authoring the Jersey-based PoC (see above) I noticed that the
> former JAX-RS RI Jersey actually internally has a default of 80 (not 8080)
> and is using URI (not distinct parameters).
>
> I'm still in favor of port 8080 and distinct parameters, even if Jersey
> handles it differently. IMO port 8080 is the more widely used default
> port and distinct parameters are much cleaner and easier to understand.
>
Absolutely in favour of 8080 too. That's the one default port that everyone
always expects and frankly the first one I always try.

Kind regards,
Arjan

jansupol:
@ggam I kind of like the idea of unspecified port. The Configuration#getProperty("javax.ws.rs.JAXRS.Port") returns the port where the application was started.

mkarg:
# Abstract #

Fifth draft will default port to 8080. Negative values and values larger than 65535 are reservered for future use (-1 is reserved particularly to enable auto-selection). **If any compliant JAX-RS engine currently is able to do port-auto-selection _for pure JAX-RS 2.1 servers on Java SE 8_, then please point me to the native API JavaDocs on the web. Also I would like to see default port of that _native_ API. Thanks.**

# Justification

**This** API shall cover what is _already existing_ in the current products. The intention solely is to get rid of product-specific APIs _for simple cases_, so with 2.2 hopefully many users can switch their existing code to this new generic API. This does _neither_ include to cover _all_ feature of _all_ products, nor any _future_ features of any product. For discussions on _non-existing_ features, like auto-ports, I would kindly ask you to do that in a _separate_ issue, as I really doubt it will make it in 2.2's time frame.

Some existing code is using Jersey native API, and relies on _that_ default port. It is absolutely acceptable that we say 8080 shall be the default for this new _generic_ API, but it would not be acceptable that Jersey its native API from 80 to 8080. As a result, the programmer has to take good care that he will get 8080 when changing from native API to generic API! I think we could accept that. But we must not change Jersey's _native_ defaults in Jersey 2.27 -- this can only be done in Jersey 3.0 as it breaks existing native code. Hence, at least for Jersey, that product soon will support **two different** defaults: 80 for _native_ API, 8080 for _generic_ API. Certainly this API is not tailored around Jersey, but I wanted to tell you this problem as an example what happens if we discuss solely our wishes and personal likes instead of looking at the existing products. THIS project -as all API projects of EE4J- is intended to _not_ discuss _product_ chages, but only to find common sense for APIs around **existing** products. While I still think this situation is totally weird and foils the EF's policy, I accept the democratic vote and set default 8080 in draft number five.

I am -1 for _not_ having a default port. The intention of this API is to make startup code as small as possible, using CoC. Explicit ports _must_ be an exception case. Having _auto-selected_ port is incompatible with existing products, hence foils the EF's policy to cover existing tools by APIs instead of writing APIs for non-existing features. Please all add auto-selection of ports in your products first and discuss adding auto-selection to this common API _afterwards_. Also, it is assumed most of the use cases for this API is definitively docker containers. For docker, not only it is irrelevant with which user ID and port a container runs (both is mapped anyways), but it is very hard to set up a docker container for an auto-selecting service. Hence, even if whe some fine day add auto-selection (maybe in 2.3 or 3.0), still we _must_ have a default port _now_. And no, I do not accept that an application author _must_ provide the port explicitly again and again, as clearly the majority of you is pro 8080 as this whole discussion shows.

chkal:
> I suspect my opinion won't be popular but what about leaving the default port unspecified?

This is an interesting option. In most cases users will want to configure the port explicitly anyway, so defining the default port to be implementation specific MAY be a good option.

chkal:
@mkarg I agree with everything you write. Especially that we should provide sensible defaults for all configuration options. However, the TCP port is such a central configuration property, that I would be fine with leaving the default unspecified if there is a disagreement about whether to choose `80` or `8080`, because in real world projects users will explicitly set it anyway.

mkarg:
@chkal There is no disagreement, as I was the only one who wanted 80. I just wanted to tell all the 8080-lovers what they have to live with as a result of their counterproposals. Not specifying a default is a no-go as docker cannot deal with dynamic ports.

omnijava:
I know it may be a minor issue but setting the default to 8080 makes it easier to teach in a polyglot environment. Our students do .Net and PHP on servers listening to port 80 and then do Java web app and web services on 8080.

chkal:
> Not specifying a default is a no-go as docker cannot deal with dynamic ports.

I don't think it is no-go just because of this. We are just talking about the default port. So if you don't explicitly set the HTTP port (which IMHO is uncommon) AND if you are switching your JAX-RS implementation THEN you MAY need to update your Docker port mapping. I think this would be acceptable.

However, if we agree that 8080 is a good choice for the default port, we should go with it.

ggam:
Convention over configuration is important, but I doubt delaying the default port discussion to a future version does any real harm.

Note that my proposal was not to support automatic selection of ports, but *leaving an open door* for that by not committing to a specific port right now. Every Java EE application server starts on an unspecified port and that hasn't been a problem as far as I know.

My point is that, once a default port is defined, we won't be able to change it later. So I personally prefer to leave it unspecified so vendors have flexibility to experiment with different configurations. The next JAX-RS version will revisit it to see whether it's beneficial to define a common port or vendors have found interesting cases for the unspecified one.

mkarg:
>>Not specifying a default is a no-go as docker cannot deal with dynamic ports.
>I don't think it is no-go just because of this. We are just talking about the default port. So if you don't explicitly set the HTTP port (which IMHO is uncommon) AND if you are switching your JAX-RS implementation THEN you MAY need to update your Docker port mapping. I think this would be acceptable.

Yes, this would be acceptable, but only in case the port is known _without_ starting booting such a container at least once, and only as long as the port does not change between restarts. Otherwise it would be impossible to use such a Microservice in elastic scenarios: How should Kubernetes or Swam or even Docker-Compose etc. know to which container-internal port to map the common static external port of each container instance...? Yes, we do not write a docker API, but we also cannot ignore the fact that the majority of JAX-RS Microservices (which are the main target of this API still) will undoubtly be containerized.

> However, if we agree that 8080 is a good choice for the default port, we should go with it.

For the sake of democracy I already accepted default 8080, while still being far from agreeing that it as a _good_ choice. Anyways, let's get past this so I can hack it in finally. ;-)

mkarg:
@ggam I will not +1 for any API that leaves any part of it open for further discussion or change. This is not just some fancy product; this is the JAX-RS API global standard and we have to be a bit more reliable than that. I do not see any good reason what should be better with tomorrow's discussion than with today's, as TCP ports are here since 30+ years...

mkarg:
# Fifth Draft
I have pushed the fifth draft of this API to my repository. Also the PoC (Jersey 2.28-SNAPSHOT) is updated accordingly. Please post your feedback as usual.

I gave thought to all your comments and here is **what I made** from it so far:

* PORT parameter has no specific default anymore. In docker environments the port now has to be explicitly given so it can be statically mapped. If not explicitly given, the implementation is free to select any port (including static defaults, auto-selections, range-scan, etc.). No particular method of port selection is mandated. (PoC is using 80 for HTTP and 443 for HTTPS)
* Instance.configuration() returns the actual configuration (could be live, too, but PoC does not support that)
* Configuration has getters for protocol, host, port, rootPath and custom properties
* stop() returns CompletionStage<StopResult> (PoC makes no use of that other than returning a dummy object)
* StopResult.unwrap() returns native stop result (PoC makes no use of that other than returning null)
* Instance.unwrap() returns native handle (PoC returns the actually used backend HTTP server instance, e. g. the Grizzly HttpServer instance)
* Minimum example demonstrates auto-selected port (separate commit) (PoC: activate one of the commented Jersey backends to see it in action; otherwise it will fail due to missing JAX-RS implementation)

I decided to **not** add specific support for the following of your ideas:

* Support for range scans - At least Jersey, the former RI, apparently does not support this in the context of Java SE bootstrapping (maybe I am wrong then please point me to the existing code) and other vendors did not answer on my question, so my assumption is that other JAX-RS implementations do not support that either. If I am wrong here then please vendors speak up, so we can look into this in the sixth draft.
* Thread pools, TLS, backend plugins - Can already be configured using property(name, value). The PoC demonstrates how to switch backend plugins by a custom property.
* Larger contexts (like booting GlassFish and other full-stack servers) - Out of scope of this PR, as the suggested API is explicitly for booting JAXRS only. IMHO an API for botting full-stack servers has to be discussed with the EE4J PMC.
* HTTP/2 is not supported by JAX-RS at all. We should open a separate PR once the JAX-RS vendors have support implemented for that, so we can see how a common API could look like. After that, we can add this to the then-existing Java SE Bootstrap API.
* More Examples - @rogergl is working on that and I don't want to interfere with him, so I just added a minimal example.
* Anything else - Either I forgot it (then please remember me) or I did not understand it (then please explain to me).

Any comments welcome. :-)

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,375 @@
...
+ * <pre>
+ * JAXRS.start(new MyApplication(), JAXRS.Configuration.builder().build());
+ * </pre>
+ * <p>
+ * The following example shows how to override default values for protocol,
+ * host, port and root path:
+ * </p>
+ *
+ * <pre>
+ * JAXRS.Configuration config = JAXRS.Configuration.builder().protocol("HTTP").host("0.0.0.0").port(80).rootPath("api")
```
chkal: We could add a few newlines here, so that each builder step is in a separate line. This makes it much easier to read.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,375 @@
...
+         * <p>
+         * Same as if calling {@link #property(String) property(PROTOCOL)}.
+         * </p>
+         *
+         * @return protocol to be used (e. g. {@code "HTTP")}.
+         * @throws ClassCastException
+         *             if protocol is not a {@link String}.
+         * @see JAXRS.Configuration#PROTOCOL
+         */
+        default String protocol() {
```
chkal: I'm not sure if this was already discussed, but do we want to use `procotol()` or `getProtocol()`?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
```diff
@@ -0,0 +1,375 @@
...
+
+            /**
+             * @return {@link Configuration} built from provided property values.
+             */
+            Configuration build();
+
+            /**
+             * Sets the property {@code name} to the provided {@code value}.
+             * <p>
+             * This method ignores unknown names and invalid values.
```
chkal: > This method ignores .... invalid values.

This sounds like you won't get any error for calling something like `.property(HOST, BigDecimal.ZERO)`. Is that correct?

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: I will replace the complete description once the API is stabilized.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: *Getter* and *Setter* (in contrast to *Accessor* and *Mutator* - unfortunately much too often treated as synonyms in literature, while it is actually *not* a synonym) is a paradigm of Java *components* API aka "Beans" (*component property* method, vs. *class property* method), which is commonly understood as "returning *the same* that was given by `set`". For our use case this paradigm does not fit at all, as our configuration API allows to e. g. put "-1" and pull "8443" in turn, which confuses people used to "Beans" (in contrast to "Pojos"). Hence, I stick with the pure OOP naming variant, i. e. without `get`.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: Yes, the API does not mandate a compliant implementation to check for validity of values this point in time. The reason is that when I coded the RI ontop of Jersey I noticed that for some properties it is not clear what a "valid" value is at that time, because of dependencies between the properties. For example, `protocol("HTTPS")` enforces TLS; so they need a `SSLContext`. But at the time of calling `protocol("HTTPS")` it is not set, so we cannot check it *here*. Some backends do such complex checks only *very very late* (I had to debug deep into Netty to learn it) and it makes no sense to have only 10% checked right here, while 90% are checked at time of actual server startup. Hence I decided not to demand such checks *here*. I thought about *build()* to do that, but even that would not work with Jersey+Netty, at least without substantial rewrite of Netty, Jetty, Grizzly, and so on to get publicly available checks *before* the actual server boot.

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
mkarg: *UPDATE* regarding TLS: Sixth API draft is half way done, supporting "dead-simple but completely safe" cross-backend TLS configuration. Works like a charm already using Netty, Jetty, Grizzly, SDK and SimpleFramework HTTP backends with Jersey 2.28-SNAPSHOT. Hope to publish the code in the next days. Stay tuned. :-)

spericas:
@mkarg I really like the latest draft, good job! I'm not totally sold on the need to unwrap non-standard types but I could live with it. On the topic of Jersey, you should push any changes to the new JAXRS_2_2 branch on that repo (I have a PR submitted with some 2.2 changes for it).

mkarg:
# Sixth Draft
I have pushed the sixth draft of this API to my repository. Also the PoC (Jersey 2.28-SNAPSHOT) is updated accordingly. Please post your feedback as usual.

Thanks for all the ancouraging comments! In this draft I solely added the follwing:

* There now is a named constant `AUTOSELECT_PORT` to better indicate the special port value `-1` introduced in the fifth draft. I think a named constant is always better to understand code than just a numeric value.
* TLS is now fully configurable (and the PR against the RI already supports it in all Jersey HTTPS backends besides "Simple" -- ran out of time): Besides simply enabling the protocol `HTTPS` (which effectively lets the JRE decide about certificates, trust stores, TLS protocol version, etc.) the application programmer can explicitly provide `SSLContext` (hence protocol version, trust store, key store, SSL provider, etc.) and explictily decide to enforce mutual authentication (hence reject untrusted clients) using the new `sslContext(...)` and `sslClientAuthentication(...)` properties. This not only gives full control on TLS to the application programmer, but also opens the possibility for an application to write its own (e. g. database-backed) keystore / truststore / provider / etc. As we pass the complete `SSLContext` object, there is no need for separate parameters, and as the `SSLContext` is provider-generated, this approach even supports provider-specific properties.

TLS configuraton was tested with Oracle JDK 8 + Jersey 2.28-SNAPSHOT + PoC + cURL and worked with Grizzly2, Jetty, Netty, JDK-HTTP, and Simple, so I hope the chosen API for TLS configuration will be well-suited for RESTeasy, CXF, etc.

Any comments welcome. :-)

jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
chkal: I'm not sure if I fully understand. 

Let assume the user calls `.property(HOST, BigDecimal.ZERO)` or `.property(HOST, "I love JAX-RS")` and starts the server afterwards. What's the expected behavior? I would expect the bootstrapping to fail with some error like "Invalid hostname/IP". 

But from the current wording the invalid value is **ignored** by the `.property()` method. If the value is ignored, the configuration would still be configured to use the default `0.0.0.0` and the server will bootstrap without any error.

I hope this makes my point clearer. I think the wording "ignore invalid value" is a bit confusing.

mkarg: I see your point. I should rephrase the Javadocs to say this: `This method does not check the validity, type or syntax of the provided value`.

chkal: Great! This wording is much better!

rogergl:
Considering this simple example: 
```
    public static final void main(final String[] args) {
        final Application application = new HelloWorld();
        final Configuration configuration = Configuration.builder().protocol("HTTP").port(8080).build();
        JAXRS.start(application, configuration).thenAccept(instance -> {
            final Configuration conf = instance.configuration();
            System.out.printf("Instance %s running at %s://%s:%d%s [Native handle: %s].%n", instance,
                    conf.protocol().toLowerCase(), conf.host(), conf.port(), conf.rootPath(),
                    instance.unwrap(Object.class));
        }).toCompletableFuture().join();
}
```

How am I supposed to register a shutdown hook in the case ?

rogergl:
```
 /**
         * Special value for {@link #PORT} property indicating that the implementation
         * MUST use its default port.
         */
        static final int DEFAULT_PORT = -1;

```
@mkarg In the announcement it is called AUTOSELECT_PORT ? BTW: I like AUTOSELECT_PORT more.

rogergl:
I'm not user if this was already discussed, but is it really necessary to allow null values here ?

```
   * @param protocol
             *            protocol parameter of this configuration, or {@code null} to use
             *            the default value.
             * @return the updated builder.
             * @see JAXRS.Configuration#PROTOCOL

```

mkarg:
@rogergl Please discuss code by typing in your comments _in the code view_. This better allows to separate threads. Thanks. Regarding `null`: Without that you won't have any chance to enable defaults again once previously passed a value explicitly.

mkarg:
@rogergl Regarding `AUTOSELECT`: I replaced it by `DEFAULT_PORT` because "autoselect" could be interpreted as "range scan", which is not supported yet. In addition, if in future we do support "range scan", we need a way to say "I do neither want AUTOSELECT (80/443), nor RANGE SCAN, but I want the implementation to do what it did in 2.2 -- which is: DEFAULT_PORT".

mkarg:
@rogergl I do not understand what you need to know:
> How am I supposed to register a shutdown hook in this case ?

rogergl:
@mkarg I would like to know if I can do something like I can do it in Spring:

```
int exitCode = SpringApplication.exit(ctx, new ExitCodeGenerator() {
@Override
public int getExitCode() {
        // return the error code
        return 0;
    }
});
```

mkarg:
@rogergl So what you like to do basically is to compute an exit code? I do not see what sense this makes, as a Java SE 8 JAX-RS application typically is not the only thing having a say in that: Typically you will have more technologoy in the boat that could fail and/or want to contribute to the final exit code. So for _our_ API, it makes more sense to compute the exit code outside of JAX-RS.

mkarg:
Finished several examples outlining most of the new features. Also hacked proof-of-concept for RESTeasy. Stay tuned for Draft _Seven_, which will be published next week (right after I'm back from Java Forum Stuttgart / BTW: Possibly that is the first public talk about JAX-RS 2.2? Would be fun to meet you there!).

omnijava:
Is anyone going to Oracle’s Code One? I’ll be there speaking on using NetBeans in education. I’ll also be speaking on NetBeans at ApacheCon here in Montreal at the end of September.

mkarg:
@omnijava So you like to talk about this new JAX-RS feature at Oracle Code One? Otherwise I wouldn't see any relation to _this_ thread.

mkarg:
Thanks to everybody for the inspiring discussions and great improvements of the original proposal!

The proposal reached a state where no substantial objections existed, the WIP covered the intended target, and the PoC worked fine. There had been no comments on seventh and eighth draft, and no issues had been left unaddressed. Hence, it is ripe for voting.

**All committers please use the Github review tool to approve this PR _within the next two weeks_.**

*Note that accepting this PR does not mean that it is carved in stone. There is much room to add PRs ontop later if we identify a need to extend the API further, as 2.2 is months away still.*

The PoC (Jersey) reached a state where it is near production quality. As soon as the API is accepted, I will propose it for inclusion in Jersey 2.28. Until then, feel free to use the published branch if you want to see the API in action.

mkarg:
Hm... is everybody on vacation, or why is there only a single approval within one week...? ;-)

mkarg:
Having two +1s, and as this PR is under active discussion _since months_, I decided to merge it _today_, even _before_ the end of the _announced_ review period, so I can go on with my work in the Jersey implementation. If anybody _really_ wanted to -1 (he must have a good reason after all these months), please do so, and I will revert the merge then.

Issue #583, opened 2018-02-17
Clarify if Feature implementations can be auto-discovered via @Provider


chkal:
I already brought up this issue on the mailing list in early 2017. Please see [this discussion](http://download.oracle.com/javaee-archive/jax-rs-spec.java.net/users/2017/04/4281.html) for reference.

The javadocs of most providers (like `MessageBodyWriter`, `ContainerResponseFilter`, `DynamicFeature`, etc) explicitly mention that implementations can be automatically discovered if annotated with `@Provider`.

Unfortunately the docs of `Feature` does not mention this, which may lead to the impression, that automatic discovery of `Feature` is not supported. However, automatic discovery works fine in Jersey. And people may expect auto-discovery to work for `Feature`, because it is explicitly supported for `DynamicFeature`.

One argument that came up in the discussion was that `Feature` can be used both on the server and on the client side, but auto-discovery only works on the server. However, I really think that this can be explained in the docs with something like:

> Features implementing this interface MAY be annotated with `@Provider` annotation in order to be discovered by JAX-RS runtime when scanning for resources and providers. Please note that this will only work for server side features. Features for the JAX-RS client must always be registered programmatically.

What do you think?

mkarg:
Still collecting more implementors, stay tuned. :-)

chkal:
No problem. I just wanted to file this as an issue and request for clarification, because back then I just posted on the EG list.

spericas:
A lot of providers can be used in the client and the server, so that's really not a very good argument. I think this is likely an omission; I recall that Feature and DynamicFeature were somewhat of a late addition.

mkarg:
Maybe our project is the wrong place for this issue, as it feels more like a spec issue to me, and the spec will IIRC into a different project. It might be correct to suspend this thread and move it into that other project later?

chkal:
Actually most issues will affect all three repositories: The API, the spec document and the TCK. I guess the question is where the issues which affect multiple repositories should be managed.

Apart from that: AFAIK the issue discussed here just affects the javadocs of the `Feature` interface. So IMO it is the correct place for this issue.

chkal:
@spericas So are you +1 for adding a note about the support for auto-discovery via `@Provider` on the server side to the javadocs of the `Feature` interface? If so, I would love to provide a pull request against `2.2-SNAPSHOT`.

spericas:
[...]

PR #620, opened 2018-04-17
Clarify that Feature supports auto-discovery via @Provider (#583)

chkal:
This pull requests adds a clarification to the API docs of the `Feature` provider interface which describes that auto-discovery via `@Provider` is supported for server side features.

Please see #583 for the corresponding discussion.

As this pull request isn't a API change and has been discussed before, the review period will be one week.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/Feature.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/Feature.java
@@ -29,6 +29,12 @@ package javax.ws.rs.core;
  * for the facility or conceptual domain it represents, such as registering additional contract providers,
  * including nested features and/or specifying domain-specific properties.
  * </p>
+ * <p>
+ * Features implementing this interface MAY be annotated with the {@link javax.ws.rs.ext.Provider &#64;Provider}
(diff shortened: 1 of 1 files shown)


arjantijms:
@chkal Two small grammar comments, otherwise +1 ;)

chkal:
@arjantijms Thanks for the review? Where did you place the comments? I don't see any comments. Or am I missing something here?

arjantijms:
@chkal I see I didn't commit the review, but I now did ;)

chkal:
@arjantijms @spericas: I agree with your comments about adding articles. But actually I just copied the sentence from `DynamicFeature` for consistency. So we should think about adjusting the wording also in the other classes.

https://github.com/eclipse-ee4j/jaxrs-api/blob/38772e3353f5f9159adfa008e4e25566df9ca6a5/jaxrs-api/src/main/java/javax/ws/rs/container/DynamicFeature.java#L31-L33

spericas:
@chkal Sure, we can, but if we don't do it as part of this issue it's fine too IMO.

chkal:
@spericas Agreed. I just updated the pull request as requested.

chkal:
Thanks for reviewing. I just merged this into the `master` branch.

Issue #565, opened 2017-08-29
Clarify Configuration#getClasses() return value


glassfishrobot:
It must be obvious what is the difference between Application#getClasses and Configuration#getClasses.

Configuration version of that method should return current configuration including auto-discovered features etc; it is not clear (from the javadoc) whether it should include root resource classes. Santiagos point is that resource classes shouldn't be included.

chkal:
Hi! I reported this issue back then. Please see [this discussion on the mailing list](https://javaee.groups.io/g/jaxrs-spec/topic/clarification_of/5911666?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,5911666) for details.

I would love to see this one clarified in the javadocs. So, let me sum up the discussion:

* `Application.getClasses()` allows the developer to manually specify root resources and providers.
* `Configuration.getClasses()` corresponds to the "runtime view". It returns classes that are active because they are either discovered automatically by the implementation, are listed in `Application.getClasses()` or were manually added via `Configurable.register()`.

So the question is: 

**Should `Configuration.getClasses()` return root resource classes or not?**

My personal view on this is that `Configuration.getClasses()` SHOULD return root resources. I don't see any good argument why it shouldn't. And that's also the behavior implemented by Jersey.

Any thoughts about this?

mkarg:
@chkal The issue was discussed in the EG but frankly spoken I forgot the outcome. I will dive into the archives. If it bears that the issue is really open for discussion, I will try to get Jersey, CXF and RestEasy into the boat so we can find a concensus.

chkal:
Thanks, Markus. This would be great!

spericas:
This one is bit tricky as it was clearly somewhat open to interpretation. I was originally of the opinion that root resource classes should not be there, but it seems many implementations have adopted the opposite approach. I agree we should look at the implementations and at some point create a TCK for this.

mkarg:
@spericas Do you have updated contact information of CXF and RESTeasy? AFAIK Sergey and Bill are not active anymore. So whom to ask?

chkal:
@mkarg Back then [Bill Burke wrote](http://download.oracle.com/javaee-archive/jax-rs-spec.java.net/jsr370-experts/2016/04/0506.html) that Alessio Soldano (@asoldano) will take over as Red Hat's representative.

asoldano:
@mkarg @chkal , right ;-)

chkal:
[...]

PR #621, opened 2018-04-17
Clarify which components are returned from Configuration#getClasses()

chkal:
This pull requests clarifies which classes are returned from `Configuration.getClasses()`. Especially it states that the resulting set will include root resource classes and that the result of this method will return a complete runtime view which includes auto-detected components.

Please see #565 for the corresponding discussion.

As this pull request isn't an API change and has been discussed before, the review period will be one week. Feel free to comment if you prefer a two week period.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/Configuration.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/Configuration.java
@@ -153,9 +153,10 @@ public interface Configuration {
     public Map<Class<?>, Integer> getContracts(Class<?> componentClass);
 
     /**
-     * Get the immutable set of registered JAX-RS component (such as provider or
+     * Get the immutable set of registered JAX-RS component (such as provider, root resource or
      * {@link Feature feature}) classes to be instantiated, injected and utilized in the scope
-     * of the configurable instance.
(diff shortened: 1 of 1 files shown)


mkarg:
@chkal I think it would be a good idea to explicitly invite @asoldano, @jansupol and @andymc12 as reviewers to have the vendors in the boat. Also it might help to get enough +1 in the reduced review time.

chkal:
@asoldano @jansupol Any chance that you could review this one? Actually the voting period is over and the vote is therefor successful, but it would be great to see your approval nevertheless.

chkal:
I just merged this one now. Thanks everyone for reviewing and voting.

mkarg:
@chkal Won't this PR close #565?

mkarg:
I don't think so. Maybe it is best to review the commits we did so far once the TCK is published.

asoldano:
@chkal sorry for the late comment. Well, we discussed this a bit https://github.com/eclipse-ee4j/jaxrs-api/issues/565 , RESTEasy will adapt ;-)

spericas:
@mkarg We should probably create issues now for compatibility testing. These changes tend to easily fall through the cracks.

Issue #623, opened 2018-04-21
Enabling JDK 8 Doclint Checks


mkarg:
Yes, I know, it is really a pain that tools tell us we are mostly imperfect with writing good code and great docs. But as this project aims at providing a global standard, possibly one of the most influencing one, we should enable Doclint Checks and fix whatever Javadoc tells us.

I prepared WIP PR #624 and already fixed ALL ERRORs.

**I hope some contributor will chime in here and provide fixes for the masses of WARNINGs.** :-)

mkarg:
Thank you, @ggam, for fixing all those nasty JavaDoc WARNINGS! I'm just waiting for the committers to review our joint work before I can merge it finally.

PR #622, opened 2018-04-21
Minor improvements to pom.xml

mkarg:
With this PR I'd like to propose some minor improvements to the pom.xml file:

- Getting rid of `groupId` lines wherevery possible, following a Maven best-practice: Using CoC for default group IDs.
- Correcting the `organization` element, which now says "Eclipse Foundation" not "Oracle" anymore.
- Bumping versions of Maven plugins to catch up with months to years of bug fixes.

PR #624, opened 2018-04-21
Enabling JDK 8 Doclint Checks (#623)

mkarg:
Improves quality of API JavaDocs by enabling JDK 8 doclint checks and fixing all WARNINGs and ERRORs (see issue #623).

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/BeanParam.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/BeanParam.java
@@ -32,9 +32,8 @@ import java.lang.annotation.Target;
  * annotation. For the POJO classes same instantiation and injection rules apply as in case of instantiation
  * and injection of request-scoped root resource classes.
  * </p>
- * <p>
  * For example:
- * <pre>
+ * <PRE>
  * public class MyBean {
  *   &#64;FormParam("myData")
  *   private String data;
@@ -56,8 +55,7 @@ import java.lang.annotation.Target;
  *
  *   ...
  * }
- * </pre>
- * </p>
+ * </PRE>
  * <p>
  * Because injection occurs at object creation time, use of this annotation on resource
  * class fields and bean properties is only supported for the default per-request resource
--- a/jaxrs-api/src/main/java/javax/ws/rs/NameBinding.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/NameBinding.java
@@ -25,7 +25,7 @@ import java.lang.annotation.Target;
 /**
  * Meta-annotation used to create name binding annotations for filters
  * and interceptors.
- * <p>
+ *
  * Name binding via annotations is only supported as part of the Server API.
  * In name binding, a <i>name-binding</i> annotation is first defined using the
  * {@code @NameBinding} meta-annotation:
@@ -77,7 +77,6 @@ import java.lang.annotation.Target;
  *      ...
  *  }
  * </pre>
- * </p>
  *
  * @author Santiago Pericas-Geertsen
  * @author Marek Potociar
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/HttpHeaders.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/HttpHeaders.java
@@ -45,8 +45,8 @@ public interface HttpHeaders {
     public List<String> getRequestHeader(String name);
 
     /**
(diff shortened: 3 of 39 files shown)


ggam:
Thanks a lot Markus! I'll take care of the warnings myself.

mkarg:
@ggam Thanks! Can't wait to review your contribution! :-)

ggam:
@mkarg I already made a PR to your branch (I cannot push directly since I'm
not a repository member).

jaxrs-api/src/main/java/javax/ws/rs/BeanParam.java
```diff
@@ -32,9 +32,8 @@
  * annotation. For the POJO classes same instantiation and injection rules apply as in case of instantiation
  * and injection of request-scoped root resource classes.
  * </p>
- * <p>
  * For example:
- * <pre>
+ * <PRE>
```
chkal: Just wondering: Are uppercase `<pre>` tags enforced by doclint? I'm just asking, because all other HTML elements are lowercase.

jaxrs-api/src/main/java/javax/ws/rs/core/HttpHeaders.java
```diff
@@ -45,8 +45,8 @@
     public List<String> getRequestHeader(String name);
 
     /**
-     * Get a HTTP header as a single string value.
-     * <p/>
+     * <p>Get a HTTP header as a single string value.
```
chkal: I may be too niggling here, but IMO we should either put both the start and end element on a line nor non of them.

So either:

```html
<p>
Foobar
</p>
```

Or:

```html
<p>Foobar</p>
```

chkal: I agree that your PR definitely improves the situation. It's just something that came to my mind while reviewing this.

mkarg:
@eclipsewebmaster I do not understand why ip validation fails. It says @ggam's signed-off-by is invalid, but in fact it looks pretty valid.

mkarg:
@ggam You signed with `Signed-off-by: Guillermo González de Agüero <z06.guillermo@gmail.com>`. Is that the _exact same_ email address you also registered with your Eclipse Foundation account and Github account? Otherwise this would explain why the ip-validation check still fails.

ggam:
@mkarg yes that's the email on both sites. @waynebeaton any idea?

mkarg:
@ggam I did some checks and I need to say that the ip validation check works pretty well as soon as I replace `z06.guillermo@gmail.com` by my own email address. It even works if I write your full name with all umlauts and accents into the sign-off. So it MUST be the wrong address! Please do check once more that you really filed _exactly_ that address with all your paperwork and that this really is the address linked with your `ggam` account on both, Github user account and Eclipse Foundation user account.

@eclipsewebmaster Can you please check the actual reason why the ip-validation of user ggam fails always? Maybe you can see the address actually expected for that user account?

Meanwhile I signed your commit with my address and added both user names. This is pretty legal as the ip check wants EITHER you or me to sign all commits. This seems to work so we can go on with this PR.

mkarg:
@chkal Can you please redo your approval? This is needed as the recent fixing of signed-off-by problems with ggam's contribution invalidated your existing voting.

@jimma I would be more than happy if you would approve this PR. It is an extension to the PR #622 which you approved recently. Thanks. :-)

chkal:
@mkarg Actually I didn't approve the changes yet, I just added the comments. I'll need some time to checkout your changes and verify the generated Javadocs. A few aspects look weird to me (before and after your changes) and I want to verify the resulting HTML just to check that everything is fine (like `@NameBinding` which doesn't contain paragraph tags at all). I'll come back to you ASAP.

jaxrs-api/src/main/java/javax/ws/rs/NameBinding.java
```diff
@@ -25,7 +25,7 @@
 /**
  * Meta-annotation used to create name binding annotations for filters
  * and interceptors.
- * <p>
```
chkal: After removing this `<p>`, the upper sentence isn't in its own paragraph any more. I don't think that this is correct. The rendered page looks weird, because there is no spacing between the first paragraph and the source code any more.

mkarg: I will merge now but ask some rookie volunteer to improve the layout of the HTML.

Issue #627, opened 2018-05-08
Problem with maven central


titusthefox04:
String  ${packaging.type} not resolved create problem in maven central and in sbt project.
 
<img width="235" alt="jax-ws-problem2" src="https://user-images.githubusercontent.com/22959765/39771538-d8038772-52f2-11e8-873c-f556a7dae746.png">

<img width="181" alt="jax-ws-problem" src="https://user-images.githubusercontent.com/22959765/39771537-d7e650a8-52f2-11e8-8891-d31e96076f6b.png">

chkal:
This is a duplicate of #572 and #571.

I think this is actually a sbt bug. Maven (and Gradle?) can handle this dynamic packaging type correctly.

mkarg:
@spericas Do you know why we don't simply use `<packaging>jar</packaging>` instead?

chkal:
AFAIK there is a special Java 9 profile which uses a different packaging type.

See: https://github.com/eclipse-ee4j/jaxrs-api/blob/d2cadd2a5b22b4c7922a3f1d7c5e00506fccdd75/jaxrs-api/pom.xml#L144-L152

mkarg:
Looks like this was done solely to make the OSGi packager (bnd) work as it is so outdated that it cannot handle JDK 9. I don't like that solution. OSGi is not part of neither Jakarta EE nor JAX-RS. I'd rather kick OSGi out right now. But what do others think?

chkal:
> Looks like this was done solely to make the OSGi packager (bnd) work as it is so outdated that it cannot handle JDK 9.

Yeah, looks like this was the reason for introducing the dynamic packing type.

> I don't like that solution. OSGi is not part of neither Jakarta EE nor JAX-RS. I'd rather kick OSGi out right now. But what do others think?

AFAIK the OSGi manifests are required for the Glassfish integration. But I'm not familiar with the details.

I think we should keep the OSGi manifest. It doesn't do any harm and OSGi is still quite popular. Maybe bnd even supports Java 9 today? Perhaps fixing this would simply be a matter of updating the `maven-bundle-plugin`?

spericas:
@mkarg Must have been OSGi. I don't have a problem fixing it to jar.

kwhat:
So, when is 2.2 being released so I can stop dealing with this problem.

mkarg:
[...]

PR #628, opened 2018-05-13
Static packaging type 'bundle' instead of property

mkarg:
Currently used maven-bundle-plugin 3.5.0 seems to work fine with module-info, so apparently there is no more need for separate packaging types on JDK 8- vs. JDK 9+.

Closing #627

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -134,7 +134,7 @@
                         <plugin>
                             <artifactId>maven-surefire-plugin</artifactId>
                             <configuration>
-                                <argLine>--add-modules java.xml.bind</argLine>
+                                <argLine>--add-modules java.xml.bind --add-opens java.ws.rs/javax.ws.rs.core=java.xml.bind</argLine>
                             </configuration>
                         </plugin>
                     </plugins>
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -23,7 +23,7 @@
     <groupId>javax.ws.rs</groupId>
     <artifactId>javax.ws.rs-api</artifactId>
     <version>2.2-SNAPSHOT</version>
-    <packaging>${packaging.type}</packaging>
+    <packaging>bundle</packaging>
     <name>javax.ws.rs-api</name>
 
     <url>https://github.com/eclipse-ee4j/jaxrs-api</url>
@@ -93,10 +93,6 @@
             <activation>
                 <jdk>[9,)</jdk>
             </activation>
-            <properties>
-                <!-- TODO: maven-bundle-plugin doesn't understand module-info yet. -->
-                <packaging.type>jar</packaging.type>
-            </properties>
             <build>
                 <pluginManagement>
                     <plugins>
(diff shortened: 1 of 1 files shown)


mkarg:
Travis fails on JDK 9 due to missing JAXB, which is strange, as this works for me:

```
[WARNING] Tests run: 42, Failures: 0, Errors: 0, Skipped: 3
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 27.573 s
[INFO] Finished at: 2018-05-13T13:26:40+02:00
[INFO] ------------------------------------------------------------------------

Markus@Zauselmaus MINGW64 ~/git/jaxrs-api-master/jaxrs-api (mkarg-packaging-type)
$ mvn -version
Apache Maven 3.5.3 (3383c37e1f9e9b3bc3df5050c29c8aff9f295297; 2018-02-24T20:49:05+01:00)
Maven home: C:\Program Files\apache-maven-3.5.3
Java version: 9.0.4, vendor: Oracle Corporation
Java home: C:\Program Files\Java\jdk-9.0.4
Default locale: de_DE, platform encoding: Cp1252
OS name: "windows 8.1", version: "6.3", arch: "amd64", family: "windows"
```
Any ideas?

chkal:
I just checked this and the build is working fine for me locally too.

But I think that I know what is causing this. The version of the `maven-surefire-plugin` used during the build is different between Travis (`2.21.0`) and my box (`2.20.1`). And there is an important difference between those versions.

From the [plugin page](https://maven.apache.org/surefire/maven-surefire-plugin/java9.html):
* Version 2.20.1 of the plugin added automatically “–add-modules java.se.ee” option to the command line of forked JVMs [...] in order to ease the transition of applications to Java 9.
* From 2.21 onwards the plugin will not add that option: the recommended way of using those modules is to add explicit dependencies of the maintained versions of such libraries.

We should really manage the version of the `maven-surefire-plugin` to get consistent builds between environments.

mkarg:
@chkal You're my here! I will change that.

mkarg:
The problem actually was not as expected (missing java.xml.bind is explicitly mentioned in our POM) but the fact that JAXB wants *reflective* access to JAX-RS. I fixed that in PR #630.

So I rebased *this* PR ontop of #630 and wait until #630 is merged first.

chkal:
@mkarg One question. My impression was that it is actually the failing test which triggered the reflection and therefore caused the failure. So maybe we don't need a new entry in the production JAR at all but just for the test? In this case I would vote for not adding this entry to the `module-info.java` but instead change the configuration of the `maven-surefire-plugin`. Disclaimer: I'm not very with Java modules, so maybe I'm totally wrong on this.

mkarg:
@chkal Yes, you are right! I now replaced general modification of `module-info` by particular opening to reflecting only for `maven-surefire-plugin`. Thanks for pointing this out!

mkarg:
@asoldano @spericas Please redo review as I had to adapt the solution: Only opening to reflection in surefire tests, not generally in module-info.

mkarg:
Merged after 13 days of review already (instead of the default 14 days) because no API is getting changed and the most active committers voted +1.

PR #635, opened 2018-07-07
Run Travis build against more JDK versions

chkal:
Although we will most likely use the Jenkins instance provided by the Eclipse Foundation in the future, we could update our Travis configuration to run the build about all relevant JDKs. Travis recently introduced support for more JDK versions, so I modified the config accordingly.

Please note that while testing the new configuration, I discovered that the build currently does NOT work with the EA versions of JDK11. I'll create a separate issue for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ sudo: false
 jdk:
   - oraclejdk8
   - oraclejdk9
+  - oraclejdk10
+  - openjdk8
+  - openjdk9
+  - openjdk10
 
 install:
   - cd $TRAVIS_BUILD_DIR/jaxrs-api


mkarg:
I do not see that it is a mutual exclusive decision. Jenkins is good for "real" builds, but is rather clumsy. Travis is agile and has the benefit that forks on Github automatically inherit the CI.

chkal:
> I do not see that it is a mutual exclusive decision. Jenkins is good for "real" builds, but is rather clumsy. Travis is agile and has the benefit that forks on Github automatically inherit the CI.

Agreed! 

> Why still supporting oraclejdk?

For JDKs <= 10 we should test against both Oracle JDK and OpenJDK, because there are known incompatibilities between them (as we know from the Jersey build issues). Of course this will change with JDK11, but for now running the build against both major distributions makes sense to me.

mkarg:
@chkal Just in case you missed the review result: You can actually merge this PR now. :-)

chkal:
Thanks everyone! I just merged this!

PR #641, opened 2018-07-25
public UriBuilder.newInstance()

mkarg:
# Intention
Allowing application programmers to use `UriBuilder` created *from scratch*, i. e. prividing *all* parts separately *without* having a valid URI/template to start with.

# Justification
There are situations where it just comes handy to construct URIs from scratch using a builder pattern, i. e. not using valid URI as a starting point. One example is when one wants to format a pretty-printed URI string from the distinct parts provided to the new `JAXRS.Configuration` builder for informational use (like the examples do) or inside of an implementation (like Jersey does). The Javadocs of `UriBuilder` already says `...for building URIs from their components` but this never was true: Until now one couldn't build *from components* but always had to build from a *valid* URI/template and then *modify* components.

# Side Effects
Virtually none, as all code had been there before, but just was `protected`. In case the provided parts are not sufficient to finally form a valid URI, the terminal `build()`'s Javadocs already mandate to throw `UriBuilderException`.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/UriBuilder.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/UriBuilder.java
@@ -67,8 +67,9 @@ public abstract class UriBuilder {
      * Creates a new instance of UriBuilder.
      *
      * @return a new instance of UriBuilder.
+     * @since 2.2
      */
-    protected static UriBuilder newInstance() {
+    public static UriBuilder newInstance() {
         return RuntimeDelegate.getInstance().createUriBuilder();
     }
 


mkarg:
I cannot see a need for that, as all the `build(...)` methods already have this one:
```java
 * @throws UriBuilderException      if a URI cannot be constructed based on the
 *                                  current state of the builder.
```
I assume that it is technically impossible to produce a malformed `URI` instance. At least the Javadocs of all `URI` constructors already say:
```java
  * @throws  URISyntaxException
  *          If the given string violates RFC&nbsp;2396, as augmented
  *          by the above deviations
```

spericas:
@mkarg I understand, but there are new use cases that did not exist before that _implementors_ need to worry about, such as `UriBuilder.newInstance().build()`. Those may or may not be covered by existing checks. Not a deal breaker.

mkarg:
@spericas Implementors were invited to review, but none of them provided a proposal for a phrasing. So I will merge as-is. If a vendor comes up with a proposal for a phrasing, I will be more than happy to +1 his subsequent PR.

PR #648, opened 2018-08-07
New CONTRIBUTING and NOTICE files.

spericas:
These are required files for release process.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,45 @@
+# Contributing to Eclipse Project for JAX-RS
+
+Thanks for your interest in this project.
+
+## Project description
+
+JAX-RS: Java API for RESTful Web Services (JAX-RS) is a Java programming
+language API spec that provides support in creating web services according to
+the Representational State Transfer (REST) architectural pattern.
+
+* https://projects.eclipse.org/projects/ee4j.jaxrs
+
+## Developer resources
+
+Information regarding source code management, builds, coding standards, and
+more.
+
+* https://projects.eclipse.org/projects/ee4j.jaxrs/developer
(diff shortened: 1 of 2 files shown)


spericas:
@mkarg These files were generated using Eclipse's legal doc tool: https://www.eclipse.org/projects/tools/documentation.php

mkarg:
@spericas In that case I don't bother. If it is legally wrong, we can just blame the EMO. ;-)

mkarg:
Thanks for merging this PR. Good to see there is progress. :-)

_Just for sake of our formal rules_ I'd like to mention that officially *all* changes have a *minimum* review period of two weeks. While there are exceptions to the rule, I want to kindly ask to *explicitly announce* any shorter review periods when opening the PR, so everybody knows he has to hurry up if he wants to give a -1. 

While I do not think anybody would have -1'ed this particular case, it is just fair, because they could have been on holidays.

spericas:
@mkarg We are to need a few changes like this to get the release out. If we wait two weeks on each one, it will take very long time. Those files are required by Eclipse, so I don't think our normal rules apply here. Same applies to pipeline changes etc that are necessary for the EE4J_8 release.

spericas:
@mkarg Forgot to add, yes, I will announce that going forward for anything related to the release.

mkarg:
@spericas I agree with all you said. My intention was to make clear *to others* that nobody gets exceptions from our rules, and you simply "forgot" to mention the shortened review period. My personal opinion is that project leads should have the right to override branch protection for cases like this one. Unfortunately the EF admins are rather reluctant in this regard. Waiting for several days just to kick-off the ip validation was really a disaster.

PR #652, opened 2018-08-14
Release Process: GPG Signing & OSSRH Nexus Staging

mkarg:
**This PR implements the needed changes ontop of branch `EE4J_8` to release JAX-RS 2.1.1 using OSSRH.**

When performing `mvn deploy -P release` the following happens:
* Signs artifacts using GPG.
* Publishes to OSSRH (either snapshot repo or staging repo for publication on The Central Repository).

Usage for publishing a release on The Central Repository:
* Step 1: *Manually* perform a release (i. e. create a PR containing the version change and let the committers review it). *I abstain from the Maven Release plugin as it brings us no measurable benefit in this particular project, does not fit in the EF's typical "committers have to vote" policy, and it won't work due to our branch protection.*
* Step 2: Start the "[Publish Release](https://ci.eclipse.org/jaxrs/job/Publish%20Release/)" job on EF's Jenkins.
* Step 3: If the EF / PMC agrees with the release, the EF admins have to manually promote the stated release on OSSRH.

Note:
* There is no dry-run, as this is not needed: Releases end up in a staging area anyways. There is no *automatic* release, as the EF's rules enforce that the EF / PMC agrees to the release.

**Zero Review Period: In exception to our two-weeks review period, this branch gets merged *immediately* once two committers +1'ed it. The reason is that the project manager needs this branch immediately to continue with the release process.**

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -140,6 +140,17 @@
                         <artifactId>maven-release-plugin</artifactId>
                     </plugin>
                     <plugin>
+                        <artifactId>maven-gpg-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>sign-release</id>
+                                <goals>
+                                    <goal>sign</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-deploy-plugin</artifactId>
                         <configuration>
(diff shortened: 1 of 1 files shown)


mkarg:
Tried it out with 2.2.1-SNAPSHOT and the result ended up on OSSRH in the snapshot repo (not in the stagin are which leads to The Central Repository). So it should work with a release version, too.

mkarg:
Yes I wanted to use the parent, but it would add a lot of additional headaches as it contains decisions that conflict with ours. So to not spend much time into solving that trouble, I decided to provide a PR which works without being dependend of the parent POM. Also, the parent POM is under discussion currently, as it is the wrong approach; a BOM would be better. I read that someone wants to provide that, and in fact, I think using that BOM (once it exists) is what we actually need here to get rid of our custom solution.

Until then, I think we should keep our custom solution.

spericas:
@mkarg OK, we can look at this later. BTW, JSONP is relaying on that parent pom.

PR #655, opened 2018-08-26
Free IP port scan using JAXRS.Configuration.FREE_PORT=0

mkarg:
New configuration option `JAXRS.Configuration.FREE_PORT=0` issues scanning for a free port when starting a server using the Java SE Bootstrap API.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
@@ -291,6 +291,14 @@ public interface JAXRS {
 
         /**
          * Special value for {@link #PORT} property indicating that the implementation
+         * MUST scan for a free port.
+         *
+         * @since 2.2
+         */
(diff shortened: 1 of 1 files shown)


mkarg:
This feature was discussed earlier this year and I had the impression there is a majority for supporting it.

Jersey is already able to handle port=0 and will scan for a free IP port, so this is a no-op for Jersey. The reason is that the JRE itself does free IP port scan when given a zero for a server port number. So I assume most (if not all) other JAX-RS implementations will also support it out of the box.

Hence I'd like to propose that we add support for this feature in 2.2.

mkarg:
The word "scan" would be misleading, as (at least on Windows) no "scanning" takes place: The operating system is asked for a free port from a managed range, and simply picks one. "scanning" would imply that some time-consumin *active iteration* over a range is performed, which is not the case.

chkal:
Good point! So lets got with the current name.

PR #660, opened 2018-09-16
Builds on JDK11

### LINKED ISSUE #636: Build the project with JDK11 fails
Currently the build does NOT work when using the latest EA version of JDK11. See [this Travis build](https://travis-ci.org/chkal/jaxrs-api/builds/401181471) for details.

mkarg:
Support for building JAX-RS API on JDK 11

Closes #636

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,11 @@ jdk:
   - oraclejdk8
   - oraclejdk9
   - oraclejdk10
+  - oraclejdk11
   - openjdk8
   - openjdk9
   - openjdk10
+  - openjdk11
 
 install:
   - cd $TRAVIS_BUILD_DIR/jaxrs-api
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -184,9 +184,78 @@
             </distributionManagement>
         </profile>
         <profile>
-            <id>jdk9+</id>
+            <id>jdk11+</id>
             <activation>
-                <jdk>[9,)</jdk>
+                <jdk>[11,)</jdk>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                    <version>${jaxb.api.version}</version>
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
+                    <groupId>javax.activation</groupId>
+                    <artifactId>javax.activation-api</artifactId>
+                    <version>${activation.api.version}</version>
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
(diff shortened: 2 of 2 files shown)


mkarg:
@chkal @spericas I added the profile as we discussed in #636, so now it builds fine on JDK 11. But now one *test* fails (author: @NiklasMehner), because of trouble with Mockito. Any idea?
```
[ERROR] Tests run: 6, Failures: 0, Errors: 6, Skipped: 0, Time elapsed: 0.677 s <<< FAILURE! - in javax.ws.rs.core.EntityTagTest
[ERROR] shouldBeEqualToTheSameInstance(javax.ws.rs.core.EntityTagTest)  Time elapsed: 0.392 s  <<< ERROR!
org.mockito.exceptions.base.MockitoException:
Mockito cannot mock this class: class javax.ws.rs.ext.RuntimeDelegate.
```

chkal:
Maybe we should try to update Mockito to the latest version. Looks like they had quite some issues with JDK11.

chkal:
BTW: Do we really need a separate `jdk11+` profile? Can't we just keep the `jdk9+` profile, remove the surefire `--add-modules` parameter to get rid of the JDK's JAXB classes and add the JAXB dependency to the dependencies? Or am I missing something here?

chkal:
To be honest, I'm not completely sure if having only a `jdk8` and a `jdk9+` profile will work. But it was the first idea I had to fix this issue.

mkarg:
Christian was right, the problem is fixed by upgrading Mockito, so I filed [CQ 17673](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=17673) and upgraded the dependency in the POM. Travis confirms that the build is fine now on Open JDK 11 and Oracle JDK 11.

mkarg:
@chkal  
>I'm still not completely sure if we really need both a JDK9-10 and a JDK11+ profile.

It depends on what we want to actually assert with that test. JAXB was under development between JDK 9, 10 and 11. If you use the JDK11+ profile for JDK 9 and 10 also, you will end up with the *wrong* JAXB version unless you add *separate versions* of the dependency. But that opens two new problems: (a) You need *separate* profiles to do that. (b) You need to know the JAXB versions used by JDK 9 and 10.

So to sum up: I doubt it is worth the attempt, but feel free to open a PR that solves this issues in a simpler way than the current one. :-)

mkarg:
@spericas @chkal Please redo your approval; had to fix a merge conflict of this PR. Thanks.

mkarg:
Thanks for the approvals. I will merge this PR at the end of the two-weeks period, so we will have JDK 11 builds next week (unless the CQ fails, which I doubt)! :-)

mkarg:
EMO approved [CQ 17673](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=17673) :-)

Issue #606, opened 2018-03-27
Clarification on the way to handle various HTTP response codes


NicoNes:
Can the JAVA SSE spec be more clear on how to handle all various response codes possible, the way  https://html.spec.whatwg.org/multipage/server-sent-events.html does ?

The java SSE spec only talks about 503 but what   about others response codes:

**1.** 204

According to WHATWG spec:

> Clients will reconnect if the connection is closed; a client can be told to stop reconnecting using the HTTP 204 No Content response code.

- Does JAXRS SSE impl should interpret this code the same way ?

- If true should we close the SseEventSource and invoke the completion listener/onComplete to notify user about the eventSource completion?

- Should 204 be treated as unrecoverrable error =>  Do we have to notify error callback ? 

**2.** 503

- Do we have to notify error callback wheither the "retry-after" is set or not? Or just when the "retry-after" is not set.

**3.** Response code other than 200, 204 and 503

According to WHATWG spec (https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model):

> Any other HTTP response code not listed here, as well as the cancelation of the fetch algorithm by the user agent (e.g. in response to window.stop() or the user canceling the network connection manually) must cause the user agent to fail the connection.

- Should JAXRS impl do the same and only accept 200, 204 and 503?
- If true  should those response codes be treated as unrecoverrable error =>  Do we have to notify error callback ?

A more general question:

Once SseEventSource is closed should the completion callback be always invoked even if no event consumer has never been triggered ?

Thanks

mkarg:
@NicoNes Can you please a bit more specific what you propose to be changed in the spec?

NicoNes:
You're right it was not that clear.

As it is done by the HTML5 SSE spec, I would like the JAVA SSE spec to be explicit on how SseEventSource must handle various HTTP response codes.

To do so and in order not to be that far form the HTML5 spec,  I suggest to follow its rules as much as possible.

So to be specific, I would like the spec to state that:

- On HTTP 204 response code

```
The SseEventSource must close itself and not try to reconnect.
It must invoke:
- completion handlers since no further events will be received.
```

- On HTTP 200 response code **with Content-Type set to text/event-stream** ( ignoring any MIME type parameters)

```
The SseEventSource must start processing incoming events.
It must invoke:
- event consumers each time a new event is received.
- error consumers when the SseEventSource encounter any uncoverable error. 
[...]

PR #670, opened 2018-09-27
Issue 606: Clarify SSE reconnect/onError/onComplete behavior

andymc12:
This is a Javadoc-only change.  Before we can close issue #606, I would recommend TCK tests to verify this behavior and updates to the spec document.

For this PR, the behavior of the `SseEventSource` implementation depends on the response received from the remote server:
200 (with Content-Type="text/event-stream") - normal operation - expect `onEvent` to be fired when new events are sent.  `onComplete` should be fired once client stops listening or the server indicates no new events.
200 (with missing or a different Content-Type) - error condition - expect `onError` and `onComplete` to be fired.
204 - normal operation indicating server has no events to send.  expect only `onComplete` to be fired.
503 (with Retry-After set to a valid value) - client should retry after specified value - until then, no events should be fired.  
503 (with missing or invalid Retry-After) - error condition - expect `onError` and `onComplete` to be fired.
Any other status code - error condition - expect `onError` and `onComplete` to be fired.

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/sse/SseEventSource.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/sse/SseEventSource.java
@@ -36,31 +36,43 @@ import jakarta.ws.rs.client.WebTarget;
  * Consumer&lt;InboundSseEvent&gt;#accept(InboundSseEvent)} method is invoked on any registered event consumers.
  * <h2>Reconnect support</h2>
  * <p>
- * The {@code SseEventSource} supports automated recuperation from a connection loss, including negotiation of delivery
- * of any missed events based on the last received SSE event {@code id} field value, provided this field is set by the
- * server and the negotiation facility is supported by the server. In case of a connection loss, the last received SSE
- * event {@code id} field value is send in the <code>{@value jakarta.ws.rs.core.HttpHeaders#LAST_EVENT_ID_HEADER}</code> HTTP
- * request header as part of a new connection request sent to the SSE endpoint. Upon a receipt of such reconnect
- * request, the SSE endpoint that supports this negotiation facility is expected to replay all missed events. Note
- * however, that this is a best-effort mechanism which does not provide any guaranty that all events would be delivered
- * without a loss. You should therefore not rely on receiving every single event and design your client application code
- * accordingly.
+ * The {@code SseEventSource} supports automated recuperation from a connection loss, including
+ * negotiation of delivery of any missed events based on the last received  SSE event {@code id} field value, provided
+ * this field is set by the server and the negotiation facility is supported by the server. In case of a connection loss,
+ * the last received SSE event {@code id} field value is sent in the
+ * <code>{@value jakarta.ws.rs.core.HttpHeaders#LAST_EVENT_ID_HEADER}</code> HTTP
+ * request header as part of a new connection request sent to the SSE endpoint. Upon a receipt of such reconnect request, the SSE
+ * endpoint that supports this negotiation facility is expected to replay all missed events. Note however, that this is a
+ * best-effort mechanism which does not provide any guaranty that all events would be delivered without a loss. You should
+ * therefore not rely on receiving every single event and design your client application code accordingly.
  * <p>
- * By default, when a connection the the SSE endpoint is lost, the event source will wait 500&nbsp;ms before attempting
- * to reconnect to the SSE endpoint. The SSE endpoint can however control the client-side retry delay by including a
- * special {@code retry} field value in the any send event. JAX-RS {@code SseEventSource} tracks any received SSE event
- * {@code retry} field values set by the endpoint and adjusts the reconnect delay accordingly, using the last received
- * {@code retry} field value as the reconnect delay.
+ * By default, when a connection to the SSE endpoint is lost, the event source will wait 500&nbsp;ms
+ * before attempting to reconnect to the SSE endpoint. The SSE endpoint can however control the client-side retry delay
+ * by including a special {@code retry} field value in any sent event. JAX-RS {@code SseEventSource} tracks any
+ * received SSE event {@code retry} field values set by the endpoint and adjusts the reconnect delay accordingly,
+ * using the last received {@code retry} field value as the reconnect delay.
  * <p>
- * In addition to handling the standard connection loss failures, JAX-RS {@code SseEventSource} automatically deals with
- * any {@code HTTP 503 Service Unavailable} responses from an SSE endpoint, that contain a
- * <code>{@value jakarta.ws.rs.core.HttpHeaders#RETRY_AFTER}</code> HTTP header with a valid value. The
- * <code>HTTP 503 + {@value jakarta.ws.rs.core.HttpHeaders#RETRY_AFTER}</code> technique is often used by HTTP endpoints as a
- * means of connection and traffic throttling. In case a
- * <code>HTTP 503 + {@value jakarta.ws.rs.core.HttpHeaders#RETRY_AFTER}</code> response is received in return to a connection
- * request, JAX-RS SSE event source will automatically schedule a new reconnect attempt and use the received
- * <code>{@value jakarta.ws.rs.core.HttpHeaders#RETRY_AFTER}</code> HTTP header value as a one-time override of the reconnect
- * delay.
+ * In addition to handling the standard connection loss failures, JAX-RS {@code SseEventSource} behaves differently to various
+ * HTTP response status codes and headers:
+ * <ul>
+ * <li>200 - with <code>{@value jakarta.ws.rs.core.HttpHeaders#CONTENT_TYPE}</code> header of "text/event-stream": This is normal
+ * operation. <code>onEvent</code> is invoked for each event.  <code>onComplete</code> is invoked when there are no more
+ * events. <code>onError</code> is invoked only if a non-recoverable error occurs during processing.</li>
+ * <li>200 - with unsupported or missing <code>{@value jakarta.ws.rs.core.HttpHeaders#CONTENT_TYPE}</code> header: This is an
+ * error condition. <code>onError</code> and <code>onComplete</code> are invoked.</li>
+ * <li>204 - This indicates that server has no events to send. Only <code>onComplete</code> is invoked.</li>
+ * <li>503 - with <code>{@value jakarta.ws.rs.core.HttpHeaders#RETRY_AFTER}</code> header set to a valid value: This indicates
+ * that the server is unavailable, but that the client should reconnect later. No consumers are invoked unless the client
+ * event source is closed, prior to reconnecting (resulting in <code>onComplete</code> invocation). After the specified
+ * delay, the client should automatically attempt to reconnect which will result in a new response.</li>
+ * <li>503 - with invalid or missing <code>{@value jakarta.ws.rs.core.HttpHeaders#RETRY_AFTER}</code> header: This is an error
+ * condition. <code>onError</code> and <code>onComplete</code> are invoked.</li>
+ * <li>Any other status code: This is an error condition. <code>onError</code> and <code>onComplete</code> are invoked.</li>
+ * </ul>
+ * <p>
+ * In the case of an error condition response, the <code>Throwable</code> passed to the <code>onError</code> consumer
+ * <i>should</i>
+ * be a WebApplicationException containing the invalid <code>Response</code> object.
  *
  * @author Marek Potociar
  * @since 2.1


arjantijms:
Hi @andymc12 

>updates to the spec document.

At the moment there's unfortunately no spec document, and nobody knows if the existing JAX-RS spec document will ever be transferred.

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
```diff
@@ -53,14 +53,20 @@
...
- * request, JAX-RS SSE event source will automatically schedule a new reconnect attempt and use the received
- * <tt>{@value javax.ws.rs.core.HttpHeaders#RETRY_AFTER}</tt> HTTP header value as a one-time override of the reconnect delay.
+ * In addition to handling the standard connection loss failures, JAX-RS {@code SseEventSource} behaves differently to various
+ * HTTP response status codes and headers:
+ * <ul>
+ * <li>200 - with <tt>{@value javax.ws.rs.core.HttpHeaders#CONTENT_TYPE}</tt> header of "text/event-stream": This is normal operation. <code>onEvent</code> is invoked for each event.
+ *   <code>onComplete</code> is invoked when there are no more events. <code>onError</code> is only invoked if a non-recoverable error
+ *   occurs during processing.</li>
+ * <li>200 - with unsupported or missing <tt>{@value javax.ws.rs.core.HttpHeaders#CONTENT_TYPE}</tt> header: This is an error condition. <code>onError</code> and <code>onComplete</code> should be invoked.</li>
+ * <li>204: This indicates that server has no events to send. Only <code>onComplete</code> should be invoked.</li>
```
mkarg: Shouldn't `204` cleary imply that the result is not an empty stream but `null`?

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
```diff
@@ -53,14 +53,20 @@
...
+ * <li>503 - with <tt>{@value javax.ws.rs.core.HttpHeaders#RETRY_AFTER}</tt> header set to a valid value: This indicates that the server is unavailable, but that the client should reconnect later.
+ * No consumers should be invoked unless the client event source is closed, prior to reconnecting (resulting in <code>onComplete</code> invocation).
+ * After the specified delay, the client should automatically attempt to reconnect which will result in a new response.</li>
```
mkarg: `SHOULD` or `MUST`?

andymc12:
@arjantijms 
> At the moment there's unfortunately no spec document, and nobody knows if the existing JAX-RS spec document will ever be transferred.

I've heard that too, but I'm still hopeful that we will get it - and be able to make changes to it.  If not, we should consider a separate work item to start drafting a new "Jakarta JAX-RS" spec document.

mkarg:
@andymc12 @arjantijms The probem is not that we couldn't write our own spec. The problem actually is that we simply don't know whether it makes sense to do that. In case we do, and in November the agreement is signed, we've squandered valueable human resources. In case we do not, and in November there still is no agreement, we lost valueable time. A lose-lose-situation!

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
mkarg: As you mentioned the HTML5 spec, can you please point me to the particular chapter where 204 in mentioned (maybe they use it in a specific way for SSE)? Thanks. :-)

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
```diff
@@ -53,14 +53,20 @@
...
- * In case a <tt>HTTP 503 + {@value javax.ws.rs.core.HttpHeaders#RETRY_AFTER}</tt> response is received in return to a connection
- * request, JAX-RS SSE event source will automatically schedule a new reconnect attempt and use the received
- * <tt>{@value javax.ws.rs.core.HttpHeaders#RETRY_AFTER}</tt> HTTP header value as a one-time override of the reconnect delay.
+ * In addition to handling the standard connection loss failures, JAX-RS {@code SseEventSource} behaves differently to various
+ * HTTP response status codes and headers:
+ * <ul>
+ * <li>200 - with <tt>{@value javax.ws.rs.core.HttpHeaders#CONTENT_TYPE}</tt> header of "text/event-stream": This is normal operation. <code>onEvent</code> is invoked for each event.
+ *   <code>onComplete</code> is invoked when there are no more events. <code>onError</code> is only invoked if a non-recoverable error
+ *   occurs during processing.</li>
+ * <li>200 - with unsupported or missing <tt>{@value javax.ws.rs.core.HttpHeaders#CONTENT_TYPE}</tt> header: This is an error condition. <code>onError</code> and <code>onComplete</code> should be invoked.</li>
```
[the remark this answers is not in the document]
mkarg: Sorry I was misleading. Didn't talk about upper/lower case. What I actually wondered is: Shouldn't we replace "...should..." by "...must..."?

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
andymc12: https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events

> Clients will reconnect if the connection is closed; a client can be told to stop reconnecting using the HTTP 204 No Content response code.

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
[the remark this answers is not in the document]
andymc12: No problem - I'll switch to use "must".  Thanks for the review comments.

andymc12:
New updates available:
- Adjusted line lengths to be more consistent with previous javadoc lines (some are still longer than 120, but very close - and no longer than some pre-existing lines).
- Removed usage of "should" when describing which consumers are invoked.
- Added that the implementation _should_ pass a `WebApplicationException` that contains the invalid `Response` object when invoking the `onError` consumer.

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
mkarg: I would propose to instead write:
`204: This indicates that the client must close the connection and must not reconnect. <code>onComplete</code> must be invoked. <code>onError</code> must not be invoked.`

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
```diff
@@ -53,14 +53,27 @@
...
- * as a means of connection and traffic throttling.
- * In case a <tt>HTTP 503 + {@value javax.ws.rs.core.HttpHeaders#RETRY_AFTER}</tt> response is received in return to a connection
- * request, JAX-RS SSE event source will automatically schedule a new reconnect attempt and use the received
- * <tt>{@value javax.ws.rs.core.HttpHeaders#RETRY_AFTER}</tt> HTTP header value as a one-time override of the reconnect delay.
+ * In addition to handling the standard connection loss failures, JAX-RS {@code SseEventSource} behaves differently to various
+ * HTTP response status codes and headers:
+ * <ul>
+ * <li>200 - with <tt>{@value javax.ws.rs.core.HttpHeaders#CONTENT_TYPE}</tt> header of "text/event-stream": This is normal
+ * operation. <code>onEvent</code> is invoked for each event.  <code>onComplete</code> is invoked when there are no more
+ * events. <code>onError</code> is only invoked if a non-recoverable error occurs during processing.</li>
```
asoldano: better changing "is only invoked if" to "is invoked only if"

andymc12:
@asoldano thanks for the review.  I've incorporated the changes from your review comment and from PR #674.  

All, since these changes are very minor, I would like to still plan on merging this after two weeks since my previous change (on September 29) - October 12. Please let me know if you disagree with this plan.

@spericas, @chkal - can you re-review in light of these new changes?

Thanks!

spericas:
@jimma That is a good question. The proposed text suggests that it should not, but I can see valid arguments either way. 

@andymc12 Any reason why we shouldn't call `onError` in this case?

spericas:
@andymc12 @NicoNes I had a similar thought about the `onRetry` callback. I agree that `onError` may be confusing, but so is not informing the application at all. As someone stated, this behavior is specific to JAX-RS, so it would make sense to add a callback for it.

andymc12:
@mkarg - thanks.  I rebased and pushed.  Looks like the ip-validation task passed now.

andymc12:
@spericas @NicoNes @jimma - Sorry for the delay...  just noticed I never pushed the comment button on this...  Would you all be ok with a separate issue/PR for this suggestion?  I think this PR addresses the original issue requesting clarification on the different HTTP status codes.  I think the idea of an `onRetry` callback makes sense, but we may need to discuss it more.  

Particularly, I'm wondering whether this new (proposed) callback should be invoked if the client has connected, and the server has sent an event with the `delay` property specified.  In this case, we would expect `onEvent` to be invoked as the event could contain usable data in addition to the request to re-connect later.

NicoNes:
Hi @andymc12 ,

 I agree to create a separate issue/PR for the `onRetry` callback suggestion.

> Particularly, I'm wondering whether this new (proposed) callback should be invoked if the client has connected, and the server has sent an event with the delay property specified. In this case, we would expect onEvent to be invoked as the event could contain usable data in addition to the request to re-connect later.

I guess you mean `retry` property instead of `delay` right? 
If so, setting `retry` property in an SSE event is just a way for the endpoint to override default (500ms) SSE client-side reconnection delay that should be used after a connection loss.
The purpose of the `retry` property is not to tell to the client to close current connection and retry XXX seconds later.
So I don't understand your point well. Can you please explain ?

andymc12:
@NicoNes 
> I guess you mean retry property instead of delay right?

Right.  The SSE property is retry.   The JAX-RS `SseEvent` interface uses `getReconnectDelay()` to map to that property.  

>If so, setting retry property in an SSE event is just a way for the endpoint to override default (500ms) SSE client-side reconnection delay that should be used after a connection loss.
The purpose of the retry property is not to tell to the client to close current connection and retry XXX seconds later.
So I don't understand your point well. Can you please explain ?
 
We should probably discuss this more in the separate issue once it is created.

My main point is that we need to be clear when the new `onRetry` should be invoked.  Is it any time that the client's default or specified retry delay timer is modified (i.e. both when the client receives a 503 with valid `Retry-After` and when the client receives an event with a retry property)?  Or is it only when it receives a 503 with valid `Retry-After`?  Or is it when the retry is about to occur (for example, if it is 12:00 when I connect and receive a 503 with `Retry-After` set to 12:05 - should we invoke `onRetry` immediately or at 12:05?)?  

This new callback method definitely falls into the category of new function - that's why I'm pushing for a new issue, since this one is primarily intended to clarify existing function.  I think it is a good idea, but I want to make sure we correctly capture the intended behavior.  

Would you mind opening the new issue?  Thanks!

@spericas @chkal - sorry to do this again.  When I pushed the commits again (to trigger IP validation), it wiped away your approvals.  Can you re-approve?  At this point, I think we have addressed all review concerns and have consensus (on the clarification at least - we may have more work to do for `onRetry`, but we can take that in a separate issue/PR).  The actual code changes have been available for over two weeks, so once you approve again, I'll merge.  Thanks!

NicoNes:
Hi @andymc12 

Ok it's clearer now and you're right we have to take time to do things well. 
Sure I will open a new issue to discuss about it :+1: .

Thanks

andymc12:
This PR is in a weird state.  It is failing because Travis is trying to run Java 10 which has expired - this prevents me from merging.  It looks like @mkarg has resolved this in PR #682.  However, if I rebase to pull in those changes, this will once again wipe out all of the reviews.

These rules seem arbitrarily and un-intuitively restrictive.  Can we change them so that either (1) we don't need Travis to pass 100% in order to merge or (2) that rebasing and/or adding minor changes does not wipe away existing reviews?

mkarg:
I have absolutely no problem with the existing rules, and none of them are arbitrary: They had been discussed and democratically voted. All you need is patience. Soon my PR will be merged, and after that yours will be able to merge again. No need to hurry, this is the master branch, so there is no time pressure.

chkal:
I agree that wiped out approvals can be annoying in such cases. But generally it is a good rule. If somebody modifies the PR request, reviewers should check the modifications to see if they still agree. Usually getting the reviewers to approve again takes just a few hours. And I think that broken Travis build is actually some kind of special case.

mkarg:
To amend what @chkal said, I think that troubles like this one will be soon over. Once Jakarta EE 8 is out the project should be a such good shape that we can simply add features and bug fixes frequently without frequent build troubles.

mkarg:
@andymc12 Your build runs well once rebased on `master` due to the included solution #682 (see this experiment on [Travis CI](https://travis-ci.org/mkarg/jaxrs-api/builds/446515594)). So I'd like to ask you to simply do the rebase and push again. BTW, this gives room to squash your three commits into a single one, which helps to keep the history clean. I am sure everybody is fine with re-casting his vote. :-)

andymc12:
done

jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java
jansupol: I strongly agree with @mkarg here. The new javadoc is actually weakens the specification. Any SHOULD can not be tested, and any implementation can behave on its will. 

> the client should automatically attempt

> consumer <i>should</i> be a WebApplicationException

andymc12:
Hi all - I don't really have time to shepherd this PR into master - especially given all of the rebasing/re-IP-checking/re-delivery-for-Java10/re-* with our current delivery process.  I'll leave the PR open, and if somebody is interested in picking it up, they are welcome to take it.

Thanks, and good luck,
Andy

andymc12:
Hey all - only a little over a year later, and I found some time to work on this. :) The only changes from over a year ago were to incorporate the new package name and other changes that occurred during that time - and then to fix the 204 formatting issue that @spericas mentioned.  This should have all of the changes requested by reviewers from the last time around. Can you all take another look and approve?  Thanks!

NicoNes:
Hey @andymc12,

Thank you for reopening the discusssion about this :+1: . 
On my side, I  still agree with the changes you made in the doc.

Thanks !

mkarg:
Seems we need some SSE expert to do a final review on this matter. Or maybe @chkal and @asoldano like to change their vote?

chkal:
I'm not very familiar with SSE and therefore abstain from the vote.

mkarg:
@asoldano I'd kindly like to ask you to check with Red Hat's SSE experts and finally aprove this issue as you announced *in October*. Thanks. :-)

andymc12:
@asoldano I believe I have addressed your previous concerns.  Can you approve this?  Thanks!

mkarg:
Do we need to mention this in the spec, at least in the history appendix?

NicoNes:
I think it would be great to mention it in the history appendix at least.

PR #671, opened 2018-10-01
JAX-RS 2.1.2 (Bug fixes ontop JAX-RS 2.1.1)

mkarg:
This is a proposal to cherry-pick *more* "non-essential" bug fixes from `master` to `EE4J_8` in the form another service release: JAX-RS 2.1.2.

*There had been several voices saying that these bug fixes are wanted in 2.1.2, so it is **up to the committers** to decide whether we actually cherry-pick them or not.*

While technically there are several commits contained, effectively what they produce simply spoken is:
* Will build fine on JDK11 (several commits needed for that, as Maven plugins are failing in EE4J_8)
* JPMS module declaration corrected
* Examples reference Jakarta EE instead of Java EE (existing bug in current EE4J_8 branch)

*As this is only a service release, and as the actual API is not touched we can just push to OSSRH without the need of another review by EMO / PMC. Having said that, I propose we do informally announce our proposal to the PMC.*

**Counter Proposal: If this PR fails the review, I will simply republish `EE4J_8` *as-is*, i. e. in the form of JAX-RS 2.1.1 with changed Maven Coordinates.**

**As the release has to be published before October 22nd, I hereby announce a shortened review period of just five days. This means, I will release EE4J_8 either with or with without this PR next week to match the deadline!**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,11 @@ jdk:
   - oraclejdk8
   - oraclejdk9
   - oraclejdk10
+  - oraclejdk11
   - openjdk8
   - openjdk9
   - openjdk10
+  - openjdk11
 
 install:
   - cd $TRAVIS_BUILD_DIR/jaxrs-api
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -23,7 +23,7 @@
 
     <groupId>jakarta.ws.rs</groupId>
     <artifactId>jakarta.ws.rs-examples</artifactId>
-    <version>2.1.2-SNAPSHOT</version>
+    <version>2.1.2</version>
     <packaging>jar</packaging>
     <name>jakarta.ws.rs-examples</name>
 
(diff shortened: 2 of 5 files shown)


mkarg:
@andymc12 Good point, but if I change that now, we all have to review again... So I will bear the risk of this slight inaccuracy.

mkarg:
After pro-forma filing this service release using the EF's developer tools, this morning I have triggered the JIPP pipeline for EE4J_8 to push 2.1.2 to OSSRH. It will take a few days, because the EF admins apparently had not yet requested permission for the JIPP instance to stage using groupId `jakarta.ws.rs` but only using groupId `javax.ws.rs`. See [Eclipse Bugzilla #539874](https://bugs.eclipse.org/bugs/show_bug.cgi?id=539874). Stay tuned. :-)

chkal:
Thanks @mkarg. Looking forward to the first release with the new groupId. :tada:

PR #676, opened 2018-10-04
Added gitignore entries for Eclipse, NetBeans and VS Code

arjantijms:
Signed-off-by: arjantijms <arjan.tijms@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,22 +1,9 @@
-# maven noise
-target
-
-# osx noise
-.DS_Store
-profile
-
-#IntelliJ Idea noise
-.idea
-*.iws
-*.ipr
+target/
+.classpath
+.project
+.settings/
 *.iml
-
-# TeX build noise
-*.aux
-*.bbl
-*.blg
-*.log
-*.out
-*.tex~
-*.thm
-*.toc
+*nbactions.xml
+.vscode/
+.idea
+nb-configuration.xml


mkarg:
Arjan, can you please file a PR just with the *project-specific* noise reduction? I would be fine with that one. Thanks.

chkal:
@mkarg The current `.gitignore` file already contains [entries for IntelliJ](https://github.com/eclipse-ee4j/jaxrs-api/blob/c0220093ec48fbb80fea823e6d3ea9e7e4c9e9a1/.gitignore#L8-L12). So would it really be so bad to also add ignore rules for Eclipse?

mkarg:
@chkal Yes because it makes a bad thing even worse. I in fact I'd rather propose to even remove the existing IntelliJ entries. Such clutter has nothing to do with our project and it is simply a wrong usage of git: That shall go into the user's or machine's `.gitignore` so it has effect upon *all* projects, instead of adding an (always growing and never comprehensive) list of IDE-specific stuff in *all* projects again and again.

chkal:
@mkarg I don't fully agree on this. If the project committers agree that we don't want IDE config files in the repository (and I'm pretty sure we all agree on this), then this rule should also by enforced by placing it in the `.gitignore` file of the repository. This will make it easier for everyone.

I know about some teams that actually prefer to commit IDE configuration files to the project source which allows them to enforce consistent code formatting rules, etc. So these developers simply cannot use the machine's global `.gitignore` file, because it would conflict with the rules for other repositories.

So I think which files to ignore may differ between projects and therefore should be enforced on the repository level.

Just my 2 cents.

chkal:
> Apparently I my feeling was correct and we need to discuss _first_ what the commiters _want_ before we discuss one _specific solution_ provided by a PR. :-)

I generally agree. But I didn't imagine this to be such a controversial topic.

> I understand your arguments, but I think `.gitignore` is the wrong tool for _enforcement_ (BTW, I did not speak about _enforcement_ so far, but just about a _helping tool_ for the contributor himself).

I think `.gitignore` does a great job for enforcement. I mean, it defines what must not go into the repository. Period. It's that easy.

However, I just wanted to see if you insist on your -1 vote or if you could imagine to change it to something that doesn't block the PR if other committers are fine with it.

chkal:
> Actually my votes are _never_ dependent on any other committer's _votes_, as I think it is essential that everybody makes up _his own_ mind and casts votes dependening on _facts_, not on _people_.

+100 for that.

> And it was exactly the latter that convinced me that to change my vote from -1 to 0: While I still am not convinced that it is a good idea to pile up an ever-growing blacklist instead of having a stable and short whitelist

I think the problem is that the GitHub review tool is a bit limited by only allowing +1, -1 and abstain. Of course there are votes in between. The [Apache Voting Rules](https://www.apache.org/foundation/voting.html) contain some nice examples for that like:

* -0.5: I don't like this idea, but I can't find any rational justification for my feelings.
* -0.9: I really don't like this, but I'm not going to stand in the way if everyone else wants to go ahead with it.
* +0.9: This is a cool idea and i like it, but I don't have time/the skills necessary to help out.

Issue #670 is a good example for that. I'm basically +1 for it, but I'm not an expert for SSE so I've, no other choice than abstaining from the vote.

mkarg:
In fact I never had a problem with the limited selection of -1, 0, and +1. In fact, the zero I cast actually is exactly what I feel: If there are enough others casting +1 I can live with it, but I will not actively support the PR *in the current form*. I do not see a need to separate 0 from -0,9. This seems somewhat academic, because due to arithmethic rounding the result will be round(sum({-0,9})) = -1 then. ;-)

Issue #679, opened 2018-10-12
RFC4918


leewalton:
RFC4918 lists a number of http response codes which are not defined in the jaxrs-api, making it difficult to cleanly respond with these codes to clients that are expecting them.

An example is the TM Forum Open APIs, which make extensive use of a 422 response code in their swagger definitions.

Are there any plans to make the jaxrs-api compliant with RFC4918?

derekm:
I’ve been saying for a while that JAX-RS needs an explicit registry of the RFCs it intends to comply with as a piece of primary documentation.

leewalton:
I've responded to this further on the pull request with my thoughts on this, so I'll just link the comment [here](https://github.com/eclipse-ee4j/jaxrs-api/pull/680#issuecomment-430202333) rather than repeating them.

leewalton:
I've closed the associated pull request, and I'm going to close this issue, as I am going to resolve this through extension of the JAX-RS API, rather than changing the api itself, as per @mkarg's recommendations in the pull request.

PR #680, opened 2018-10-13
Added response codes to allow compliance with RFC 4918

leewalton:
As per [Issue 679](https://github.com/eclipse-ee4j/jaxrs-api/issues/679), I've added added the [RFC 4918](https://tools.ietf.org/html/rfc4918) Http Response Codes. No other changes.

Signed-off-by: Lee Walton <leewalton@gmail.com>


spericas:
@leewalton Are you planning to sign an ECA?

mkarg:
The intented way according to my change in JAX-RS 1.1 is that you create a `WebDavResponse.Status` class which implements `Response.StatusType`. Actually I invented it exactly for WebDAV, as I wrote that particular class for "WebDAV Extension for JAX-RS". Instead of extending the *JAX-RS core*, you mike like to simply use that one (https://gitlab.com/headcrashing/webdav-jaxrs/blob/master/src/main/java/net/java/dev/webdav/jaxrs/ResponseStatus.java)?

mkarg:
Nobody has to hunt around. What TM Forum collected might be useful for some, but not for others, and is not part of HTTP itself. It is just one of lots of possible *use cases* of HTTP. JAX-RS is not a complete generic application for all kinds of use cases, it is just a core framework which is *extendable by intention*. So feel free to write one single extension (i. e. `Feature`) which simply adds all the needed stuff of TM Forum, and then put it on Maven Central as `org.tmforum:jaxrs-extension`. That's how the JAX-RS framework is intended to be used, otherwise it will gather more and more application level stuff over time, which cannot be the target.

PR #682, opened 2018-10-21
Travis CI does not support oraclejdk10 anymore

mkarg:
Due to Oracle's recent changes in their support policy, Travis CI does not suppot oraclejdk10 anymore.

It fails the build with the following message:
```
oraclejdk10 is deprecated. See https://www.oracle.com/technetwork/java/javase/eol-135779.html for more details. Consider using openjdk10 instead.
```

This PR follows that given advice and simply removes oraclejdk10, as openjdk10 is already supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ sudo: false
 jdk:
   - oraclejdk8
   - oraclejdk9
-  - oraclejdk10
   - oraclejdk11
   - openjdk8
   - openjdk9


mkarg:
@arjantijms Again you broke several of our rules and just followed your personal likes and schedules, and this really makes me angry. How shall people work together if everybody ignores rules? First of all, I am the assignee of this PR, not you. Second, without any technical or organizational reason your merged this PR before the end of the two-weeks *minimum* review period the commiters of this PR agreed upon. I understand your reasons, but you MUST subordinate to the existing rules.

PR #685, opened 2018-10-26
CI on JDK12

mkarg:
This PR prepares CI for building on JDK12.

It intentionally only adds OpenJDK but not OracleJDK anymore, due to several reasons:
* Oracle claims there is no difference between OracleJDK and OpenJDK anymore other than branding and support.
* Prevent frequent CI add-and-remove actions due to Oracle's recent support changes.
* Prepare a level playing field for all vendors. Otherwise we would have to add support of Azul and others who also provide repackaged OpenJDKs.

**As of today, this build will break as Travis CI does not support JDK12 already. This PR solely is solely *preparing* support, so we could merge it as soon as Travis CI is ready. Nevertheless, feel free to discuss and review already!**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ jdk:
   - openjdk9
   - openjdk10
   - openjdk11
+  - openjdk12
 
 cache:
   directories:


chkal:
AFAIK you can use `openjdk-ea` to get the EA builds of the next JDK version.

Example: https://github.com/DataSecs/Hydra/blob/3d460d1972465b60afc8f49771f0f3ef81b35232/.travis.yml#L16

But I never tried this myself.

chkal:
> Good idea, see #688. :-)

In this case maybe we should close this one?

mkarg:
Why closing it? The day will come when 12 is released, then we will need it. As the title says, this is just for preparation for that day (which will here [quite soon](https://openjdk.java.net/projects/jdk/12/)).

mkarg:
@eclipse-ee4j/eclipse-jaxrs JDK12 is released, so let's build on it! Please review this PR! :-)

Issue #681, opened 2018-10-16
Api version 2.1.2 issues


jansupol:
We [hit a few issues with latest 2.1.2 API](https://github.com/eclipse-ee4j/jersey/pull/3962) for which the API should be re-released. We still examine consequences of the issue, so the list is not full, yet

OSGI manifest:
==========
  o  OSGi specifically requires Java SE 9: Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=9.0))". This is the main reason for which the API needs to be re-released. The Felix maven plugin automatically fills this, but it is possible to let him know Java SE 8 is enough.
  o Bundle-Version: 2.1.99.b01. This should likely be 2.1.1.x
  o Export-Package: version="2.1.99"
  o Specification-Version: 2.1.99.01 - Should be 2.1
  o Automatic-Module-Name: java.ws.rs - not needed when there is module.info already
 
 o OSGI is tough one. For bing sure the OSGI is ok, I'd recommend at least trying to load it in GF. Ideally, the JAX-RS TCK should be run with the new API before it is released. Santiago or me can do that. 

Dependency Scope:
===============
  o JAX-B - while for JDK 11 the dependency for maven is in provided scope, module-info has it transitive. In JDK 11 there is no JAX-B, which means the API would not work in standalone out of the box unless the user will specifies the dependency. While it can be by design, it differs from "transitive" in module-info.

Versions - The goal is to depend on jakarta prefixed API jars. While these were not released, yet, it would make sense to wait with the release after the dependencies are released:
 o JAX-B - javax.xml.bind:jaxb-api
 o javax.activation:javax.activation-api 

Module name:
==========
 o As the discussion goes on it looks like "jakarta" should be in module name eventually

mkarg:
Feel free to provide a PR.

Regarding our own module name, this cannot be changed in 2.x as it would break backwards compatibility. Such a change can only be done in 3.x.

mkarg:
Regarding the module name, we can stay with the current one, as the PMC today announced that this is just a *recommendation*. It would break backwards compatibility, so we can only change it in 3.x.

chkal:
[...]

PR #687, opened 2018-10-27
Removed Automatic-Module-Name from manifest

chkal:
As @jansupol pointed out in #681, we don't need `Automatic-Module-Name` anymore, because we now have a real `module-info.java`. Therefore, it should be safe to remove the manifest entry.

As this PR only doesn't change the JAX-RS API, I would like to apply the rule for the reduced review period of one week.

Please note that we should also backport this PR to the `EE4J_8` branch.

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -444,7 +444,6 @@
                             <Specification-Version>${spec.specification.version}</Specification-Version>
                             <Specification-Vendor>Oracle Corporation</Specification-Vendor>
                             <specversion>${spec.specification.version}</specversion>
-                            <Automatic-Module-Name>java.ws.rs</Automatic-Module-Name>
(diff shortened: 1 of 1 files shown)


chkal:
Just one quick note: One (possible) downside of this PR is that the artifact won't include any information about the module name when the build is run with JDK8. But maybe that is not a real problem?

mkarg:
As JDK8 is unable to do something with the module name, I do not see that the removal would make a problem on JDK8. For the backport into EE4J_8 I actually do not see any good reason: The existing declaration makes no harm, it just is useless. So I would say "COULD" be backported but not "SHOULD".

chkal:
Thanks for your approvals.

As @mkarg already mentioned, it is not strictly required to backport this to the `EE4J_8` branch. However, feel free to speak up if you think we should do so.

PR #689, opened 2018-10-27
Replaced spec-version-maven-plugin with explicit configuration

chkal:
This pull request is basically a proof of concept that shows how the build could look like if we replace the `spec-version-maven-plugin` with explicit configuration of the relevant manifest entries.

Please note that the PR removes some Maven properties which aren't required anymore. It also removes the concept of the `nonFinal` flag which was the root cause of the weird version numbers created.

Most of the manifest attributes are generated exactly the same way as before. The only noteworthy difference is the handling of the version related attributes:

  * `Bundle-Version`
    * This is the OSGi bundle version
    * The previous value was `2.1` or `2.1.2` with the `final` profile and `2.1.99.b01` without it.
    * The value is now the same as the Maven project version normalized to the OSGi format. For the current master branch it would be `2.2.0.SNAPSHOT`
  * `Specification-Version`
    * This is basically the version of the specification. Something like `2.2` or `3.0`.
    * The previous value was something like `2.2` with the `final` profile and `2.1.99.01` without it.
    * This value is now defined via a Maven property.
  * `Implementation-Version`
    * This is the version number of the API classes.
    * For some reason this was set to `2.2` with and without the `final` profile before.
    * The value is now the same as the Maven project version which IMO makes most sense.

I know that there is no official decision yet if we want to get rid of the `spec-version-maven-plugin` or not. But please take the chance to review the PR to see if it could work in theory and if it improves the maintainability of the build.

And please also note that this PR is targeting the `master` branch. IMO we should keep the spec-version plugin for the `EE4J_8` branch and just explore different ways to do it in the future.

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -76,19 +76,6 @@
 
     <profiles>
         <profile>
-            <id>final</id>
-            <activation>
-                <activeByDefault>false</activeByDefault>
-            </activation>
-            <properties>
-                <!-- Final API bundle version properties -->
-                <spec.version>${next.final.spec.version}</spec.version>
-                <new.spec.version />
-                <milestone.number />
-                <non.final>false</non.final>
-            </properties>
-        </profile>
-        <profile>
             <!-- Use it with release-perform goal to skip another test run. -->
             <id>skip-tests</id>
             <activation>
@@ -401,32 +388,6 @@
                     </executions>
                 </plugin>
                 <plugin>
-                    <!-- This plugin generates the spec.* properties used in maven-bundle-plugin -->
-                    <groupId>org.glassfish.build</groupId>
-                    <artifactId>spec-version-maven-plugin</artifactId>
-                    <version>1.5</version>
-                    <configuration>
-                        <specMode>jakarta</specMode>
-                        <spec>
-                            <nonFinal>${non.final}</nonFinal>
-                            <jarType>api</jarType>
-                            <specVersion>${spec.version}</specVersion>
-                            <newSpecVersion>${new.spec.version}</newSpecVersion>
-                            <specBuild>${milestone.number}</specBuild>
-                            <specImplVersion>${project.version}</specImplVersion>
-                            <apiPackage>${api.package}</apiPackage>
-                        </spec>
-                    </configuration>
-                    <executions>
-                        <execution>
-                            <goals>
-                                <goal>set-spec-properties</goal>
-                                <goal>check-module</goal>
-                            </goals>
-                        </execution>
-                    </executions>
-                </plugin>
-                <plugin>
                     <groupId>org.apache.felix</groupId>
                     <artifactId>maven-bundle-plugin</artifactId>
                     <version>${maven.bundle.plugin.version}</version>
@@ -436,14 +397,14 @@
                             <_failok>true</_failok>
                             <Build-Id>${buildNumber}</Build-Id>
                             <Bundle-Description>Java API for RESTful Web Services (JAX-RS)</Bundle-Description>
-                            <Bundle-Version>${spec.bundle.version}</Bundle-Version>
-                            <Bundle-SymbolicName>${spec.bundle.symbolic-name}</Bundle-SymbolicName>
+                            <Bundle-Version>${project.version}</Bundle-Version>
+                            <Bundle-SymbolicName>${api.package}-api</Bundle-SymbolicName>
                             <DynamicImport-Package>*</DynamicImport-Package>
-                            <Extension-Name>${spec.extension.name}</Extension-Name>
-                            <Implementation-Version>${spec.implementation.version}</Implementation-Version>
-                            <Specification-Version>${spec.specification.version}</Specification-Version>
+                            <Extension-Name>${api.package}</Extension-Name>
+                            <Implementation-Version>${project.version}</Implementation-Version>
+                            <Specification-Version>${spec.version}</Specification-Version>
                             <Specification-Vendor>Oracle Corporation</Specification-Vendor>
-                            <specversion>${spec.specification.version}</specversion>
+                            <specversion>${spec.version}</specversion>
                             <_versionpolicy>[$(version;==;$(@)),$(version;+;$(@)))</_versionpolicy>
                             <_nodefaultversion>false</_nodefaultversion>
                         </instructions>
@@ -606,10 +567,6 @@
                 <artifactId>maven-bundle-plugin</artifactId>
             </plugin>
             <plugin>
-                <groupId>org.glassfish.build</groupId>
-                <artifactId>spec-version-maven-plugin</artifactId>
-            </plugin>
-            <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
             </plugin>
             <plugin>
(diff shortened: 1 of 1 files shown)


chkal:
This the manifest file created for the current `master` branch with the new configuration:

```
Manifest-Version: 1.0
Automatic-Module-Name: java.ws.rs
Bnd-LastModified: 1540634850329
Build-Id: 10/27/2018 12:07 PM
Build-Jdk: 1.8.0_181
Built-By: ck
Bundle-Description: Java API for RESTful Web Services (JAX-RS)
Bundle-DocURL: https://www.eclipse.org/org/foundation/
Bundle-License: http://www.eclipse.org/legal/epl-2.0, https://www.gnu.
 org/software/classpath/license.html
Bundle-ManifestVersion: 2
Bundle-Name: javax.ws.rs-api
Bundle-SymbolicName: javax.ws.rs-api
Bundle-Vendor: Eclipse Foundation
Bundle-Version: 2.2.0.SNAPSHOT
Created-By: Apache Maven Bundle Plugin
DynamicImport-Package: *
Export-Package: javax.ws.rs;version="2.2.0.SNAPSHOT";uses:="javax.net.
 ssl,javax.ws.rs.core",javax.ws.rs.container;version="2.2.0.SNAPSHOT";
 uses:="javax.ws.rs.core",javax.ws.rs.sse;version="2.2.0.SNAPSHOT";use
 s:="javax.ws.rs.client,javax.ws.rs.core",javax.ws.rs.ext;version="2.2
 .0.SNAPSHOT";uses:="javax.ws.rs,javax.ws.rs.core",javax.ws.rs.client;
 version="2.2.0.SNAPSHOT";uses:="javax.net.ssl,javax.ws.rs,javax.ws.rs
 .core",javax.ws.rs.core;version="2.2.0.SNAPSHOT";uses:="javax.ws.rs,j
 avax.xml.bind.annotation,javax.xml.bind.annotation.adapters,javax.xml
 .namespace"
Extension-Name: javax.ws.rs
Implementation-Version: 2.2-SNAPSHOT
Import-Package: javax.net.ssl,javax.ws.rs;version="[2.2,3)",javax.ws.r
 s.client;version="[2.2,3)",javax.ws.rs.core;version="[2.2,3)",javax.w
 s.rs.ext;version="[2.2,3)",javax.xml.bind.annotation,javax.xml.bind.a
 nnotation.adapters,javax.xml.namespace
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
Specification-Vendor: Oracle Corporation
Specification-Version: 2.2
Tool: Bnd-3.5.0.201709291849
```

mkarg:
One question, as I have no clue of OSGi: What we provide in our published JARs is not an *implementation* (that would be Jersey) of the spec, it *is* the specification. So is it really correct to have an `Implementation-Version` line at all? Just for interest.

chkal:
My understanding is that there is a *spec version* (like JAX-RS 2.1) and possibly multiple *spec implementation versions* which contain the spec's classes (there was a 2.1.0 release and maintenance releases like 2.1.1 and 2.1.2 which are actually *implementations* of JAX-RS 2.1). So one could say the former one is the actual specification and the latter one is a Maven artifact. On top of that you also have the actual implementations like Jersey. And yes, the wording is confusing. :laughing: 

From the [Maven and OSGi packaging, naming, and versioning rules](https://javaee.github.io/glassfish/wiki-archive/Maven%20Versioning%20Rules.html):

```
We define the following additional variables:

SPEC_VERSION		- version number of the JCP specification,
			  always of the form <major>.<minor>
SPEC_IMPL_VERSION	- version number of the API classes, derived from
			  SPEC_VERSION by adding an optional micro version
			  number
IMPL_VERSION		- version number of the implementation
```

So for the JAX-RS build we would have:
* "Specification-Version" manifest entry = SPEC_VERSION
* "Implementation-Version" manifest entry = SPEC_IMPL_VERSION

But all that is just my understanding. So maybe I'm totally wrong about that. :laughing:

chkal:
BTW: AFAIK the manifest entries I described above aren't actually OSGi related (as far as I know). Only `Bundle-Version` is relevant for OSGi. The other ones are just metadata added to the JAR artifact. Not sure if Glassfish is actually doing something with `Specification-Version` and `Implementation-Version`.

andymc12:
Should `Specification-Vendor: Oracle Corporation` be changed to something like `Specification-Vendor: Eclipse Foundation`?

mkarg:
@spericas @jansupol Can you please review this PR?

mkarg:
@chkal Can you please modify both commits e. g. by adding a blank line, and push the branch again? I hope this kicks the hanging IP bot. Thanks.

chkal:
@mkarg Thanks for the hint. I rebased against master and the ip-validation now passes.

@andymc12 Could you redo your approval? Thanks!

jansupol:
While I kind of like the idea of removing the plugin, the properties may not be quite fine.
The jar name 2.2-snapshot can be ok, but in OSGi, the version name usually is 2.1.0.x. The reason is that OSGi uses a range of versions. and possibly you would not want to use 2.2. final as well as 2.2. snapshot, both of which fit into the range [2.2,3). This is what the plugin did: it put 2.1.99.01 version to the manifest. Also, you possibly want to distinguish between snapshot versions, so instead of SNAPSHOT, a version number is used. 

So 2.2 would be used just for final release of 2.2, 2.1.x for snapshots

mkarg:
@jansupol Understood and agreed, but we can simply write that OSGi-specific version numbers *by hand*. In the end this is just text in a file stored in our repo.

BTW, according to [OSGi Semantic Versioning](https://www.osgi.org/wp-content/uploads/SemanticVersioning.pdf) the version scheme of OSGi (major.minor.bugfix.build) seems to be quite the same as Maven's (major.minor.bugfix-build).

Maven has *exactly* the same problem with `SNAPSHOT` in version ranges (this is not a OSGi-specific problem). But `2.2-SNAPSHOT` is *not* a *post*-release patch ontop of `2.1`, so `2.1.99.01` is actualy *wrong* as that means "First build of 99th bugfix AFTER 2.1", it is the exact opposite: It is a *pre*-release (in the sense of 2.2.0.alpha-1, i. e. "First build BEFORE 2.2). The correct solution is the best practice enforced by OSSRH: SNAPSHOTs MUST NOT (and CANNOT) be published to Maven Central as it is a RELEASE-type repository (it really is technically impossible), but it can ONLY be uploaded into SNAPSHOT-type repositories (that's why there always are *two* repositories), and Maven dependencies MUST NOT be written as `[2.1, 2.2)` but in fact MUST be written as either `[2.1, 2.1.99999]` (or alternative solutions which explicitly filter pre-releases; in fact the Maven community since long time discusses the possibility to explicitly mark exclusion of SNAPSHOTs). In the same way the consumer can simply exclude snapshots in the OSGi range.

Such a *text-based* solution is much more intuitive and clear to understand compared to the magic of any plugin and any misuse of the originally semantic version digits.

chkal:
I agree that snapshots are problematic, both in the Maven and in the OSGi world.

Just quick note regarding how bundle versions will be generated with this pull request:

| Maven Version | OSGi Bundle Version |
| --- | ---|
| 2.2 | 2.2.0 |
| 2.2.0 | 2.2.0 |
| 2.2-SNAPSHOT | 2.2.0.SNAPSHOT |

However, I agree with @mkarg that a good solution is actually to distinguish between snapshot and release repositories.

arjantijms:
@chkal I do wonder what's so special about JAX-RS that from a dozen or so projects, it's the only one that had to remove the plug-in. What does it do so differently then?

chkal:
@arjantijms I don't think that JAX-RS is the only project running into issues. The JPA 2.2.2 release had [similar problems](https://www.eclipse.org/lists/ee4j-pmc/msg01417.html) with weird bundle versions. And I don't really see any benefit of using the spec-version-plugin. Actually it makes many things more complicated to reason about. Especially because there is no documentation anywhere.

chkal:
Although I created this PR as a POC, we already have enough approvals for merging it. And the PR is open for more than two weeks now. So I'll merge it later today.

chkal:
We have enough +1 votes and the PR is open for over three weeks now. So I'll merge this now. Thanks everyone for your reviews.

PR #696, opened 2018-11-08
MediaType improvements

tareqhs:
The struct is immutable, reuse the result of hash code

This makes it more efficient to use MediaType as keys in hash maps, caches and other hash-based strucutres.

* single computation of hash code
* final fields

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
@@ -35,9 +35,9 @@ import javax.ws.rs.ext.RuntimeDelegate;
 @SuppressWarnings("JavaDoc")
 public class MediaType {
 
-    private String type;
-    private String subtype;
-    private Map<String, String> parameters;
+    private final String type;
+    private final String subtype;
+    private final Map<String, String> parameters;
     private final int hash;
 
     /**
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
@@ -38,7 +38,7 @@ public class MediaType {
     private String type;
     private String subtype;
     private Map<String, String> parameters;
-    private int hash;
+    private final int hash;
 
     /**
      * The media type {@code charset} parameter name.
@@ -250,6 +250,7 @@ public class MediaType {
             parameterMap.put(CHARSET_PARAMETER, charset);
         }
         this.parameters = Collections.unmodifiableMap(parameterMap);
+        this.hash = Objects.hash(this.type.toLowerCase(), this.subtype.toLowerCase(), this.parameters);
     }
 
     /**
@@ -372,12 +373,7 @@ public class MediaType {
     @SuppressWarnings("UnnecessaryJavaDocLink")
     @Override
     public int hashCode() {
-        int h = this.hash;
-        if (h == 0) {
-            h = Objects.hash(this.type.toLowerCase(), this.subtype.toLowerCase(), this.parameters);
-            this.hash = h;
-        }
-        return h;
+        return this.hash;
     }
 
     /**
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
@@ -374,7 +374,7 @@ public class MediaType {
     public int hashCode() {
         int h = this.hash;
         if (h == 0) {
-            h = Objects.hash(this.parameters, this.subtype.toLowerCase(), this.type.toLowerCase());
+            h = Objects.hash(this.type.toLowerCase(), this.subtype.toLowerCase(), this.parameters);
             this.hash = h;
         }
         return h;
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
@@ -19,6 +19,7 @@ package javax.ws.rs.core;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Map;
+import java.util.Objects;
 import java.util.TreeMap;
 
 import javax.ws.rs.ext.RuntimeDelegate;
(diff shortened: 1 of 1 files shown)


mkarg:
@tareqhs Please amend your commit using a valid Sign-Off, so the IP validation will pass. Thanks.

tareqhs:
@mkarg done

chkal:
Wouldn't it be simpler to precalculate the hash in the constructor?

BTW: The ip-validation check is broken (again).

spericas:
LGTM. While we are at it, we could probably add `final` to the member variables.

mkarg:
+1 for precomputed hash
+1 for final -- it took me several minutes to understand that the existing code is actually "effectively final"

mkarg:
@chkal I hate that bot. But it should work once the proposed changes are pushed.

tareqhs:
hi @spericas , @mkarg I would prefer avoiding redundant calculations - this is tbe whole point of my change. A lazy-initialization approach makes a decent optimistic guess given MediaType is not necessarily used in hash structures and it isup to the consumer to choose whether instances are reused or not.

mkarg:
@tareqhs Looking at the complexity of the code, the used data types, the typical length of the strings, and the power of decent hardware, I doubt that your code actually will provide any *measurable* benefit in the real world at all, at least once Hot Spot decided to compile the method `hashCode`. Having said that, I still must agree with @chkal that precomputing the hash code is really better than checking for zero all the time, as the constructor is called once for a few media types at runtime, but hashCode will be invoked millions of time over the livetime of an application. If you don't want to fix that, I can send a separate PR for that, just tell me.

jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
```diff
@@ -37,6 +37,7 @@
     private String type;
     private String subtype;
     private Map<String, String> parameters;
+    private int hash = 0;
```
arjantijms: 0 is already the default value. Can be omitted.

arjantijms:
@tareqhs Shall we change the title of this PR to "Use pre-computed MediaType hash codes" or something along those lines?

"Memorize" is not a typically term here, and might be confusing if it shows up in release notes ;)

tareqhs:
@arjantijms Thanks, my mistake. I meant "memoize", not "memorize". (My PR is about memoization, not pre-computation)

mkarg:
@tareqhs Not your fault, but mine. I changed as I thought you actually meant pre-computation (as @chkal proposed).

arjantijms:
@tareqhs Thx, my mistake too as I should have seen you meant "memoize" and not "pre-compute" indeed.

tareqhs:
@mkarg @chkal Thanks for the approvals. Please review my final change (composite hash code) and feel free to merge if you see it good.

chkal:
@tareqhs Thanks! The code looks great! But the ip-validation check still fails. Could you perhaps squash the commits into a single one with a correct Signed-off-by line? See the [details of the failed check](https://dev.eclipse.org/eclipse-webhook/services/status_details.php?id=5be7f6fa8c685) provided here. And AFAIK the email must match the email of your Eclipse account which you used to sign the [ECA](https://www.eclipse.org/legal/ECA.php).

chkal:
Yeah, I just [saw it is still failing](https://github.com/eclipse-ee4j/jaxrs-api/pull/696#issuecomment-437658466). AFAIK @mkarg mentioned some workaround for such cases. IIRC something like added your GitHub username in the Signed-Off line. But maybe @mkarg can tell us more about it?

jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
```diff
@@ -367,7 +371,14 @@ public boolean equals(final Object obj) {
     @SuppressWarnings("UnnecessaryJavaDocLink")
     @Override
     public int hashCode() {
-        return (this.type.toLowerCase() + this.subtype.toLowerCase()).hashCode() + this.parameters.hashCode();
+        int h = this.hash;
+        if (h == 0) {
+            h = this.parameters.hashCode();
```
mkarg: The PR's title is wrong now, as now it is not only about memoization, it also is about using a different algorithm. You should separate the PR into two commits, one for the memoization, and one for the changed algorithm.

jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
```diff
@@ -367,7 +371,14 @@ public boolean equals(final Object obj) {
...
+            h = 31 * h + this.subtype.toLowerCase().hashCode();
+            h = 31 * h + this.type.toLowerCase().hashCode();
```
mkarg: DRY... You copied an algorithm implemented by the JRE. Instead of *copying* it, you simply should *invoke* `Object.hash(this.parameters, this.subtype, this.type)`.

tareqhs:
@mkarg @chkal - `Objects.hash(Object...)` requires Java 1.7 so I had to use the slightly uglier ``Arrays.hashCode(Object[])`` ...

mkarg:
@tareqhs Regarding the IP bot: You signed with `Signed-off-by: Tareq Sharafy <tareq.sha@gmail.com>`. Is this really the exact same email address you signed the ECA with? Then this is a known bug and you should try to replace your full name by simply `tareqhs`; this helped in all previous cases.

@eclipsewebmaster Please fix https://bugs.eclipse.org/bugs/show_bug.cgi?id=534451!

mkarg:
@tareqhs The minimum Java version of JAX-RS 2.1+ is 8, so please use `Objects.hash`. Thanks.

mkarg:
@tareqhs Please rebase the commits found in my PR (https://github.com/tareqhs/jaxrs-api/pull/1) ontop of your `patch-1`. Thanks. :-)

jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java
mkarg: Squashing will not help the IP bot. Looking at all the time it fails, I really that bot is simply buggy. The only solution you have is: When it hangs, push again. When it fails, replace full name by Github ID.

mkarg:
@tareqhs As you can see, we liked your contributions. Maybe you are interested in providing more code optimizations? Like replacing custom code by JRE 8 utilities?

Issue #690, opened 2018-10-30
Release to OSSRH staging repository


m0mus:
The previous 2.1.2 release didn't uptake the new Eclipse dependencies. It's been released to public, so there is no way of changing it. The new release (2.1.3) is needed. Please release it to OSSRH staging only. Don't deploy it to Maven Central until everything is tested.

- [x] Uptake new Eclipse versions of dependencies (if any). Use [GlassFish 5.1 Release Tracker](https://wiki.eclipse.org/Eclipse_GlassFish_5.1_Components_Release_Tracker) wiki to find components versions to use (see #692).

- [x] Release to OSSRH staging repository

- [x] Update [GlassFish 5.1 Release Tracker](https://wiki.eclipse.org/Eclipse_GlassFish_5.1_Components_Release_Tracker) with released version number.

spericas:
We should fix the OSGi issues as well #686.

mkarg:
Guys, just as an good advice from a long-time Github user: Try to open one issue per problem. This makes it easier to manage the parallel work on all of them. :-) We now have two issues open both with a collection of problems inside it. Nobody knows who is working on which item. :-(

mkarg:
@m0mus This big-bang-approach won't work outside Oracle. To be able to build against Jakarta dependencies, these dependencies MUST be available on Github: Travis CI will fail otherwise, in turn blocking the PR on Github, so we cannot merge it into JAX-RS 2.1.3. So the right way is to *not* stop the world, but to bootstrap the dependencies *bottom-up* on Maven Central *or* on Eclipse Nexus. Note that not all projects pull dependencies from OSSRH staging because the EF wants us to use Eclipse Nexus for this purpose.

@bravehorsie When can we expect to have a public release of JAXB available to committers outside Oracle?

@bshannon When can we expect to have a public release of JAF available to committers outside Oracle?

bshannon:
Markus KARG wrote on 10/30/2018 11:17 AM:
>
> Note that not all projects pull dependencies from OSSRH staging because the EF
> wants us to use Eclipse Nexus for this purpose.
>
Can you point to this statement from Eclipse and instructions for how to
implement what they want?

> @bshannon <https://github.com/bshannon> When can we expect to have a public
> release of JAF available to committers outside Oracle?
As soon as I figure out how to do a release properly.  I currently have it
staged at OSSRH (I think), but if something else is required I need to know what
to do.

mkarg:
[...]

PR #699, opened 2018-11-13
Excluding JRE's implied JAXB implementation with JDK11+

senivam:
As JDK11+ explicitly requests a particular version of JAXB, it is more safe to exclude the version of JAXB implicitly contained by the JRE itself.


senivam:
pom.xml is reduced only to lines:

lines changed:
`<release>9</release>`
to
`<source>9</source>
<target>9</target>`

Because using <release> makes the JDK to use only those libraries which are a part of  JDK (9). Hovewer JDK 11 does not contain javax.xml.bind. We provide it externally (using maven dependency). Thus it is preferable no to limit JDK only to internal API because JDK 11 will not find required classes.

mkarg:
I don't get it. Christian do you know what he means? I am willing to believe he is right, I just don't understand the actual detail.

chkal:
Ah, I think I understand the problem. The `release` argument works like this:

> For the platforms predefined in javac, --release N is equivalent to -source N -target N -bootclasspath &lt;bootclasspath-from-N&gt;

So basically using `release=9` means that we are getting the bootclasspath of JDK9 even if we are compiling with JDK11. But this also means that we are getting JAXB via the bootclasspath, although this profile adds JAXB as an additional dependency (because JAXB isn't available with JDK11).

I think it is fine to merge this PR. I guess using the `release` flag doesn't really make sense in this profile. Especially because the real target JDK is actually 8.

chkal:
@senivam It would be great if you could fix the indentation. Looks like some mixed spaces/tabs problem. See the diff in the files changed tab.

mkarg:
@chkal Ah I see, right! I missed that! Thanks! Would be cool if future PR description would be a bit more elaborate, so voting gets easier. :-)

mkarg:
@senivam Thanks for this PR! I have adjusted title and description to make a bit more clear what it talks about. Please confirm that the new text now matches the core point of your proposal. :-)

senivam:
@mkarg, @chkal indentation is fixed, title adjustment reflects what I've meant, thank you :)

mkarg:
@senivam Assuming that `EE4J_8` will be published before the minimum release review duration of two weeks (!) is over, and given the fact that your PR is not *essential* for EE4J_8 (since we will skip JDK 11 support for *this* release anyways), you maybe like to switch target branch to `master`?

mkarg:
@spericas @chkal We dont' need this for OSSRH staging, so we have two options: Staging *without* (i. e. I will change target branch of this PR to `master`) *or* including before staging (i. e. we should *fast-track* this PR). Opinions?

jansupol:
Should this be a blocker for #690, I vote for fast tracking this.

mkarg:
@jansupol It is not a blocker as the POM skips this dependencies now that we decided not compile on JDK11, but it is just "dirty" because the dependencies actually are contained in the JDK11 profile of the POM file.

chkal:
I don't think we need to merge this before staging the release. The affected profile isn't active during the release, so I don't see any reason to merge this now.

mkarg:
*Replaced by https://github.com/eclipse-ee4j/jaxrs-api/pull/704*

PR #700, opened 2018-11-13
Improve MediaType compatibility check

### LINKED ISSUE #543: Improve MediaType.isCompatible
MediaType.isCompatible will match a concrete non wildcard media type against a wildcard type whose subtype has a qualifier, for example, "application/xml" will be found to be compatible with "**/**+json" which confuses the users.

Consider updating the followng

```
(type.equals(MEDIA_TYPE_WILDCARD) || other.type.equals(MEDIA_TYPE_WILDCARD)
```

to

```
(type.equals(MEDIA_TYPE_WILDCARD) || other.type.equals(MEDIA_TYPE_WILDCARD) && 

(subtype.equals(MEDIA_TYPE_WILDCARD) || other.subtype.equals(MEDIA_TYPE_WILDCARD)
```
#### Affected Versions
[2.0.1]

mszabo-wikia:
As described in https://github.com/eclipse-ee4j/jaxrs-api/issues/543, the `javax.ws.rs.core.MediaType#isCompatible` method is not behaving as expected when one of the media types to be compared has a wildcard type and explicit subtype, e.g. `*/json`. The referred media type is considered to be compatible with any type, even if the subtypes do not match (e.g. `*/json` will be considered compatible with `application/octet-stream`). As such, if an implementation of JAX-RS relies on `javax.ws.rs.core.MediaType#isCompatible` to determine if the media type of an incoming request is compatible with the accepted media types of a resource method, as per Section 3.5 of the specification (via Step 3a in Section 3.7.2), it might result in a surprising behaviour for the end-user.

This change aims to improve and clarify the logic of `javax.ws.rs.core.MediaType#isCompatible` so that it handles wildcard types better, and adds unit tests for the referred method.

Resolves #543 

Signed-off-by: mszabo-wikia <mszabo@wikia-inc.com>

### DIFF
--- a/jaxrs-api/src/test/java/javax/ws/rs/core/MediaTypeTest.java
+++ b/jaxrs-api/src/test/java/javax/ws/rs/core/MediaTypeTest.java
@@ -16,13 +16,16 @@
 
 package javax.ws.rs.core;
 
+import static org.hamcrest.CoreMatchers.not;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertThat;
+
+import org.hamcrest.Description;
+import org.hamcrest.DiagnosingMatcher;
+import org.hamcrest.Matcher;
+import org.junit.Test;
 
 import java.util.Map;
-
-import org.junit.Test;
 
 /**
  * {@link MediaType} unit test.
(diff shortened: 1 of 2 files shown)


mkarg:
@struberg What do you think?

spericas:
@mszabo-wikia If I'm reading this correctly, this is saying: given t1/s1 and t2/s2 then the condition is:

(t1 = t2 OR t1 = * OR t2 = *) AND (s1 = s2 OR s1 = * OR s2 = *)

If so, perhaps easier to write it this way. Also, I suggest updating the Javadoc rather than adding comments inside the method.

mszabo-wikia:
Thanks! I've improved the format as you suggested.

mkarg:
+1 for tests, but I think we should use unit tests instead of adding FIT ontop of our stack

spericas:
There is currently no direct support for "+suffix" in the MediaType class. Testing things like "xml+*" or "*+json" is beyond scope IMO. Adding support for "+suffix" would need to be done carefully to maintain backward compatibility as it is currently part of subtype.

mszabo-wikia:
Great, so just to clarify, shall I add a separate test case verifying the (in)compatibility of common media types, as suggested above?

mkarg:
@mszabo-wikia I think unit tests are always a good thing. Feel free to add test cases for all that the specification demands but is missing currently. But take care to really only test what is free of discussion, otherwise there might pop up more requests to change that code once written.

struberg:
Usually in Java specs we do only test things which are clearly specified. If it's clearly specified that `text/xml+*` *MUST NOT* work, only then it should be tested. 

If it's just not specified then we must not test it but leave it to vendors.

mkarg:
@eclipse-ee4j/eclipse-jaxrs All committers please review this change request! :-)

mkarg:
@asoldano @andymc12 Please review this PR.

mkarg:
Shall we adopt this bug fix into 2.1.6, too?

jaxrs-api/src/test/java/javax/ws/rs/core/MediaTypeTest.java
```diff
@@ -58,4 +60,110 @@ public void testNullConstructorValues() {
...
+    @Test
+    public void testMediaTypeWithWildcardTypeCompatibleWhenSubtypeMatches() {
+        MediaType wildcard = new MediaType(MediaType.MEDIA_TYPE_WILDCARD, "json");
+
+        assertTrue(wildcard.isCompatible(MediaType.APPLICATION_JSON_TYPE));
+        assertTrue(MediaType.APPLICATION_JSON_TYPE.isCompatible(wildcard));
+    }
+
+    /**
+     * Test that a wildcard type {@link MediaType} value with explicit subtype is not compatible with
```
mkarg: This comment tells a different story than this tests actual source code.

jaxrs-api/src/test/java/javax/ws/rs/core/MediaTypeTest.java
```diff
@@ -58,4 +60,110 @@ public void testNullConstructorValues() {
...
+
+    /**
+     * Test that a {@link MediaType} value with explicit type and wildcard subtype
+     * is compatible with a {@link MediaType} value with the same type and an explicit subtype.
+     */
+    @Test
+    public void testMediaTypeWithWildcardSubtypeCompatibleWhenTypeMatches() {
+        MediaType wildcard = new MediaType("application", MediaType.MEDIA_TYPE_WILDCARD);
+
+        assertTrue(wildcard.isCompatible(MediaType.APPLICATION_JSON_TYPE));
```
mkarg: Actually a more specific naming would make it easier to understand the tests, so the comments would be mostly unnecessary:
```java
assertThat(mediaTypeApplicationAny, isCompatibleWith(MediaType.APPLICATION_JSON_TYPE));
```

mkarg:
@mszabo-wikia Can you please check my comments? Thanks. :-)

mszabo-wikia:
Thanks @mkarg I've implemented your suggestions :)

jaxrs-api/src/test/java/javax/ws/rs/core/MediaTypeTest.java
mkarg: @mszabo-wikia Did I misunderstand the JavaDoc, or is another commit in preparation? :-)

mszabo-wikia: @mkarg I took a look and this was a copy-paste of the first test case's description by accident, sorry about the confusion :)

mkarg:
@andymc12 @spericas @chkal Can you please re-review (your votes got dismissed by latest push). Thanks. :-)

mszabo-wikia:
@mkarg squashing is OK with me :)

mkarg:
@chkal @spericas This is a bug fix. Would you deny it if I cherry-pick it into 2.1.6-SNAPSHOT?

chkal:
@mkarg I'm fine with cherry-picking to 2.1.6-SNAPSHOT!

PR #704, opened 2018-11-20
Excluding JRE's implied JAXB implementation with JDK11+

senivam:
As JDK11+ explicitly requests a particular version of JAXB, it is more safe to exclude the version of JAXB implicitly contained by the JRE itself.

*Replaces https://github.com/eclipse-ee4j/jaxrs-api/pull/699*

**The review period ends after seven days already, as this is a replacement of #699, which already started seven days ago.**


senivam:
rebased to master from #699

PR #719, opened 2019-01-08
Fixed Javadoc warning in ClientBuilder

chkal:
I just noticed that the Maven build logs the following warning:

```
[WARNING] Javadoc Warnings
[WARNING] /home/ck/projects/jaxrs/jaxrs-api/jaxrs-api/src/main/java/javax/ws/rs/client/ClientBuilder.java:254: warning - Tag @see: reference not found: SseEventSource.Builder#reconnectingEvery(long, TimeUnit)
```

It looks like using the FQCN for referencing `SseEventSource` fixes the problem.

This is a very minor fix, so I propose a fast-track review period of one day.


mkarg:
BTW, do you have an explanation why this qualification is actually needed?

chkal:
@mkarg Could you approve again. I just updated the copyright year.

Regarding the cause of the issue: I guess javadoc gets confused by the inner class. So something like `SseEventSource.Builder` is interpreted as a FQCN with `SseEventSource` being the package and `Builder` being a top-level class.

PR #724, opened 2019-01-24
Reducing number of build platforms

mkarg:
In the past months we had ongoing trouble by the always changing set of JDKs, and by several bugs particularly in EA builds. The steady change of implies a lot of work and trouble, but brings no real benefit. In an attempt to find an at-most stable set of JDKs to build upon, I am proposing to reduce the set of JDKs massively using this PR.


mkarg:
@jansupol We do not avoid *running* we only avoid *building on Travis / Jenkins*. What is your counterproposal? Explicitly adding and removing Oracle JDKs each time they change their support matrix?

jansupol:
Yeah, running the Travis, that's building the API, you are saying the same I do. What support removal do you refer to? Do you refer to the JDK 8 EOL? Do you think that the users stop using JDK 8 instantly?

mkarg:
@jansupol I assume users will stay with JDK 8 a very long time, but this PR is not about *using* it is solely about which JDK *we* are using to build on Travis and Jenkins. It has no impact at all on users.

jansupol:
@mkarg In contrary, it has a big impact on users. If we stop testing on Oracle JDK and it will stop working there, it will stop working for users. We cannot tell them to use OpenJDK, mostly JDK version is not under their control

spericas:
I share the concern from @jansupol. Oracle JDK is still the most popular JDK today.

mkarg:
As Oracle pretends that there is no more difference between Oracle JDK and Open JDK I so not share these concerns, but for the sake of getting this argument solved, I will modify the PR and re-add Oracle JDK again... Stay tuned.

mkarg:
@jansupol @spericas @chkal Still building on Oracle JDK 8, 9 and 11 again as requested. Please re-review so I can merge and hence fix the hanging build. Thanks. :-)

mkarg:
**Fast-Track Announcement**: As this is just a build change and no API is touched at all, I will fast-track this PR. Following our current committer rules, this means I will merge tomorrow if nobody objects by casting a -1 vote.

Issue #721, opened 2019-01-14
Moving Eclipse Jenkins from old infra to new one


jansupol:
EF is moving old Jenkins instances from old infra to a new infra. The Jenkins is available at https://jenkins.eclipse.org/jaxrs/, the build job has been transferred.

EF is asking whether [it is possible to drop](https://bugs.eclipse.org/bugs/show_bug.cgi?id=541740) the old infra.

@mkarg Can you please check the transferred release job? 

@everyone Can the old infra be disposed in favor of the new one?

mkarg:
Existing infra still working well. New infra failing due to missing JDKs labels. Will fix this after 2.1.5, which I do with the old infra.

mkarg:
@jansupol I am a bit confused. The referenced ticket does not say the EF wants to migrate us, but instead *you explicitly asked* for the migration and in particular, to shutdown the existing and quite fine working Jenkins instance. As I am using that old infra for releasing and was pretty happy with it, I wonder if you like to elaborate a bit why you requested the shutdown of that service without prior discussing us / me *first*?

arjantijms:
@jansupol 
>@everyone Can the old infra be disposed in favor of the new one?

Yes, when everything is transferred please go ahead. It's very confusing to only have JAX-RS on this old infra.

mkarg:
I'm still working on this. Apparently we need to change the supported set of JDKs. Please find PR #724  so I can continue.

jansupol:
I fail to see how removing Oracle JDK from Travis helps with building on Jenkins.

mkarg:
@jansupol Jenkins is improved by replacing Oracle JDK 8 by OpenJDK Latest in the Jenkinsfile. The Travis changes are for Travis probles, not for Jenkins problems. So for #721 we only need one of those fixes, the other fix is needed for Travis problems.

jansupol:
The PR is a bit problematic, There are 3 parts of it, in a single PR:
1) Replacing Oracle JDK by OpenJDK in Jenkins. That's fine. Oracle JDK is available in bash only, it's fine to build the JAX-RS API with OpenJDK.
2) Removing the EA from the build. It's fine, it's causing issues, we can remove it.
3) Removing Oracle JDK 8+. I do not see any rationale behind this. I do not see how this fixes any Travis problem.
I would suggest splitting the commit per the issue it addresses.

mkarg:
**This issue is dependent of #724.**

[...]

PR #726, opened 2019-01-26
New Eclipse Foundation CI Infrastructure (Jenkins)

mkarg:
This PR modifies the branch `EE4J_8` so it can be built using the new Eclipse Foundation CI Infrastructure (Jenkins).

See #721 and https://wiki.eclipse.org/EE4J_Build.

**As these are no API changes, I propose a fast-track review period of just one day as per our [recently update committer rules](https://github.com/eclipse-ee4j/jaxrs-api/wiki/Committer-Conventions#minimum-length-of-review-period-before-merge--close-of-prs-and-closing-of-issues).**

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -139,6 +139,12 @@
                                 <goals>
                                     <goal>sign</goal>
                                 </goals>
+                                <configuration>
+                                    <gpgArguments>
+                                        <arg>--pinentry-mode</arg>
+                                        <arg>loopback</arg>
+                                    </gpgArguments>
+                                </configuration>
                             </execution>
                         </executions>
                     </plugin>


mkarg:
1. There are two reasons why I was reluctant to adoption of the parent POM, besides the free spirit of this project. The first is, that it is simply not needed -- we live great without it and while I set up the build process I never missed it; I tried adding it but it was easier to *not* use it. The second is, it does not easen things but actually produces more complexity. The truth is, that we have full control now, and I want to keep this control by *our* committers. The parent pom adds stuff which we do neither want nor need, and as it is controlled by a *different* group, we have no control of our own build anymore, which some of us, including me, strictly do not want. So as long as I am a committer in this project, and as long as the parent pom group refuses to reduce the parent POM to the absolute bare minimum (i. e. turn it into an *include dependency* instead of a *parent*), there is a solid -1 from me for adopting the parent pom.

2. It might happen that between now and the actual GF publication next week we find a showstopper and must fix it. So we must stay able to publish 2.1.6 on short notice. As we want to get rid of the old infra ASAP, we must make the new infra working instantly on EE4J_8. Once GF is published I will (as always) provide a 2.1.6-rollup into master as a subseqent PR, so we stay in sync. So it is simply a safety measure to have this in EE4J_8 *before* we have it in *master*.

mkarg:
@jansupol Please cast your +1 not just as comment but really as a review vote. Otherwise the merge button stays blocked. Thanks.

arjantijms:
@jansupol 
>I understand the free spirit of this project does not want to be bound by ee4j

I'm not really aware of JAX-RS having such a free spirit. Personally at least I see JAX-RS as Jakarta EE's web service API first and foremost, and therefor I'm totally fine with being bound to EE4J and certainly with aligning with EE4J.

PR #728, opened 2019-01-27
Fixed JavaDocs: <tt> must either be <code> or {@code}

mkarg:
Latest JavaDoc does not allow `<tt>` anymore as it was deprecated with HTML 5, hence replacing it by `<code>` or `{@code}`.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -23,7 +23,7 @@
 
     <groupId>jakarta.ws.rs</groupId>
     <artifactId>jakarta.ws.rs-examples</artifactId>
-    <version>2.1.5</version>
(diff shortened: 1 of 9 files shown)


mkarg:
Unfortunately we do not have agreed upon a generic 2.1.x branch, so unless we can guarantee that a 2.1.6 is never needed, we need to collect bug fixes in EE4J_8. :-(

mkarg:
Actually it is a bug fix, because the new Jenkins infrastructure fails to compile due to JavaDoc complaining about this (ERROR, not just WARNING).

chkal:
Ah, ok. I wasn't aware that this is causing issues. So strong +1 for including the fix in EE4J_8.

PR #732, opened 2019-02-05
New method #hasProperty

mkarg:
In some circumstances it might be beneficial to have a new method `InterceptorContext#hasProperty(String name)` which simply returns `true` if a property name is known to the context. As of JAX-RS 2.1 it is possible to workaround this by `InterceptorContext#getProperty(String name) != null`, but this might imply a performance penalty in case the implementation needs *additionional time* to actually retrieve the value of the property. Some implementations might benefit from having this method, and some application programmers might like the brevity of this new method.

This PR implements this as a `default` interface method, so implementations can decide to either implement a performance improved variant, or simply stay with the default solution without breaking backwards compatibility.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
@@ -298,7 +298,7 @@ public interface JAXRS {
          * Returns whether the property with the given name is configured, either explicitly or by default.
          *
          * @param name a {@code String} specifying the name of the property.
-         * @return {@code true} if the property has a value other than {@code null}, or {@code false} otherwise.
+         * @return {@code false} if no property exists matching the given name, {@code true} otherwise.
          * @since 2.2
          */
         default boolean hasProperty(String name) {
--- a/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
@@ -295,6 +295,17 @@ public interface JAXRS {
         Object property(String name);
 
         /**
+         * Returns whether the property with the given name exists.
+         *
+         * @param name a {@code String} specifying the name of the property.
+         * @return {@code true} if the property exists, or {@code false} otherwise.
+         * @since 2.2
+         */
+        default boolean hasProperty(String name) {
+            return property(name) != null;
+        }
+
+        /**
          * Convenience method to get the {@code protocol} to be used.
          * <p>
          * Same as if calling {@link #property(String) (String) property(PROTOCOL)}.
--- a/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/JAXRS.java
@@ -295,10 +295,10 @@ public interface JAXRS {
         Object property(String name);
 
         /**
-         * Returns whether the property with the given name exists.
+         * Returns whether the property with the given name is configured, either explicitly or by default.
          *
          * @param name a {@code String} specifying the name of the property.
-         * @return {@code true} if the property exists, or {@code false} otherwise.
+         * @return {@code true} if the property has a value other than {@code null}, or {@code false} otherwise.
          * @since 2.2
          */
         default boolean hasProperty(String name) {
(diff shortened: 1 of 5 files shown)


jansupol:
If this is changed on `InterceptorContext` should it not be changed on other places (such as Configuration) as well?

mkarg:
@jansupol Good idea. :-)

chkal:
@jansupol I agree!
@mkarg Should we include it in this PR?

mkarg:
Great, I'll push some more commits ontop. Stay tuned. :-)

mkarg:
Added `hasProperty(String)` in more places. Did I miss one? Please cast your votes (got dismissed by commit).

chkal:
I'm wondering whether the default implementation will work as expected in all cases, especially if the actual value of the property has been explicitly set to `null`.

For `InterceptorContext`, `ContainerRequestContext` and `ContainerResponseContext` the API docs of `setProperty()` explictly state that passing a `null` value is the same as calling `removeProperty()`.

For `Configuration` there seems to be no description on how `null` is handled and `JAXRS` defines that `null` is equivalent to the default value.

However, maybe this isn't relevant at all, because if you store `null` for a property, you cannot distinguish it from a missing property, because all you can do is call `getProperty()` which will return `null` in both cases.

mkarg:
> For `Configuration` there seems to be no description on how `null` is handled and `JAXRS` defines that `null` is equivalent to the default value.

Good catch. I should change the JavaDocs of `hasProperty` from "whether it exists" to "whether it has a set value", as the idea behind hasProperty is *not* in the sense of `hasPropertyName` but *solely* is in the sense of `hasPropertyValue` (so in case one can detect whether the default will chime in or not). Will do that later today. Thanks, Christian!

jansupol:
@chkal Can the Configuration property value be `null`? Properties is a `Map` and the Map [cannot ](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#put-K-V-)have `null` values.

chkal:
> @chkal Can the Configuration property value be `null`?

Good question. The API docs aren't mentioning it, so I would assume `null` is allowed.

> Properties is a `Map` and the Map [cannot ](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#put-K-V-)have `null` values.

`HashMap` allows `null` values. And `Map` doesn't force implementations to forbid `null` values.

mkarg:
Improved JavaDocs. Hope it is unambiguous now. :-)

jansupol:
Jersey actually does not allow the `Configuration` property value to be `null`. Jersey has the behavior aligned with the javadoc on `InterceptorContext`, `ContainerRequestContext` and `ContainerResponseContext`. If the value is `null`, the property is [removed](https://github.com/eclipse-ee4j/jersey/blob/6f425d9cd75192c09dc7e0233d35938d561e15b0/core-common/src/main/java/org/glassfish/jersey/model/internal/CommonConfig.java#L372). So the simple change in the javadoc introduces incompatible behavior for Jersey as well as possibly for any other implementation, and the logic would be needed to be reimplemented. 

Is allowing `null` on some places and not allowing them on another really the intention of JAX-RS API?

mkarg:
@jansupol It is not my intention to allow `null`. What JavaDoc change makes you think I somebody wants `null` to be allowed?

chkal:
> Is allowing `null` on some places and not allowing them on another really the intention of JAX-RS API?

I'm +1 for aligning the handling of `null` values in the API docs of `Application.setProperty()`. But we should file a separate pull request for this. WDYT?

mkarg:
*This* PR is solely about the addition of the `#hasProperty` method. It does *not* intend to change the `null` handling, but solely adds syntactic sugar (and room for performance improvements) on the existing methods and intents. So if somebody wants to *change* something different from that, a *new* PR is needed definitively.

If there *still* is a misunderstanding of the JavaDocs then please point me to the exact phrase that needs to be aligned more with the existing `#getProperty` method.

jansupol:
@mkarg See the `Configuration` javadoc: 

> @return {@code true} if the property with the specified property name has a value other than {@code null}

To me it sounds like `null` value is allowed.

mkarg:
> @mkarg See the `Configuration` javadoc:
> > @return {@code true} if the property with the specified property name has a value other than {@code null}
> To me it sounds like `null` value is allowed.

I did not intend to allow `null`. Any ideas for a better phrase?

jansupol:
1) Align with javadoc on `InterceptorContext`, `ContainerRequestContext` and `ContainerResponseContext`. That would, however, disallow `null`.
2) @return {@code false} if the `getProperty(name)` returns `null`, or {@code true} otherwise

mkarg:
@jansupol I followed your second proposal but left out the direct reference to the `getProperty` method to not imply that implementation must actually invoke that one. LGTY? +1?

jansupol:
@mkarg Can we align the javadoc of `JAXRS#hasProperty` ideally with `*Context#hasProperty`? Or at least with `Configuration#hasProperty`?

mkarg:
@jansupol Done. LGTY? +1?
@chkal Dismissed your vote by another commit. Please vote again. :-)

Issue #711, opened 2018-12-10
Clarify when to invoke MBR and MBW pipeline


NicoNes:
Hi guys,

AFAIK the current spec does not clearly state when to invoke MBR and MBW pipeline and this leads to situation like following:

Consider the given resource and providers below registered in a JAX-RS application:

```
@Path("test")
public class TestResource {
	@GET
	public Response getMethodWithEntity(Object entity) {
		return Response.ok().build();
	}
}
```
```
public class TestFilter implements ContainerRequestFilter {

	public static final String HAS_ENTITY = "Has entity";

	@Override
	public void filter(ContainerRequestContext containerRequestContext ) throws IOException {
		requestContext.setProperty(HAS_ENTITY, containerRequestContext .hasEntity());
	}
}
```
```
public class TestReaderInterceptor implements ReaderInterceptor {

	@Override
	public Object aroundReadFrom(ReaderInterceptorContext context) throws
 IOException, WebApplicationException {
		boolean hasEntity = (boolean) context.getProperty(TestFilter.HAS_ENTITY);
		if (!hasEntity) {
			System.out.println("Invocation of ReaderInterceptor even if there is 
no request entity to read");
		}
		return context.proceed();
	}
}
```
```
public class TestMBR implements MessageBodyReader<Object> {
	@Override
	public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
		return true;
	}

	@Override
	public Object readFrom(Class<Object> type, Type genericType, Annotation[] annotations, MediaType mediaType,
			MultivaluedMap<String, String> httpHeaders, InputStream entityStream)
			throws IOException, WebApplicationException {
		return new Object();
	}
}
```
When calling the GET method of `TestResource `from my browser without sending no request body I was
not expecting the `TestReaderInterceptor.aroundReadFrom()` neither `TestMBR.readFrom()`  to be invoked, but they are although there is no request body to process.

In contrast on the response side, the MBW pipeline is not invoked when there is no response body.

This behavior is the same whatever you are doing a GET, POST, PUT. I did it in both the RI and RESTEASY

So what I suggest is to clearly state that both MBR and MBW pipeline MUST be invoked only when there is a request/response body to process. 
In other words when no entity is identified in previous filter chain step, containerRequestContext.hasEntity()/containerResponseContext.hasEntity() returns false, MBR/MBW MUST not be invoked.

WDYT? Do you think it is acceptable ?

[...]

PR #741, opened 2019-02-10
2.1.x-Backport

mkarg:
This PR is a *backport* of select **bug fixes** between `2.1.5` and `master:HEAD` into 2.1.6-SNAPSHOT (`EE4J_8`).

**As these are no API changes, and as all these changes are already contained in `master` I propose a fast-track review period of just one day as per our [recently update committer rules](https://github.com/eclipse-ee4j/jaxrs-api/wiki/Committer-Conventions#minimum-length-of-review-period-before-merge--close-of-prs-and-closing-of-issues).**

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/ext/MessageBodyReader.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/ext/MessageBodyReader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2017 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -28,7 +28,8 @@ import java.lang.reflect.Type;
  *
  * A {@code MessageBodyReader} implementation may be annotated
  * with {@link javax.ws.rs.Consumes} to restrict the media types for which it will
- * be considered suitable.
+ * be considered suitable. The {@code MessageBodyReader} pipeline is executed if the matching 
+ * resource method declares an entity parameter or uses at least one {@link javax.ws.rs.FormParam}.
  * <p>
  * Providers implementing {@code MessageBodyReader} contract must be either programmatically
  * registered in an API runtime or must be annotated with
--- a/jaxrs-api/src/main/java/javax/ws/rs/ext/MessageBodyWriter.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/ext/MessageBodyWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2017 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -28,7 +28,8 @@ import java.lang.reflect.Type;
  *
  * A {@code MessageBodyWriter} implementation may be annotated
  * with {@link javax.ws.rs.Produces} to restrict the media types for which it will
- * be considered suitable.
+ * be considered suitable. The {@code MessageBodyWriter} pipeline is only invoked if there is 
+ * a non-null response entity.
  * <p>
  * Providers implementing {@code MessageBodyWriter} contract must be either programmatically
  * registered in an API runtime or must be annotated with


mkarg:
@mszabo-wikia Can you please renew your [Eclipse Contributor Agreement](https://accounts.eclipse.org/legal/eca/validation/727)?

mkarg:
@asoldano Yes the behaviour would be changed, but in the way of a bug fix, so I think it is correct to backport it.

mkarg:
@mszabo-wikia Máté apparently your current email address registered at the Eclipse Foundation (2721291+mszabo-wikia@****.noreply) is not the one you used for the sign-off (mszabo-wikia <mszabo@wikia-inc.com>). Which one will you continue to use? Either I need to fix the sign-off, or you need to register the correct one. Please respond. Thanks.

mkarg:
Now that we have reactivated the branch `EE4J_8` we should take care to keep it up-to-date wrt to bug fixes and clarifications. Hence I propose to cherry-pick (at least) the these two commits attached. If you think we should add more, let's keep things simple: Instead of having one PR per commit to cherry-pick, you can simply send a commit number to add to this PR. So we only need one voting for all. :-)

Regarding the Travis fail: This is a bug of Travis, I already opened a ticket with their support.

mkarg:
Thanks to Travis CI Support! All checks now have passed now.

spericas:
Although these are really good changes and clarifications, I'm on the fence about pushing them to Jakarta EE 8. As I understand it, existing EE 8 implementations will _not be updated_ for Jakarta EE 8, and the clarification regarding MBR's and MBW's may result in existing implementations to be "incompatible". I'm less worried about the other changes, but I see @asoldano has some concerns.

mkarg:
@spericas We had the discussion already, and the majority was for backporting it into Jakarta EE 8. The justification was that the existing products are not *incompatible* but simply *buggy*. Hence they SHOULD get fixed, but won't lose their *compliance state* even if not.

chkal:
@spericas You already agreed to backport those fixes. See https://github.com/eclipse-ee4j/jaxrs-api/pull/700#issuecomment-460337253 and https://github.com/eclipse-ee4j/jaxrs-api/issues/711#issuecomment-470120916. Any reason for changing your opinion about this?

spericas:
@chkal Yes, it's in my comment above. I didn't considered at the time that only API jars will be touched for Jakarta EE 8. Also note that I didn't vote -1, but simply shared the concern from @asoldano.

spericas:
> @spericas We had the discussion already, and the majority was for backporting it into Jakarta EE 8. The justification was that the existing products are not _incompatible_ but simply _buggy_. Hence they SHOULD get fixed, but won't lose their _compliance state_ even if not.

If you get the votes, you should push it. I'm +0.

mkarg:
@spericas @chkal Let's ask the one why is affected mostly...: @asoldano What is your vote on this?

asoldano:
@mkarg @spericas @chkal please give me a bit of time to check a couple of things here before answering.

jimma:
If this change will break backward compatibility somehow, I think we'd better to only fix this in 2.2 and
don't back port to 2.1.x branch. This will be helpful for user migrate to JakartaEE8 with less pain. If user's application, provider or filer uses jaxrs 2.1 version isCompatible() works, it's better that this still works in JakartaEE8 without change their code and rebuild all things.

spericas:
@asoldano If you can provide some feedback today, it would help with the release process. Thanks.

asoldano:
I've been thinking about this for some time and frankly I'm still concerned about backward compatibility here. My understanding is that JakartaEE 8 is expected to be equivalent to JavaEE 8. Vendors are most likely going to "move" to JakartaEE in minor releases, as it comes with no new features. Users will be expecting the same behaviour out of their applications after having moved to JakartaEE 8. Because of that, the change to MediaType is risky, as an application might be *directly* relying on it. This looks like an edge case and not really a *critical* fix, which is what would justify the merge on EE4J_8 branch.
As for the other change, as long as the TCK does not enforce the clarified behaviour and hence implementations that have been assuming a different behaviour so far would still be considered as compliant, I would be fine with merging. Even if that's not a great situation (spec implementors would have to be fixed while having a configuration option for retaining the previous backward compatible behavior), at least user application wouldn't be affected by default.
As a consequence, despite the changes here being good, I'm sorry but think I need to cast a -1 to this PR. These changes fit 2.2+ IMO.

spericas:
Thanks for the feedback @asoldano. In that case, I think we should close this PR. Do you agree @mkarg?

mkarg:
@spericas I agree to drop all commits that are -1, but do I understand that @asoldano is OK with eacb77e428160469ed0a91524432ac31fc1050f5. Is that correct? So I would reduce the PR to that one.

mkarg:
@asoldano @chkal Dropped the disputed commited from the PR. LGTY?

asoldano:
@mkarg I still see the MediaType changes included here.

mkarg:
> @mkarg I still see the MediaType changes included here.

Sorry, my fault. Fixed it.  Thanks! :-)

spericas:
@mkarg Once we get this one in, I'd like to initiate the release process and work with the TCK team.

mkarg:
@spericas Agreed. @asoldano +1?

asoldano:
@mkarg @spericas for the MBR/MBW related change, my comment (https://github.com/eclipse-ee4j/jaxrs-api/pull/741#issuecomment-510522135) still applies. So assuming the TCK won't be enforcing this, I'm +0 on this and we can proceed, considering approvals from others ;-)

mkarg:
Alessio's vote means that the backport will not happen, as we do not have enough +1's. @spericas Do not wait for this PR, it will not make it in time.

mkarg:
@eclipse-ee4j/eclipse-jaxrs I'm a bit disappointed to see that the majority of committers do not even vote with 0, but completely abstain from casting votes at all. :-(

mkarg:
@asoldano Can you please remove your -1 as we fixed your requested changes?

asoldano:
> @asoldano Can you please remove your -1 as we fixed your requested changes?

What do I need to do exactly? I "reviewed" again with a comment say +0 but that does not seem to have changed anything.

spericas:
Given that it has not received the necessary votes, I will initiate the release process for Jakarta EE 8 _without_ this PR. We can later decide to close it or revisit this discussion (if another release is necessary).

PR #744, opened 2019-02-16
Clarify in which situations MBW/MBR pipelines are executed (#711)

chkal:
This pull request addresses #711 by clarifying, in which situations MBW/MBR pipelines are executed. There was already some discussions about the reasoning for this wording in #711, so please refer to these comments if you want to learn more about the details.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/ext/MessageBodyWriter.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/ext/MessageBodyWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2017 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -27,7 +27,8 @@ import javax.ws.rs.core.MultivaluedMap;
  * Contract for a provider that supports the conversion of a Java type to a stream.
  *
  * A {@code MessageBodyWriter} implementation may be annotated with {@link javax.ws.rs.Produces} to restrict the media
- * types for which it will be considered suitable.
+ * types for which it will be considered suitable. The {@code MessageBodyWriter} pipeline is only invoked if there is 
+ * a non-null response entity.
  * <p>
  * Providers implementing {@code MessageBodyWriter} contract must be either programmatically registered in a JAX-RS
  * runtime or must be annotated with {@link javax.ws.rs.ext.Provider &#64;Provider} annotation to be automatically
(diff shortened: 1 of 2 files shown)


jaxrs-api/src/main/java/javax/ws/rs/ext/MessageBodyWriter.java
```diff
@@ -27,7 +27,8 @@
  * Contract for a provider that supports the conversion of a Java type to a stream.
  *
  * A {@code MessageBodyWriter} implementation may be annotated with {@link javax.ws.rs.Produces} to restrict the media
- * types for which it will be considered suitable.
+ * types for which it will be considered suitable. The {@code MessageBodyWriter} pipeline is only invoked if there is
```
mkarg: Lessions learned from other PRs: The term "there is" might lead to discussions about the meaning of `null` in future, as one could say "there was an entity, but it was null" vs "The invoked method has a return type of `void`". So maybe it would be more unambiguous if we say "...is only invoked if the invoked resource method returned a non-`null` value"?

chkal: I agree that using the term "non-null" is much better and I updated the pull request accordingly. See the updated version.

But I don't think that we should refer to the return value of the resource method directly. Mostely, because there are other components (like `ContainerResponseFilter`) which can alter the response entity as part of the processing chain. In the end it just matters what the response entity is just before the MBW chain is invoked. And at this point the entity could differ from the return value of the resource (if somebody called `ContainerResponseContext.setEntity()` for example).

mkarg: I do not see what components you think have the power to prevent the invocation *of the MBW* according to the JAX-RS 2.1 spec text and JavaDocs:
* A `ContainerResponseFilter` is invoked *after* an MBW is invoked, so this is irrelevant for your PR, as even a `null`-result of the filter could not prevent the invocation *of the MBW*.
* A `WriterInterceptor` is invoke itself *only if an MBW* is invoked as it is "wraps around calls to MBW#write"* according to its JavaDocs, so even if the interceptor produces `null` this would not stop *the MBW* from being called.

chkal: No, `ContainerResponseFilter` is invoked after the resource, but before the MBW. And `ContainerResponseFilter.setEntity(Object)` effectively allows to "change" the entity returned from the resource.

img

mkarg: LOL, I'm participating to JAX-RS spec since more than a decade but actually learn new things every day! ;-)

While it is totally weird to have response filters being called *before* the writer, I now understand your argument!

spericas: @chkal @mkarg There are some diagrams about this in Appendix C in the spec document. I think those are a bit more complete. It is indeed the case that response filters are called before the MBW chain.

mkarg:
@chkal Merge?

chkal:
@mkarg Yeah, sorry, almost forgot about this one. :smile:

PR #751, opened 2019-03-27
JAX-RS Extensions

mkarg:
This proposed feature standardizes an API for **JAX-RS Extensions**.

While JAX-RS components since ever could be *explicitly declared* to be loaded by `Application` classes, only *some* environments could auto-detect them *without* such explicit declaration of the application. In environments *not* supporting auto-detection this is not nice, because JAX-RS components might be provided on the classpath to be used with *any* application, i. e. *without* both, neither the application provider's knowledge or explicit grant, nor the runtime vendor's. Examples could be tracing and monitoring functionality, possibly from third parties (like cloud providers) *combining* JAX-RS *applications* with JAX-RS *runtimes*.

So JAX-RS Extensions provide a third origin of components on the classpath, besides runtime vendors and application providers: Deployers and Administrators.

JAX-RS Extensions are *not* a new kind of component, but simply are a declarative way to turn *any* kind of JAX-RS component into a forcefully **auto-loaded** component, just by declaring a marker interface in the way Java's ServiceLoader API works. In fact, the proposed mechanism does not actually instantiate components, but just detects the declared classes, and registers these into the JAX-RS configuration, so the already existing JAX-RS component lifecycle and scope will be in applied. That way this new extension is not only most flexible, but also least complex to add to existing runtimes: [A sample implementation](https://github.com/mkarg/jersey/commit/af3ea47681bac6f51d7d0fc5c81c969bd50e37e6) proofs that the new mechanism can be supported by e. g. Jersey without even touching Jersey's source code, just by adding a new external incubator module! Please also find the [demo application](https://github.com/mkarg/jaxrs-extensions-example) showing how Java SE bootstrapping works with this new feature.

**As CDI adoption is a long way to go, and as Jersey apparently can so easily support this, I'd like to propose that we adopt this rather small but really useful mechanism to JAX-RS 2.2.**

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/ext/Extension.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/ext/Extension.java
@@ -16,6 +16,8 @@
 
 package javax.ws.rs.ext;
 
+import java.util.ServiceLoader;
+
 /**
  * <p>
  * An extension to the JAX-RS runtime.
@@ -26,9 +28,9 @@ package javax.ws.rs.ext;
  * runtimes.
  * </p>
  * <p>
- * JAX-RS implementations MUST automatically register classes implementing this interface in
- * {@link javax.ws.rs.core.Configuration runtime configurations} if the fully qualified name of the implementing class
- * is found in the file {@code javax.ws.rs.ext.Extension} in the {@code META-INF/services} directory.
+ * JAX-RS implementations MUST automatically register a class in {@link javax.ws.rs.core.Configuration runtime
+ * configurations} if the class is a <em>provider class</em> for the {@link Extension} <em>service type</em> according
+ * to the rules set out in the description of {@link ServiceLoader}.
  * </p>
  *
  * @author Markus KARG (markus@headcrashing.eu)
--- a/jaxrs-api/src/main/java/javax/ws/rs/ext/Extension.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/ext/Extension.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2018 Markus KARG. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package javax.ws.rs.ext;
+
+/**
+ * <p>
+ * An extension to the JAX-RS runtime.
+ * </p>
+ * <p>
+ * All kinds of JAX-RS components (i. e. filters and interceptors, features, providers, etc.) can be marked as an
+ * extension to the JAX-RS runtime, hence will become available to <em>all</em> applications running on thusly extended
+ * runtimes.
+ * </p>
+ * <p>
+ * JAX-RS implementations MUST automatically register classes implementing this interface in
+ * {@link javax.ws.rs.core.Configuration runtime configurations} if the fully qualified name of the implementing class
+ * is found in the file {@code javax.ws.rs.ext.Extension} in the {@code META-INF/services} directory.
+ * </p>
+ *
+ * @author Markus KARG (markus@headcrashing.eu)
+ */
+public interface Extension {
+}


chkal:
I'm basically +1 for this. 

I just wonder if we need to clarify how the JAX-RS implementation is supposed to create instances of classes found in `META-INF/services` files. Especially, if the type is known to the CDI container, the JAX-RS implementation should obtain an instance from CDI to get injection and interception to work correctly.

However, maybe this behavior should be clarified in a more global context so that we get consistent behavior also for `Application.getClasses()` which also registers a component type instead of an actual instance.

BTW: Please note that the mechanism proposed here already exists in a very similar way in both Jersey (see [ForcedAutoDiscoverable](https://github.com/eclipse-ee4j/jersey/blob/master/core-common/src/main/java/org/glassfish/jersey/internal/spi/ForcedAutoDiscoverable.java) SPI) and RESTEasy (see this [part of the documentation](http://docs.jboss.org/resteasy/docs/3.6.3.Final/userguide/html_single/#d4e2835)).

mkarg:
@chkal I wouldn't like to explicitly *repeat* the registration mechanism here, to note open a special case. We instead should find a phrase that says something like "does basically the same as Application.getClasses()", which is mostly the technical truth and what we all want *for now*. Certainly CDI is the long term target, but that target has to be adressed by a different PR, as it has to work in other places, too. And for *now* (JAX-RS 2.2), actually, CDI is not even a SHOULD; in Jersey for example this mechanism currently works pretty well with other DI providers and I would like to keep that possibility unless CDI is a MUST. BTW, the idea of this PR is to standardize *exactly* what Jersey's `ForcedAutoDiscoverable` does. So it is in no competition, but it just provides a *common* way to reach that target in a *vendor-independent* way.

spericas:
My initial reaction is that this problem should be tackled when CDI becomes mandatory, not before. How does this relate to `@Provider`? An empty interface smells like a disguised annotation IMO. Not a fan of the proposal at this point.

mkarg:
@spericas This does not relate to `@Provider` at all, as the provider annotation is ignored at least in Java SE environments. It is an alternative solution that works the same in *all* environments. Both can be combined if wanted. It is available right now so people could use it without waiting months for CDI. BTW, it even works without implementing the empty interface, at least in Jersey. In fact people are waiting for this, if you remember the discussion about cloud tracing.

spericas:
> @spericas This does not relate to `@Provider` at all,

 Exactly, and that is not necessarily a good thing.

> as the provider annotation is ignored at least in Java SE environments. It is an alternative solution that works the same in _all_ environments. 

 I understand, and with CDI will have yet another alternative (as we do in MP environments today).

 > Both can be combined if wanted. It is available right now so people could use it without waiting months for CDI. BTW, it even works without implementing the empty interface, at least in Jersey. In fact people are waiting for this, if you remember the discussion about cloud tracing.

 I agree that this is necessary, I just don't agree with the need to add another mechanism before CDI. In any case, I'd love to hear what others think and perhaps I can be convinced to support this. Keeping an open mind as they say ;)

mkarg:
@spericas I agree that having lots of alternatives is a bad thing, so my favorite would be deprecating all other solutions and keeping just the new one ;-). The problem is just (a) the long wait time till CDI, and (b) scanning *all* loaded classes for `@Provider` being way slower than explicitly picking just the classes from the services folder.

andymc12:
I'm not clear about the use case for this feature.  Is the idea that a user might be migrating an existing JAX-RS application from an EE environment to SE and now the app fails because SE doesn't support `@Provider` automatic discovery?  Or that automatic discovery is expensive?

In either of these cases, there is already the solution of explicitly adding the providers via the `Application` `getClasses` or `getSingletons` methods.  Then the benefit to this feature is just style, right?  I.e. "I don't like hard-codings lists of providers, but I do like writing ServiceLoader files"?

If we are only adding style points, then I agree with @spericas that we should wait for the CDI approach.

If the use case is that the user cannot modify the application code (perhaps they are migrating an app developed by a proprietary third-party), then I think it would be nice to make the migration easier.  But iiuc, this new feature proposal would still require modification to the application code - all providers would need to be updated to implement the `Extension` interface, right?  If this is the use case we care about, then I think we need a different solution - something that allows providers to be declared without altering the application code.  MicroProfile Config would be an ideal solution here - it would allow the providers to be declared in various config sources, and it is consistent with declaring providers for the MicroProfile Rest Client.  Perhaps an optional dependency on MP Config would solve this?  A cruder approach that requires no new dependencies might be to use a defined system property to specify a list of provider classes (e.g. `javax.ws.rs.providers=com.mycompany.MyProvider1,...`).  

I don't see a lot of value with this proposal, but maybe I'm missing something.

mkarg:
@andymc12 Yes, you miss something. `Extension`s are *neither* provided by applications *nor* by JAX-RS implementations, but by cloud services. Example: Party A is writing a JAX-RS application, and I sells it in a shop. Party B buys from that shop. Party A and Party B do not know each other then. Party B builds a docker container using a pre-build docker-hub-hosted base-image for running JAX-RS JARs provided by Party C to run that application. Party C bundles Jersey into that image, but does not know neither Party A nor Party B. Party C wants the base image to be able to trace requests, so they bundle an `Extension` which will be auto-loaded always independent of any application. Party B does not know how Party C makes the tracing, but they love it. Party A does not even know of the tracing. So no existing solution will work in this scenario (which is real-world; we do exactly that; so we need it *now*). Currently Part C has to use Jersey's `ForceAutoDiscoverable`, which prevents replacing Jersey by RESTeasy. Solution: Using `Extension` instead.

andymc12:
@mkarg Ah, that is a more interesting use case.  Thanks for the clarification.

In the javadoc, I would suggest changing `META-INF/services` to `ServiceLoader`, since the META-INF/services approach is changed in Java 9+ with JPMS (`provides X with Y` in module-info).

I also wonder if you could accomplish the same thing without the new interface.  Maybe automatically install providers based on their own interface name (e.g. `META-INF/services/javax.ws.rs.container.ContainerRequestFilter` or `provides java.ws.rs.ext.MessageBodyWriter with ...`) or modify the provider interfaces to extend the `Extension` interface so that extenders just need to implement the provider interfaces and then expose them as service implementations.

arjantijms:
@andymc12 

This may also solve the issue when essentially portable features are provided via jar files that are put in an internal server path (e.g. `glassfish/modules` in Payara). These jar files are not scanned for annotations, so META-INF/services has to be used. JAX-RS has no portable solution for that.

See e.g. https://github.com/payara/Payara/blob/master/appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/jaxrs/RolesAllowedAutoDiscoverable.java#L64

If however `javax.ws.rs.container.DynamicFeature` would be defined as a CDI bean in the feature, which is a step beyond what is currently proposed* for 3.0 I think, then this would would perhaps not be needed, or only until we get to 3.0 or 3.0+.

(* My understanding was that the proposal for 3.0 was that resources would become CDI beans by default, and that all @Context usage would be replaced by CDI @Inject. 

A step further would that all, or most, other JAX-RS artefacts would be CDI beans as well, meaning the JAX-RS runtime would ask CDI for all beans implementing `DynamicFeature`, potentially making @Provider a qualifier, or even do away with it.

Though I love to see that next step, I don't think we've discussed that)

mkarg:
@andymc12 Updated the JavaDocs according your request. I also removed the word "implement", so it should be possible to load classes that do not declare `implements Extension` according to the description of `ServiceLoader`, as long as the class is *assignable* to `Extension` - which all components are as `Extension` has no members by exactly that intention. On the other hand, I do not see a problem with adding "implements Service" to existing services to get them turned into JAX-RS 2.2 compliant extensions.

jimma:
@mkarg Do we need to make this work for jaxrs client too ?  Then the features, interceptors etc can be registered automatically to client without writing lines code to register.

mkarg:
@jimma Yes, the intention of this PR is to make this work *everywhere*, even in JAX-RS clients.

jimma:
@mkarg  Then would it better to add more info to this api to tell this is only required to load for client or server ? Considering this works with Bootstrap api, if both jaxrs client and server runtime packaged in an executable fat jar file, we may need register different class(auto load different Extension) to client from server runtime.

mkarg:
@jimma I don't understand your point. My proposal is that client and server both MUST respect this new mechanism, and the JavaDocs actually tell exactly this already. What actually are you missing?

mkarg:
@jimma I think we do not need to *add* that, as this already exists: https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/ConstrainedTo.java.

jaxrs-api/src/main/java/javax/ws/rs/ext/Extension.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) 2018 Markus KARG. All rights reserved.
```
chkal: I guess this should be 2019?

chkal:
> This may also solve the issue when essentially portable features are provided via jar files that are put in an internal server path (e.g. `glassfish/modules` in Payara). These jar files are not scanned for annotations, so META-INF/services has to be used. JAX-RS has no portable solution for that.

Exactly. I ran into this issue so many times and it would be great to have a portable solution for that. As [mentioned earlier](https://github.com/eclipse-ee4j/jaxrs-api/pull/751#issuecomment-477447809), almost all JAX-RS implementation have some non-portable solution for that.

> If however `javax.ws.rs.container.DynamicFeature` would be defined as a CDI bean in the feature, which is a step beyond what is currently proposed* for 3.0 I think, then this would perhaps not be needed, or only until we get to 3.0 or 3.0+.

I think that using CDI as a replacement for ServiceLoader can be tricky. I recommend [this great blog post](https://struberg.wordpress.com/2012/05/11/the-tricky-cdi-serviceprovider-pattern/) by @struberg about this topic and some pitfals.

jaxrs-api/src/main/java/javax/ws/rs/ext/Extension.java
mkarg: I can change it to 2019, but indeed the code was written in 2018 already.

spericas:
> On Apr 6, 2019, at 12:06 PM, Christian Kaltepoth <notifications@github.com> wrote:
> 
> I think that using CDI as a replacement for ServiceLoader can be tricky. I recommend this great blog post <https://struberg.wordpress.com/2012/05/11/the-tricky-cdi-serviceprovider-pattern/> by @struberg <https://github.com/struberg> about this topic and some pitfals.

 Yes, using alternatives introduce some challenges, but even though this adds some complication, it does not prevent the use of CDI for this. As explained in that blog’s comments, there are ways of getting all the beans (including alternatives) and then work with that set. A lot of apps and frameworks are using CDI for this purpose.

— Santiago

mkarg:
So how to go on? I still think my PR is very beneficial in the *short time* as it *immediately* puts an existing Jersey feature on the cross-vendor level, and it does *not* stand in the way of neither CDI nor `@Provider` (in fact, one could write a cross-vendor JAX-RS extension using my PR that actually would register all `@Provider` classes even in a pure JavaSE or Client environment). So +1 from anybody would be great. :-)

spericas:
As I said, I have mixed feelings about this one, but won't oppose to it. I'm +0 (or is it -0?).

mkarg:
@arjantijms @andymc12 +1?

anthonyvdotbe:
@mkarg #947 is what its title says: an alternative to #933/#751.
My main concern with the current solution is exemplified by the following scenario: say I have an EAR which contains: (1) multiple REST applications, all with an explicit `Application` subclass, and (2) a utility library (possibly as a (transitive) dependency) with all kinds of `Feature` & other JAX-RS interface implementations, some of which are annotated with `Extension`. Now when this EAR is deployed on a server with JAX-RS 3.0, the `Extension`-annotated implementations aren't registered. However, when the server is upgraded to JAX-RS 3.1, suddenly the `Extension`-annotated implementations get registered. So as I see it: (1) this solution is a backward-incompatible change, and (2) there's no way to opt-out of it.
It also seems inflexible to me: what if I have a `Feature` which some people want to deploy server-wide for all applications, while others want to embed it in their WAR for use by just their `Application`?
On a final note, even with this PR, I believe the solution is still underspecified.

(By the way, I think it would be a good rule to only allow approving PRs that contain both the required spec and API changes. Sometimes the spec changes raise concerns about the API changes & vice versa. When there's separate PRs, as in this case, it seems there's little to no room for altering the already-merged API/spec changes, even though that might be the right thing to do.)

PR #592, opened 2018-02-24
Fix issue 581: Problem with RuntimeDelegate.getInstance()

NiklasMehner:
This implements a fix for issue #581: Problem with RuntimeDelegate.getInstance().

Issue #581, opened 2018-02-02
Problem with RuntimeDelegate.getInstance()


glassfishrobot:
`RuntimeDelegate.getInstance()` currently returns an implementation namespaced under [org.glassfish.jersey.internal.RuntimeDelegateImpl](https://github.com/jax-rs/api/blob/1181f3b9caf69e36362d86bb01fc6e89083824ba/jaxrs-api/src/main/java/javax/ws/rs/ext/RuntimeDelegate.java#L70). However, that implementation does nothing except [throw an exception](https://github.com/jersey/jersey/blob/88c6d7d9009223cb1d769f6d76236315d7051434/core-common/src/main/java/org/glassfish/jersey/internal/RuntimeDelegateImpl.java#L68). Is this intentional? Is there a reason why it's not returning [this](https://github.com/jersey/jersey/blob/12e5d8bdf22bcd2676a1032ed69473cf2bbc48c7/core-server/src/main/java/org/glassfish/jersey/server/internal/RuntimeDelegateImpl.java) implementation?

NiklasMehner:
I have create PR https://github.com/eclipse-ee4j/jaxrs-api/pull/592 for this issue.

mkarg:
Closing issue due to merged PR #592.

PR #755, opened 2019-04-20
[581] Remove Jersey as default implementation

mkarg:
This PR finishes the work started by PR #592: While #592 solved issue #581 in one location, this new PR solves the same issue on all remaining locations - it removes the dependency to Jersey.

Also this PR removes the fallback class name parameter completely, as it is unused now, as suggested by @andymc12.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/client/ClientBuilder.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/client/ClientBuilder.java
@@ -43,10 +43,6 @@ public abstract class ClientBuilder implements Configurable<ClientBuilder> {
      * {@link ClientBuilder#newBuilder()}.
      */
     public static final String JAXRS_DEFAULT_CLIENT_BUILDER_PROPERTY = "javax.ws.rs.client.ClientBuilder";
-    /**
-     * Default client builder implementation class name.
-     */
-    private static final String JAXRS_DEFAULT_CLIENT_BUILDER = "org.glassfish.jersey.client.JerseyClientBuilder";
 
     /**
      * Allows custom implementations to extend the {@code ClientBuilder} class.
@@ -62,8 +58,7 @@ public abstract class ClientBuilder implements Configurable<ClientBuilder> {
      */
     public static ClientBuilder newBuilder() {
         try {
-            Object delegate = FactoryFinder.find(JAXRS_DEFAULT_CLIENT_BUILDER_PROPERTY,
-                    JAXRS_DEFAULT_CLIENT_BUILDER, ClientBuilder.class);
+            Object delegate = FactoryFinder.find(JAXRS_DEFAULT_CLIENT_BUILDER_PROPERTY, ClientBuilder.class);
             if (!(delegate instanceof ClientBuilder)) {
                 Class pClass = ClientBuilder.class;
                 String classnameAsResource = pClass.getName().replace('.', '/') + ".class";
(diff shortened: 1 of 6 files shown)


mkarg:
@andymc12 Good idea. I have added fea3b2e78e583b8ea879e2e0c537e2463d283bc9 to remove the parameter.

mkarg:
@andymc12 @chkal I found two more code locations where Jersey was mentioned, so i fixed those two and rewrote the commit history. I assume you still are both +1, but I have to bother you once more to press the +1 button in Github... Hope you don't mind. :-)

mkarg:
@chkal Sigh. Done. Thanks for the pointer. :-)

jansupol:
This definitely breaks current versions of Jersey, so it should be in 2.2, IMO, not in a minor version of JAX-RS change.

mkarg:
@jansupol I see that as a bug solely of Jersey, which is nothing that shall be dealt with in a cross-vendor standard.  Cannot this be fix rather quickly in Jersey?

jansupol:
We must fix this in Jersey, but it is a backward incompatibility in JAX-RS, a new API will break an old implementation. Users will grab the latest 2.1.x and won't understand why it stops working suddenly

jansupol:
For a reason, jersey does not specify `META-INF/services/*` for the implementation classes, which is why #592 was merged into a master. I am not sure, why the services are not specified, perhaps it is difficult to create automated tests for it when the API has a fallback to the default implementation classes, but the API should not introduce backward incompatible changes in a dot release, no matter Jersey will fix it in a release today or in a year.

jansupol:
I have discussed this with the application server guys. The missing `META-INF/services/*` was by design in many projects developed in Oracle, to allow other implementations to be grabbed by the ServiceLoader if available and the RIs were used only when the ServiceLoaded did not find any implementation specified by  `META-INF/services/*`.

mkarg:
I understand the concerns Jan raises and we should try not to break things *easily*. But just to make clear: This fix in JAX-RS is *not* a change in the *API* of JAX-RS. The behavior Jersey relies upon is undocumented *internal* utility code, not even public classes, and it never was the intend of the JAX-RS Expert Group members that any implementation would rely upon. So while I agree that we could withhold this fix until JAX-RS 2.2, Jersey should be fixed ASAP. I am definitively -1 for waiting several months. But this is just my personal opinion. What do the other vendors think? What is IBM's, Red Hat's, and Apache's opinion?

chkal:
IMO we should modify the PR to target the `master` branch instead. I absolutely agree that having a Jersey implementation class hardcoded in the JAX-RS API is weird, and we should change that. But it is not really a bug that should go to `EE4J_8`.

mkarg:
@andymc12 @asoldano Ok for keeping the bug open in 2.1.x but only fixing it in 2.2?

andymc12:
I'm fine with waiting until JAX-RS 2.2.  Thanks.

mkarg:
As requested this PR now aims for `master` (i. e. JAX-RS 2.2), not `EE4J_8` (i. e. JAX-RS 2.1.6) anymore. As a consequence, please use the review tool once more to cast your votes.

jansupol:
I refrain from my vote as I am not sure this necessarily a good thing. It's nice to be independent, but having a fallback implementation (not necessarily Jersey) was there for a reason since 1.0.

mkarg:
@jansupol If someone remembers a good reason why *the standard* itself should provide a fallback, it would be a good thing to have a second PR actively introducing a *configurable* fallback mechanism instead of making a direct reference to one particular product. Anyways, IIRC you mentioned that the GlassFish team asked for such a fallback, so I wonder why *they* don't add a such a facility into their *product* instead?

jansupol:
@mkarg No, I did not mention the Glassfish team. Anyway, it is much less work much less error prone on the API side than multiply the effort on multiple application servers.

jansupol:
Actually, since `META-INF/services` does not support priorities, a configurable fallback mechanism sounds like something to overcome that issue.

PR #762, opened 2019-05-05
Clarify that ParamConverters must be preferred and work with all types

chkal:
The `ParamConverter` API is mentioned only briefly in section 3.2 of the specification document like this:

> Valid parameter types for each of the above annotations are listed in the corresponding Javadoc, however in general (excluding `@Context` ) the following types are supported:
> 1. Types for which a `ParamConverter` is available via a registered `ParamConverterProvider`. See
Javadoc for these classes for more information.
> 2. [...]

Unfortunately JAX-RS implementations aren't handling all aspects of this API in the same way. RESTEasy for example completely ignores `ParamConverters` for all primitive types (see [RESTEASY-2222](https://issues.jboss.org/browse/RESTEASY-2222) for details). Although the spec isn't very explicit about which types are supported, I don't see any reason why primitive types should be handled differently. Jersey seems to support the `ParamConverter` API for all types, which is exactly what I would expect from an JAX-RS implementation.

This PR adds the following clarifications to the `ParamConverter` API docs:

  * You can register converters for all Java types (including primitives)
  * If there is a converter, it must be preferred over all other conversion strategies mentioned in the spec.

Let me know what you think.

Please note that I created this pull request for the `EE4J_8` as I consider it to be a clarification. Feel free to mention it if you disagree with this.

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/ext/ParamConverter.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/ext/ParamConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2017 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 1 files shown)


chkal:
> Merging into 2.1.x would break existing products

This depends on whether you consider the current behavior of RESTEasy to be compliant with the current wording of the spec or not. IMO it is not. The clarification only makes this a bit more obvious. But that's just my view. Others may feel different about this.

mkarg:
@chkal I think you remember the similar idiscussion about using undocumented behavior of the API classes by Jersey, and there it was a huge problem that it would break. So to be fair, we now have to ask RESTEasy what *they* think. We cannot say Jersey gets an exemption for their bug, but RESTEasy is not. Basically I am +1 for breaking *always without further discussion*, but *including* Jersey, not just its competitors.

chkal:
Agreed! Of course all implementations should be treated in the same way. No doubt about that. However, actively removing behavior that breaks an implementation is something completely different from clarifying what the spec is actually expecting from implementations. But of course I would be happy to hear thoughts from the RESTEasy team. That's why I invited @asoldano for review. :-)

spericas:
I'm with @chkal on this one. This is a clarification.

jansupol:
With the recent events, it does not look like a new 2.1.x is to be released. So while this seems to be a clarification, I'd agree with putting this to the master branch.

chkal:
I don't think that the recent events are relevant for the EE4_8 branch. Especially because Jakarta EE 8 will most likely be built from this branch.

mkarg:
@chkal I share the opinion of Christian and Santiago. Let's put it into both, EE4J_8 *and*
 master.

spericas:
Looks like we have some agreement. OTOH, it would be nice to hear from other implementors before pulling the trigger IMO.

asoldano:
@chkal @spericas well, this is fixable in RESTEasy; I'm not sure about the reasons that lead to this behavior in RESTEasy, perhaps some type of performance enhancement. Anyway, we can live with the clarification.

chkal:
Could everyone who is fine with this clarification please approve the PR. This makes it easier to track the votes. Thanks!

chkal:
Thanks everyone for reviewing. I'll create a separate pull request to get this change into the `master` branch.

PR #774, opened 2019-06-30
Create Exceptions

### LINKED ISSUE #772: NotImplementedException to go with Response.Status.NOT_IMPLEMENTED & friends
Provide equivalent exception types for the new v2.0 Response.Status fields.

jeyvison:
Closes #772 .

A few points:

- I didn't add an exception to PAYMENT_REQUIRED on purpose. Didn't know if i was supposed to.
- I was thinking about move all the exceptions to a exception package but i would generate a Major bump. If it's ok, let me know and i change them to there.
- I'm reeealy not sure about the javadoc descriptions on the exception classes. Let me know if and how i need to fix them.

This is my first PR so i'm expecting a lot of fixes ^^ Let me know what you wanna change :)

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
@@ -1,0 +1,105 @@
+/********************************************************************************
+ * Copyright (c) 2019 Jeyvison Nascimento
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ ********************************************************************************/
+
+package jakarta.ws.rs;
+
+import jakarta.ws.rs.core.Response;
+
+/**
+ * A runtime exception indicating that the server, while acting as a gateway or proxy,
+ * received an invalid response from the upstream server it accessed in attempting to fulfill the request.
+ *
+ * @author Jeyvison Nascimento (jeynoronha@gmail.com)
+ * @since 2.2
+ */
+public class BadGatewayException extends ServerErrorException {
+
+    private static final long serialVersionUID = 7264647684649480265L;
+
+    /**
+     * Construct a new bad gateway exception.
+     */
+    public BadGatewayException() {
+        super(Response.Status.BAD_GATEWAY);
+    }
+
+    /**
+     * Construct a new bad gateway exception.
+     *
+     * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
+     */
+    public BadGatewayException(final String message) {
+        super(message, Response.Status.BAD_GATEWAY);
+    }
+
+    /**
+     * Construct a new bad gateway exception.
+     *
+     * @param response error response.
+     * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 502}.
+     */
+    public BadGatewayException(final Response response) {
+        super(validate(response, Response.Status.BAD_GATEWAY));
+    }
+
+    /**
+     * Construct a new bad gateway exception.
+     *
+     * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
+     * @param response error response.
+     * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 502}.
+     */
+    public BadGatewayException(final String message, final Response response) {
+        super(message, validate(response, Response.Status.BAD_GATEWAY));
+    }
+
+    /**
+     * Construct a new bad gateway exception.
+     *
+     * @param cause the underlying cause of the exception.
+     */
+    public BadGatewayException(final Throwable cause) {
+        super(Response.Status.BAD_GATEWAY, cause);
+    }
+
+    /**
+     * Construct a new bad gateway exception.
+     *
+     * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
+     * @param cause the underlying cause of the exception.
+     */
+    public BadGatewayException(final String message, final Throwable cause) {
+        super(message, Response.Status.BAD_GATEWAY, cause);
+    }
+
+    /**
+     * Construct a new bad gateway exception.
+     *
+     * @param response error response.
+     * @param cause the underlying cause of the exception.
+     * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 502}.
+     */
+    public BadGatewayException(final Response response, final Throwable cause) {
+        super(validate(response, Response.Status.BAD_GATEWAY), cause);
+    }
+
+    /**
+     * Construct a new bad gateway exception.
+     *
+     * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
+     * @param response error response.
+     * @param cause the underlying cause of the exception.
+     * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 502}.
+     */
+    public BadGatewayException(final String message, final Response response, final Throwable cause) {
+        super(message, validate(response, Response.Status.BAD_GATEWAY), cause);
+    }
+}
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
@@ -18,7 +18,7 @@ import jakarta.ws.rs.core.Response;
  * received an invalid response from the upstream server it accessed in attempting to fulfill the request.
  *
  * @author Jeyvison Nascimento (jeynoronha@gmail.com)
- * @since 2.2
+ * @since 3.1
  */
 public class BadGatewayException extends ServerErrorException {
 
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 Jeyvison Nascimento
+ * Copyright (c) 2020 Jeyvison Nascimento
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/GatewayTimeoutException.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/GatewayTimeoutException.java
@@ -15,10 +15,10 @@ import jakarta.ws.rs.core.Response;
 
 /**
  * A runtime exception indicating that the server, while acting as a gateway or proxy,
- * received an invalid response from the upstream server it accessed in attempting to fulfill the request.
+ * did not get a response in time from the upstream server that it needed in order to complete the request. .
  *
  * @author Jeyvison Nascimento (jeynoronha@gmail.com)
- * @since 2.2
+ * @since 3.1
  */
 public class GatewayTimeoutException extends ServerErrorException {
 
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/GatewayTimeoutException.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/GatewayTimeoutException.java
@@ -1,0 +1,105 @@
+/********************************************************************************
+ * Copyright (c) 2019 Jeyvison Nascimento
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ ********************************************************************************/
+
+package jakarta.ws.rs;
+
+import jakarta.ws.rs.core.Response;
+
+/**
+ * A runtime exception indicating that the server, while acting as a gateway or proxy,
+ * received an invalid response from the upstream server it accessed in attempting to fulfill the request.
+ *
+ * @author Jeyvison Nascimento (jeynoronha@gmail.com)
+ * @since 2.2
+ */
+public class GatewayTimeoutException extends ServerErrorException {
+
+    private static final long serialVersionUID = 6523870074504974712L;
+
+    /**
+     * Construct a new gateway timeout exception.
+     */
+    public GatewayTimeoutException() {
+        super(Response.Status.GATEWAY_TIMEOUT);
+    }
+
+    /**
+     * Construct a new gateway timeout exception.
+     *
+     * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
+     */
+    public GatewayTimeoutException(final String message) {
+        super(message, Response.Status.GATEWAY_TIMEOUT);
+    }
+
+    /**
+     * Construct a new gateway timeout exception.
+     *
+     * @param response error response.
+     * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 504}.
+     */
+    public GatewayTimeoutException(final Response response) {
+        super(validate(response, Response.Status.GATEWAY_TIMEOUT));
+    }
+
+    /**
+     * Construct a new gateway timeout exception.
+     *
+     * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
+     * @param response error response.
+     * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 504}.
+     */
+    public GatewayTimeoutException(final String message, final Response response) {
+        super(message, validate(response, Response.Status.GATEWAY_TIMEOUT));
+    }
+
+    /**
+     * Construct a new gateway timeout exception.
+     *
+     * @param cause the underlying cause of the exception.
+     */
+    public GatewayTimeoutException(final Throwable cause) {
+        super(Response.Status.GATEWAY_TIMEOUT, cause);
+    }
+
+    /**
+     * Construct a new gateway timeout exception.
+     *
+     * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
+     * @param cause the underlying cause of the exception.
+     */
+    public GatewayTimeoutException(final String message, final Throwable cause) {
+        super(message, Response.Status.GATEWAY_TIMEOUT, cause);
+    }
+
+    /**
+     * Construct a new gateway timeout exception.
+     *
+     * @param response error response.
+     * @param cause the underlying cause of the exception.
+     * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 504}.
+     */
+    public GatewayTimeoutException(final Response response, final Throwable cause) {
+        super(validate(response, Response.Status.GATEWAY_TIMEOUT), cause);
+    }
+
+    /**
+     * Construct a new gateway timeout exception.
+     *
+     * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
+     * @param response error response.
+     * @param cause the underlying cause of the exception.
+     * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 504}.
+     */
+    public GatewayTimeoutException(final String message, final Response response, final Throwable cause) {
+        super(message, validate(response, Response.Status.GATEWAY_TIMEOUT), cause);
+    }
+}
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/GatewayTimeoutException.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/GatewayTimeoutException.java
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 Jeyvison Nascimento
+ * Copyright (c) 2020 Jeyvison Nascimento
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/HttpVersionNotSupportedException.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/HttpVersionNotSupportedException.java
@@ -18,21 +18,21 @@ import jakarta.ws.rs.core.Response;
  * the HTTP protocol version that was used in the request message.
  *
  * @author Jeyvison Nascimento (jeynoronha@gmail.com)
- * @since 2.2
+ * @since 3.1
  */
 public class HttpVersionNotSupportedException extends ServerErrorException {
 
     private static final long serialVersionUID = -2162737983746617138L;
 
     /**
-     * Construct a new bad client request exception.
+     * Construct a new http version not supported exception.
      */
     public HttpVersionNotSupportedException() {
         super(Response.Status.HTTP_VERSION_NOT_SUPPORTED);
     }
 
     /**
-     * Construct a new bad client request exception.
+     * Construct a new http version not supported exception.
      *
      * @param message the detail message (which is saved for later retrieval by the {@link #getMessage()} method).
      */
@@ -41,7 +41,7 @@ public class HttpVersionNotSupportedException extends ServerErrorException {
     }
 
     /**
-     * Construct a new bad client request exception.
+     * Construct a new http version not supported exception.
      *
      * @param response error response.
      * @throws IllegalArgumentException in case the status code set in the response is not HTTP {@code 505}.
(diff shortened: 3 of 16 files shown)


mkarg:
@jeyvison Is this just *releated* to issue #772 or will it *fix* it? In the latter case, please write "Closes #772" in the description, and Github will automatically close the issues after the merge. This solves the problem that people merge PRs but forget to close the issues... ;-)

mkarg:
@jeyvison
* Regarding continuous-integration/travis-ci/pr: I have filed a help request with Travis CI Support. Apparently they have trouble downloadig JDK 11. Let's wait for their response.
* Regarding eclipsefdn/eca: The [ECA Validation Tool](https://www.eclipse.org/contribute/cla) says: *There is no valid ECA on file for jeyvison.nascimento@liferay.com*. Please check back with the EMO if you think that this is wrong. Possibly you signed with a different email address?

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (c) 2012, 2019 Oracle and/or its affiliates. All rights reserved.
```
mkarg: This is wrong. Neither are you working for Oracle, nor was your code written in 2012. Please provide the truth here.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
```diff
@@ -0,0 +1,110 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package jakarta.ws.rs;
```
mkarg: @eclipse-ee4j/eclipse-jaxrs Do we really want to collect all kinds of error-code-exceptions as separate classes, and do we really want to have them directly in our root namespace?

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
```diff
@@ -0,0 +1,110 @@
...
+
+import jakarta.ws.rs.core.Response;
+
+/**
+ * A runtime exception indicating that the server, while acting as a gateway or proxy,
+ * received an invalid response from the upstream server it accessed in attempting to fulfill the request.
+ *
+ * @author Jeyvison Nascimento
```
mkarg: Adding your email address (like `Jeyvision Nascimento (jeyvision.nascimento@liferay.com)`) allows readers of your code to get in touch with you. :-)

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
```diff
@@ -0,0 +1,110 @@
...
+ * @since 2.0
```
mkarg: This is wrong. Your code did not exist in 2.0. You must always mention the first API version that contained your code. Here, it would be `2.2`.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class BadGatewayException extends ServerErrorException {
+
+    private static final long serialVersionUID = 7264647684649480265L;
```
mkarg: @eclipse-ee4j/eclipse-jaxrs Do we really want to use `serialVersionUID`s? It implies that if we change the class we MUST modify the UID *always*.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
chkal: I agree that defining exceptions for all status codes is an overkill. IMO we should only add exceptions which users want to throw/catch in real world applications. I'm fine with `NotImplementedException`, but I don't think that somebody needs a `ProxyAutheticationRequiredException` in real applications.

mkarg: > I don't think that somebody needs a `ProxyAutheticationRequiredException` in real applications.

In fact I think that it would be even *wrong* to have this particular exception, as it is not an *exceptional state* to request proxy authentication.

chkal: Well, the corresponding HTTP status code is `407`, so it is classified as a client error. However, the actually need to throw this kind of error on the server side or handle it in the client case is extremely rare. Therefore, I don't think that it is worth adding it. Some for most of the other ones.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: Sure, I just used the same i found in BadRequestException. I should only remove this oracle part right?

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: On it as soon as i figure out how to use my personal email in this PR  instead of liferay's one.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: On it.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
chkal: You could use something like:

    Copyright (c) 2019 Jeyvison Nascimento

Or a more generic one like:

    Copyright (c) 2019 Contributors to the Eclipse Foundation

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: Agreed to all you said. I would even say that JAX-RS itself should not contain any kind of "pure sugar" classes; we should instead have a separate "official extensions" place to put those "common comfort stuff", but this is a different story.

Regarding 407: It is an *HTTP* error certainly, but it is *not* an exceptional state of the calling REST application. The application cannot know that the server expects authentication, so it sends the request. The server answers 407, so the application knows *now* that it needs to ask the user for credentials, and will request a second time. This is more like a *login handshake* than handling an actual *programming error*. Hence, no need for an *exception*, but possibly a need for some kind of new API (like a callback asking for credentials).

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: @jeyvison Here is the long answer: https://www.eclipse.org/projects/handbook/#ip-copyright-headers

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: @jeyvision You need to set up your local git config correctly and then rewrite the git history
See https://help.github.com/en/articles/setting-your-commit-email-address-in-git and https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit/28845565.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: No need to reply until fixed. It will just involve all readers without actual need to react. ;-)

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
chkal: > Regarding 407: It is an _HTTP_ error certainly, but it is _not_ an exceptional state of the calling REST application.

I see your point! And I think you are correct. However, I also think that 407 is a low level status code and not very often used. So I don't like the idea of added a special exception type for this.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: It's used in other exceptions classes but i get the point if it's really needed.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: fixed.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: fixed

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: @jeyvison What do you think? Do you agree to remove those exceptions that are not essentially low-level status?

jeyvison: copyright and `@since`  but i won't resolve this conversation 'till we decide if we keep or remove `ProxyAutheticationRequiredException` . For what you guys are saying i think we will remove it right?

jeyvison: > @jeyvison What do you think? Do you agree to remove those exceptions that are not essentially low-level status?

I agree. We should keep the code clean and add more exceptions as necessary later.

jeyvison: Thinking a bit more about this, i think all the exceptions like `ProxyAutheticationRequiredException` should stay. If we think exceptions as recoverable errors it makes sense for the client to send a second request with the proxy information in response to an exception (that would work as the _login handshake_ @mkarg mentioned before.  IMHO, an API for that would be an overkill.

I know that all can be verified by the response code but exceptions are more direct and really serves the purpose here.

mkarg: But this kind of handshake is the job of the JAX-RS framework, not the job of the business application. I cannot see why each application programmer should take care of that.

andymc12: For a 407, doesn't that mean that the client is unauthenticated to the proxy?  That could mean that the client needs to authenticate, but could also mean that the credentials that the client sent are incorrect.  In that case, the application would need to react accordingly - i.e. prompt the user to re-enter their password, etc.

mkarg: This is correct, but it does not imply that an exception is the right tool of chooice. Like for other parts of JAX-RS, e. g. normal (i. e. non-proxy) client authentication, this should not be part of the business application code (which an exception always would be), but part of a separate layer implemented either by an additional Client API, or by means of filters. Here is an example what I mean: http://www.adam-bien.com/roller/abien/entry/client_side_http_basic_access.

mkarg:
@jeyvison FYA all checks pass now (Travis CI Support fixed the problem with JDK11 on their side)

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: From this example you provided, if the user provides a wrong user or password the error still needs to be treated by the developer at business level right?

andymc12: Yes, that is my understanding.

mkarg: I do not think so. If the application registers a callback for handling this case (like showing a login dialog), there still is no need for a business level *exception*. In case of 407 the callback simply would be invoked more often, just as a browser shows a login dialog unless the password is correct.

jeyvison: Hmmm, so , just so i'm sure i'm not understanding it the wrong way , what you're suggesting is that we have something like `onError(...).` that the developer can use to define a callback function and then treat the _response code_ (not the exception) the way they want and if they want?

mkarg: I explicitly do *not* want anything that is called "...error..." for non-application-errors (and 407 is *not* an application error, it is just an error code *of HTTP*). JAX-RS is *not* an HTTP wrapper - the Java API for that is *Servlets* - but JAX-RS is a *REST* API. My proposal is to make up our minds what HTTP status codes are actually worth getting covered by JAX-RS as standalone exception classes, and which not. For 407 in particular, I would plea for a method to register a `onProxyLogin(lambda)` callback which provides the algorithm to fulfil in case of missing proxy credentials.

jeyvison: understood. As long as we open an issue to deal with this kind of situation, i agree that we can remove some exceptions.

This is a list of exceptions that i believe that doesn't need to me commited:

- ProxyAutheticationRequiredException
- NetworkAuthenticationRequiredException
  
**maybe**:
- ExpectationFailedException
- PreconditionRequiredException

Do you folks agree?

mkarg: @eclipse-ee4j/eclipse-jaxrs Opinions?

chkal: My personal preference:

Include:
  * ExpectationFailedException
  * NotImplementedException
  * PreconditionRequiredException
  * TooManyRequestsException

Maybe:
  * RequestedRangeNotSatisfiableException  (Should be RangeNotSatisfiableException has header was renamed in RFC)
  * RequestEntityTooLargeException
  * RequestHeaderFieldsTooLargeException
  * MethodNotAllowedException
  * LenghtRequiredException

Exclude
  * BadGatewayException
  * GatewayTimeoutException
  * ProxyAutheticationRequiredException
  * RequestUriTooLong
  * HttpVersionNotSupportedException
  * NetworkAuthenticationRequiredException
  * RequestTimeoutException

Reasoning: The exceptions in the excluded list are very technically and/or refer to proxies. The ones in the "maybe" could be useful in some situations, but only in very special cases.

jeyvison: So, i agree with @chkal 's list but i would remove the  maybe list too. We can always put it back if it's needed. What you think @mkarg ?

mkarg: Looking at the RFC's description for NotImplementedException I do wonder what use case one should use that for, as this is actually what the JAX-RS engines does automatically.

mkarg: The same for PreconditionRequired.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: @eclipse-ee4j/eclipse-jaxrs Opinions?

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
chkal: @mkarg I could imagine that developers would want to use `NotImplementedException` if a certain feature of the resource is not fully implemented yet and therefore throw it manually from the resource. However, I agree that it is a rare case.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
chkal: Well, `Throwable` implements `Serializable`. And the API docs recommends defining the `serialVersionUID` manually. I don't think that the exceptions will be changed in a binary incompatible way very often, so I'm fine with including the `serialVersionUID` and adapting it if required.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: As it is a *rare* case I do not see the benefit of having it in the standard (possibly good for some "official extensions lib").

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: I don't see a problem `serialVersionUID ` keeping them too. as @chkal said, we can adapt it later if we need to.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: Just removed `NotImplementedException `

mkarg:
@jeyvison Can you please tell us a little bit why you just replaced your contribution?

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: Dumb question, why not simply using `1L;`? I know IDEs create these fancy numbers, but what are these scary codes good for in the sense of "better than 1L"?

jeyvison:
> @jeyvison Can you please tell us a little bit why you just replaced your contribution?

Hi @mkarg . You mean the last push force I sent? it was just to remove `NotImplementedException` . Shouldn't I have done it this way?

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: According to the [serialization spec](https://docs.oracle.com/javase/6/docs/platform/serialization/spec/version.html#6519), this is to guard against classes with the same name being identified as versions of the same class.

mkarg:
@jeyvison The problem with force-push is that the history gets lost, so we do not see the difference between what was already reviewed, and what changed since then. So we have to start the review from scratch...

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
mkarg: @jeyvison I know what a SUID is good for. Actually I do not see why a SUID of 1L wouldn't suffice. It identifies the stream version as good as any other number.

pzygielo:
> so we do not see the difference between what was already reviewed, and what changed since then. So we have to start the review from scratch...

It's not lost completely: https://github.com/eclipse-ee4j/jaxrs-api/compare/6c50c98..a0c5c73

jeyvison:
Sorry @mkarg . I did it before in this same PR so I thought it was ok =\

mkarg:
@pzygielo Right. Let's say it this way: It makes reviews harder for us.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
jeyvison: Then I'm ok with just 1L. It's cleaner.

mkarg:
@jeyvison PRs are intended as a discussion stream. Just add a commit ontop and give the comment why you added that commit. This gives us a nice and easy to follow history. After the final voting, the committers will be able to squash together all commits into a single one. So there is no need to do that upfront. :-)

jeyvison:
@mkarg got it ;)

jaxrs-api/src/main/java/jakarta/ws/rs/LenghtRequiredException.java
```diff
@@ -0,0 +1,106 @@
...
+
+import jakarta.ws.rs.core.Response;
+
+/**
+ * A runtime exception indicating the server refuses to accept the request without a defined Content-Length.
+ *
+ * @author Jeyvison Nascimento (jeynoronha@gmail.com)
+ * @since 2.2
+ */
+public class LenghtRequiredException extends ClientErrorException {
```
ronsigal, line 22: lenght -> length throughout

jaxrs-api/src/main/java/jakarta/ws/rs/GatewayTimeoutException.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ ********************************************************************************/
+
+package jakarta.ws.rs;
+
+import jakarta.ws.rs.core.Response;
+
+/**
+ * A runtime exception indicating that the server, while acting as a gateway or proxy,
+ * received an invalid response from the upstream server it accessed in attempting to fulfill the request.
```
ronsigal, line 18: This description is identical to the one for BadGatewayException. Shouldn't it mention "timeout"?

jaxrs-api/src/main/java/jakarta/ws/rs/HttpVersionNotSupportedException.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @author Jeyvison Nascimento (jeynoronha@gmail.com)
+ * @since 2.2
+ */
+public class HttpVersionNotSupportedException extends ServerErrorException {
+
+    private static final long serialVersionUID = -2162737983746617138L;
+
+    /**
+     * Construct a new bad client request exception.
```
ronsigal, line 28: I'm not sure why this is described as "bad client request exception". It's very generic, whereas 505 is more specific. Same throughout the definition of HttpVersionNotSupportedException.

jaxrs-api/src/main/java/jakarta/ws/rs/RequestedRangeNotSatisfiableException.java
```diff
@@ -0,0 +1,105 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ ********************************************************************************/
+
+package jakarta.ws.rs;
+
+import jakarta.ws.rs.core.Response;
+
+/**
+ * A runtime exception indicating that the client sent a {@link jakarta.ws.rs.core.Response.Status#REQUESTED_RANGE_NOT_SATISFIABLE}
```
ronsigal, line 17: "client sent a ... by the server" doesn't make sense.

jaxrs-api/src/main/java/jakarta/ws/rs/BadGatewayException.java
```diff
@@ -0,0 +1,105 @@
...
+package jakarta.ws.rs;
+
+import jakarta.ws.rs.core.Response;
+
+/**
+ * A runtime exception indicating that the server, while acting as a gateway or proxy,
+ * received an invalid response from the upstream server it accessed in attempting to fulfill the request.
+ *
+ * @author Jeyvison Nascimento (jeynoronha@gmail.com)
+ * @since 2.2
```
ronsigal, line 21: Shouldn't this be 3.0 now?

Same for each class.

andymc12: I think it should be 3.1 - and the PR should target `3.1-SNAPSHOT` instead of `master`.

spericas: Yes, since 3.1 and please target 3.1-SNAPSHOT

mkarg:
@jeyvison There are open comments with this PR. Please address them ASAP so we could start the official voting for merging into JAX-RS 3.1. Thanks.

jaxrs-api/src/main/java/jakarta/ws/rs/LenghtRequiredException.java
jeyvison: Can you clarify here, @ronsigal ? Didn't understand the comment

jeyvison: No need, got it

jeyvison:
Done. See https://github.com/eclipse-ee4j/jaxrs-api/pull/918

jeyvison:
@mkarg Not sure if the Travis error is related to my changes.

spericas:
What is the status of this PR? Are we still targeting it for 3.1?

chkal:
I'm still not sure if we really need exception types for everything. Especially types like `RequestHeaderFieldsTooLargeException` or  `RequestedRangeNotSatisfiableException` won't be used in real world project very often. So IMO it would be fine if the developer would have to create such an exception type himself.

spericas:
I agree with @chkal. I am personally not a big fan of these super fine grained exception types, and hope we can get this new set to a minimum.

mkarg:
I am hereby closing this PR as it did not reach the needed quorum.

PR #779, opened 2019-07-08
Changes for Jakarta EE 8 Release

spericas:
Purpose of this PR is to eliminate the use of the JAX-RS acronym where appropriate. In most cases, the use of the acronym as an adjective was redundant and easily inferred from context. Still some ongoing discussions about the use of a new acronym/short name, but this PR does _not_ propose the use of one, instead refers to "this API" when necessary. 

See https://wiki.eclipse.org/How_to_Prepare_API_Projects_to_Jakarta_EE_8_Release

Since there are no real API changes, I propose to fast-track review period of just one day.

Signed-off-by: Santiago Pericas-Geertsen <santiago.pericasgeertsen@oracle.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,12 +1,12 @@
-# Contributing to Jakarta RESTful Web Services
+# Contributing to the Jakarta RESTful Web Services Project
 
 Thanks for your interest in this project.
 
 ## Project description
 
-JAX-RS: Java API for RESTful Web Services (JAX-RS) is a Java programming
-language API spec that provides support in creating web services according to
-the Representational State Transfer (REST) architectural pattern.
+Jakarta RESTful Web Services provides a specification document, TCK and foundational 
+API to develop web services following the Representational State Transfer (REST) 
+architectural pattern.
 
 * https://projects.eclipse.org/projects/ee4j.jaxrs
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing to Eclipse Project for JAX-RS
+# Contributing to Jakarta RESTful Web Services
 
 Thanks for your interest in this project.
 
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -15,12 +15,6 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>net.java</groupId>
-        <artifactId>jvnet-parent</artifactId>
-        <version>1</version>
-    </parent>
-
     <groupId>jakarta.ws.rs</groupId>
     <artifactId>jakarta.ws.rs-examples</artifactId>
     <version>2.1.6-SNAPSHOT</version>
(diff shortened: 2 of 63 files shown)


CONTRIBUTING.md
```diff
@@ -1,12 +1,12 @@
-# Contributing to Eclipse Project for JAX-RS
+# Contributing to the Jakarta RESTful Web Services Project
```
mkarg: I wonder why we not simply adopt #779 instead?

examples/pom.xml
```diff
@@ -66,14 +60,10 @@
...
         <developer>
-            <email>santiago.pericasgeertsen@oracle.com</email>
-            <id>Santiago</id>
-            <name>Santiago Pericas-Geertsen</name>
-            <organization>Oracle</organization>
-            <roles>
-                <role>JAX-RS Spec Lead</role>
-            </roles>
-            <timezone>EST</timezone>
+            <id>developers</id>
```
mkarg: According to https://maven.apache.org/pom.html#Developers this is wrong. The `developers` section shall only refer to people being responsible for the code, not *any* contributor (the POM has a separate secion for contributors).

CONTRIBUTING.md
spericas: This is #779. If covered by another PR, will drop it from here.

mkarg: You are right, I just noticed that a minute ago. Sorry for the confusion.

Can you please cherry-pick 66ce6b3b9ea61daab64153789bababc2c14a3d84 and rebase ontop instead of copying the content? That would be fair. :-)

examples/pom.xml
spericas: I thought we had this discussion and made a similar change to the API's pom already. In any case, either way is fine by me.

mkarg: You are right, it is the same in `master`.

jaxrs-api/src/main/java/javax/ws/rs/Priorities.java
```diff
@@ -17,19 +17,19 @@
 package javax.ws.rs;
 
 /**
- * A collection of built-in priority constants for the JAX-RS components that are supposed to be
+ * A collection of built-in priority constants for the API components that are supposed to be
```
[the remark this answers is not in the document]
spericas: I find "components" without any qualification too generic in certain contexts, so I think being more specific is a good thing. After all, JAX-RS _is_ the API and that _was_ the qualifier used before.

CONTRIBUTING.md
spericas: Does that mean you'll approve it if I do all that work? :-)

jaxrs-api/src/main/java/javax/ws/rs/Priorities.java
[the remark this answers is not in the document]
mkarg: I'd either keep JAX-RS or simply write "...constant for components that..."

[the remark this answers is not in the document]
spericas: Ok, I don't think it's better, but in the interest of making progress I will make the change.

CONTRIBUTING.md
mkarg: "all that work"? You simply need to `git rebase -i HEAD^` and add `p 66ce6b3` in the first line and the `git push -f` to have a fair commit history. Should be done in one minute.

As I said, I am not +1 for removing the word "JAX-RS", but I do not stand in the way of a majority. So my vote will be "0".

spericas: It was done, and I was joking.

mkarg: Thanks! :-) :tada:

mkarg: I am a bit disappointed that you sqashed the changes for the merge, after you agreed to explicitly keep my original commit standalone.

PR #783, opened 2019-07-12
Updated references to the Jakarta XML Binding API

spericas:
For consistency with other changes and according to (1.3) of the Jakarta release document.

On **fast track** like all other Jakarta EE 8 PR's.

Signed-off-by: Santiago Pericas-Geertsen <santiago.pericasgeertsen@oracle.com>

### DIFF
--- a/jaxrs-api/src/main/java/javax/ws/rs/core/Link.java
+++ b/jaxrs-api/src/main/java/javax/ws/rs/core/Link.java
@@ -422,7 +422,7 @@ public abstract class Link {
 
     /**
      * Value type for {@link javax.ws.rs.core.Link} that can be marshalled and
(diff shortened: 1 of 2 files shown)


spericas:
> JAXB is still part of Java SE 8, which is required by Jakarta EE 8.
> 
> Jakarta XML Binding is not part of Jakarta EE 8 and likely won't be approved at the same time.
> 
> These references to JAXB should remain for now. Jakarta EE 9 may include Jakarta XML Binding,
> at which point these references can be updated.

 Thanks for the input @bshannon. In that case, I will not merge this PR for now.

spericas:
As indicated by @bshannon, there is no need for this change at this time. We shall continue to use JAXB for this reference.

PR #791, opened 2019-08-08
Changes requested for Jakarta EE release

spericas:
These changes were requested during the release process. Needs fast tracking.

### DIFF
--- a/jaxrs-spec/pom.xml
+++ b/jaxrs-spec/pom.xml
@@ -22,7 +22,7 @@
 
     <groupId>jakarta.ws.rs</groupId>
     <artifactId>jakarta.ws.rs-spec</artifactId>
-    <version>2.1.6</version>
+    <version>2.1</version>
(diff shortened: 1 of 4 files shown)


jaxrs-spec/pom.xml
```diff
@@ -22,7 +22,7 @@
 
     <groupId>jakarta.ws.rs</groupId>
     <artifactId>jakarta.ws.rs-spec</artifactId>
-    <version>2.1.6</version>
+    <version>2.1</version>
```
mkarg: I do not see who has a benefit from hiding the fact that this is just a bug fix of an existing specification.

bshannon: It is not a bug fix of an existing specification.  It's the first version of this specification approved by Eclipse.

mkarg: Maybe we should write 2.1-1 then? In maven words this means "second build of same version".

PR #799, opened 2019-10-01
GH-770 Rename JAXRS interface

### LINKED ISSUE #770: Rename JAXRS interface
JAX-RS 2.2 branch contains [`JAXRS`](https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/jakarta/ws/rs/JAXRS.java) interface. Along with renaming JAX-RS, the interface name should be renamed. 

As it is responsible for bootstrapping a JAX-RS application in Java SE environment, a name reflecting that functionality should be chosen. Possibilities are:
- `Bootstrap` - simple and short, but does not reflect the fact it is for a standalone Java SE application bootstrap only
- `SeBootstrap` - Aligned with CDI `Se` prefixed [classes](https://docs.jboss.org/cdi/api/2.0/javax/enterprise/inject/se/SeContainer.html)
- `StandardBootstrap` - reflects usage in plain _Java Standard Edition_ environment
- `StandaloneBoostrap` - reflects the fact it does not need a support of a ServletContainer/ApplicationServer, but JAX-RS Spec. does not use the `Standalone` term (yet)
- or any other name

mkarg:
Renamed Java SE Bootstrap interface from `JAXRS` to `SeBootstrap` as proposed by @jansupol and objected by nobody (see discussion in #770).

Closes #770.

*According to our [Committer Conventions](https://github.com/eclipse-ee4j/jaxrs-api/wiki/Committer-Conventions) the review time for this PR is two full weeks.*

### DIFF
--- a/examples/src/main/java/jaxrs/examples/bootstrap/BasicJavaSeBootstrapExample.java
+++ b/examples/src/main/java/jaxrs/examples/bootstrap/BasicJavaSeBootstrapExample.java
@@ -18,8 +18,8 @@ package jaxrs.examples.bootstrap;
 
 import java.net.URI;
 
-import jakarta.ws.rs.JAXRS;
-import jakarta.ws.rs.JAXRS.Configuration;
+import jakarta.ws.rs.SeBootstrap;
+import jakarta.ws.rs.SeBootstrap.Configuration;
 import jakarta.ws.rs.core.Application;
 import jakarta.ws.rs.core.UriBuilder;
 
@@ -46,9 +46,9 @@ public final class BasicJavaSeBootstrapExample {
     public static final void main(final String[] args) throws InterruptedException {
         final Application application = new HelloWorld();
 
-        final JAXRS.Configuration requestedConfiguration = JAXRS.Configuration.builder().build();
+        final SeBootstrap.Configuration requestedConfiguration = SeBootstrap.Configuration.builder().build();
 
-        JAXRS.start(application, requestedConfiguration).thenAccept(instance -> {
+        SeBootstrap.start(application, requestedConfiguration).thenAccept(instance -> {
             Runtime.getRuntime()
                     .addShutdownHook(new Thread(() -> instance.stop()
                             .thenAccept(stopResult -> System.out.printf("Stop result: %s [Native stop result: %s].%n",
--- a/examples/src/main/java/jaxrs/examples/bootstrap/ClientAuthenticationJavaSeBootstrapExample.java
+++ b/examples/src/main/java/jaxrs/examples/bootstrap/ClientAuthenticationJavaSeBootstrapExample.java
@@ -18,9 +18,9 @@ package jaxrs.examples.bootstrap;
 
 import java.net.URI;
 
-import jakarta.ws.rs.JAXRS;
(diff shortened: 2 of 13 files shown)


jansupol:
Would it be possible to update the copyright year to 2019?

mkarg:
> Would it be possible to update the copyright year to 2019?

Sure I can do that. Is it really necessary? It breaks the approvals, so everbody has to re-approve then.

spericas:
@mkarg Did not check the EF rules recently, but don't you need to update the copyright year on updates, no just creation? If you need to make the change, I'll re-approve free of charge :)

chkal:
I'm also fine with doing the approval again. We are in the same TZ, so you should get it almost immediately.

mkarg:
> @mkarg Did not check the EF rules recently, but don't you need to update the copyright year on updates, no just creation? If you need to make the change, I'll re-approve free of charge :)

*I did* check the [EF rules](https://www.eclipse.org/projects/handbook/#ip-copyright-headers) once more, and I asked Wayne Beaton about that topic earlier this year, and the result is that the PR as-is is 100% compliant to EF rules: Copyright headers are actually OPTIONAL ("should", not "must"). We MUST NOT change the original copyright year. We MAY add an updated *end-of-range* *if we want that* (I do not want that).

Hence I will merge the PR as-is after the review period. :-)

jansupol:
I must say I do not agree with this copyright year interpretation. According to that, it would make no sense to use a range of years. 

While it is nice that we are not forced by the EF to update the copyright year, it is for the good of the project to do so. First, validity of copyright is extended for another year (at least that's what I understand the copyright is good for). Second, everyone who comes to the file can see that there was an activity and that the file is not a dead piece of code. Third, it would keep the customs the API had in previous years.

mkarg:
> I must say I do not agree with this copyright year interpretation. According to that, it would make no sense to use a range of years.
> 
> While it is nice that we are not forced by the EF to update the copyright year, it is for the good of the project to do so. First, validity of copyright is extended for another year (at least that's what I understand the copyright is good for). Second, everyone who comes to the file can see that there was an activity and that the file is not a dead piece of code. Third, it would keep the customs the API had in previous years.

This is just your personal opinion (which I respect but do not share) and neither a necessarity nor the EF default. This project now is an EF project, and as the constant updating of the copyright is a burden, I am strong -1 against your request.

jansupol:
@mkarg I agree, it's a burden, but the advantages overweight the few seconds of doing it. Do you want me to do it?

mkarg:
@jansupol Feel free to personally add a commit ontop anything we commit here in future if you like to have that, but note that I will never take care of copyright ranges (only care for EF rules) or even wait for that when reviewing / merging PRs. But actually I really *do* dislike the complete copyright updating idea, as I personally think it is a real pain in the ass and do not see any benefit at all. The copyright is valid without that proof, even in the US since few years. Activity is much better proven by Git log / Github history. The customs of the last year SHOULD be broken by intention as not everybody was pleased by them. In fact, while reviews I hate see the diff being cluttered with senselessly updated copyright ranges, and the Github stats shows a wrong picture of the number of modified lines due to that.

Issue #804, opened 2019-10-23
Spec: Todo


mkarg:
This issue serves as an aggregated placeholder for the open TODOs with the Spec document, so we can organize to work in parallel without solving problems twice. The following list shows the open issues. If you like to provide a solution for one of these issues, please post a comment below.

- [x] (@chkal: #801) Initial contribution of the spec asciidoc "as-is" to `master`
- [x] (@mkarg: #801) URL fixes
- [x] (@mkarg: #801) Maven Coordinates
- [x] (@mkarg: #806) Java EE -> Jakarta EE
- [x] (@mkarg: #802) Can also build spec on JDK 9 and later now
- [x] (@mkarg: #802) Travis CI/CD builds spec
- [x] (@mkarg: #806) 2.x -> 2.2-SNAPSHOT
- [x] (@mkarg: #808) POM Cleanup. Maybe drop assembly?
- [x] (@mkarg: #806) Changes mentioned in [CQ 20980](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20980): *During our initial review of this submission, we noted that the source code headers are either incorrect or missing. As this is a housekeeping issue only, you may proceed with checkin, however we would ask that the correct copyright and license headers be added to the files per Eclipse standards. Please ensure that this is done and notify us via the CQ once this is completed.* -- @ivargrimstad triggered @waynebeaton to provide the wanted copyright header template.
- [x] (@mkarg: #808) Publish nightly build of snapshot PDFs/HTMLs *Waiting for PMC answer regarding target location*
- [x] (@mcserra: #806, @chkal #883) Modifying references to the JCP with corresponding Eclipse processes.
  - [x] #806 - Everything excluding Activation and JAX-WS
  - [x] #883 - JAX-WS
- [x] (@thadumi: #848) What do we do with references to JSRs? [EJB] Use spec. Name and version number of the specification instead.
- [ ] (@edwin-amoakwa: TBA) References to existing Java EE technologies, acronyms, and JSRs are okay as long as they are properly attributed.
- [x] Be careful not to modify the semantics or behavior — we need Jakarta EE 8 to be functionally equivalent to Java EE 8.
- [x] (@mkarg: #806) Use the original Java EE specification as a guide for your expected end result. Look at the TOC, the figures, the tables, the indentations, etc. The Jakarta EE specification should look very similar to the Java EE counterpart.
[...]

PR #807, opened 2019-10-28
[#804] Spec: Split Chapters into single files

mkarg:
This PR splits the spec into one file per chapter.


chkal:
Looks great! I compared the original PDF and the new one, and they are exactly the same.

However, I think we should have one sub directory for each chapter. This way it is much simple to find the corresponding files quickly. I created mkarg/jaxrs-api#5 for this.

mkarg:
> I think we should have one sub directory for each chapter. This way it is much simple to find the corresponding files quickly. I created [mkarg#5](https://github.com/mkarg/jaxrs-api/pull/5) for this.

@chkal Good idea! This PR now includes your proposal (d33a7136bf3626c38b15b033d377ddc2a6e0d364).

@eclipse-ee4j/eclipse-jaxrs Review please! :-)

chkal:
Comparison of both documents:

https://draftable.com/compare/tsjpPCJKqqPW

Issue #554, opened 2017-04-06
Clarify documentation ambiguities


glassfishrobot:
A number of Resteasy JIRA issues turn out to arise from ambiguities in the JAX-RS specification.

1\. From [https://issues.jboss.org/browse/RESTEASY-1162](https://issues.jboss.org/browse/RESTEASY-1162) "Response.created doesn't work as advertised with relative URIs":

The javadoc for javax.ws.rs.core.Response.created() says, "If a relative URI is supplied it will be converted into an absolute URI by resolving it relative to the request URI" However, the implementation of Response.created() is

```
public static ResponseBuilder created(URI location) {
        return status(Status.CREATED).location(location);
    }
```

and the javadoc for javax.ws.rs.core.Response$ResponseBuilder.location() says, "If a relative URI is supplied it will be converted into an absolute URI by resolving it relative to the base URI of the application".

2\. From [https://issues.jboss.org/browse/RESTEASY-1505](https://issues.jboss.org/browse/RESTEASY-1505) "@DefaultValue on @PathParameter not working":

The javadoc for @DefaultValue says, "Defines the default value of request meta-data that is bound using one of the following annotations: javax.ws.rs.PathParam, javax.ws.rs.QueryParam, javax.ws.rs.MatrixParam, javax.ws.rs.CookieParam, javax.ws.rs.FormParam, or javax.ws.rs.HeaderParam. "

On the other hand, Appendix A of the specification "JAX-RS: Java API for RESTful Web Services", Version 2.0, says that @DefaultValue "Specifies a default value for a field, property or method parameter annotated with @QueryParam , @MatrixParam , @CookieParam , @FormParam or @HeaderParam." Note that the spec omits @PathParam.

I think that the problem is the use of something like @Path("/a/b/{c}"). If the value for "c" is missing, then the annotated resource method just isn't going to get called. But what about @Path(/a/b/c{d}e)?

3\. From [https://issues.jboss.org/browse/RESTEASY-1516](https://issues.jboss.org/browse/RESTEASY-1516) "Cookies sent by resteasy-client are not spec compliant":

Not exactly an ambiguity. I'll just refer to #435.
#### Affected Versions
[2.0]

ronsigal:
For the first problem, Response.created(), I would suggest turning the javadoc for created() to
```
    /**
     * Create a new ResponseBuilder for a created resource, set the location
     * header using the supplied value.
     *
     * @param location the URI of the new resource. If a relative URI is
[...]

PR #810, opened 2019-11-10
[#554] Fix javadoc for Response.created()

ronsigal:
Closes item 1 of issue #554.

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
@@ -612,7 +612,7 @@ public abstract class Response implements AutoCloseable {
      * Create a new ResponseBuilder for a created resource, set the location header using the supplied value.
      *
      * @param location the URI of the new resource. If a relative URI is supplied it will be converted into an absolute URI
-     * by resolving it relative to the request URI (see {@link UriInfo#getRequestUri}).
+     * by resolving it relative to the base URI (see {@link UriInfo#getRequestUri}).
      * @return a new response builder.
      * @throws java.lang.IllegalArgumentException if location is {@code null}.
      */


mkarg:
@ronsigal You don't need an sign-off for the PR itself, just for the contained commit(s).

chkal:
I'm basically +1 for this change. However, I'm not sure for which release we should target it. As @spericas already mentioned in #554, this is most likely a spec bug. However, the fix is an incompatible change.

ronsigal:
Hi @chkal,

Actually, there's an argument to be made that it's not really a substantive change. The implementation of Response.created() is
```
    public static ResponseBuilder created(URI location) {
        return status(Status.CREATED).location(location);
    }
```
and the javadoc for ResponseBuilder.location() says
```
        /**
         * Set the location.
         *
         * @param location the location. If a relative URI is supplied it will be
         *                 converted into an absolute URI by resolving it relative to the
         *                 base URI of the application (see {@link UriInfo#getBaseUri}).
         *                 If {@code null} any existing value for location will be removed.
         * @return the updated response builder.
         */
```
so if ResponseBuilder.location() is implemented correctly, then Response.created() already adheres to the updated javadoc. 

-Ron

ronsigal:
For the record, I also submitted a test for the TCK: https://github.com/eclipse-ee4j/jakartaee-tck/pull/135.

mkarg:
oops... I hate the "Close and commit" button... ;-)

mkarg:
@ronsigal Let's see what other vendors think. I just invited them to share their opinion.

mkarg:
I'm convinced that @ronsigal is right and it is simply a typo in the spec. So if we fix the spec as proposed (i. e. keep the implementation), then the actual behaviour would be backwords compatible. As a result, we could add it to JAX-RS 2.2.

ronsigal:
Thanks, everyone.

mkarg:
Do we need to mention this in the spec, at least in the history appendix?

chkal:
IMO adding a note to the history appendix is sufficient.

mkarg:
> IMO adding a note to the history appendix is sufficient.

I drafted https://github.com/eclipse-ee4j/jaxrs-api/pull/936. WDYT?

PR #820, opened 2019-12-17
Migration of Jenkins files (Fixed Nightly Build)

senivam:
Fixed: Nightly Build fails after after migration to ci.eclipse.org

Solution: Not using configuration files for Jenkins jobs

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,19 +19,13 @@ pipeline {
 				}
 			}
 			steps {
-				configFileProvider([
-					configFile(fileId: 'a79c37a7-bfd9-4699-8077-f992027b2c1b', targetLocation: '/home/jenkins/.m2/settings.xml'),
-					configFile(fileId: 'b78b4e15-215f-42bc-81ac-53cd3e2ef708', targetLocation: '/home/jenkins/.m2/')
-				]) {
-					dir ('jaxrs-api') {
-						sh "$MVN deploy"
-					}
-					dir ('examples') {
-						sh "$MVN deploy"
-					}
+				dir ('jaxrs-api') {
+					sh "$MVN deploy"
+				}
+				dir ('examples') {
+					sh "$MVN deploy"
 				}
 			}
 		}
 	}
 }
-


mkarg:
How will GPG find the keyring then?

senivam:
Here [migration.bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=551863#c7) I was answering to Santiago's question - configuration files are still there, so basically script works the same way. The only difference is that there is no need to provide them externally for each job. They are already inside JIPP's maven configuration.

mkarg:
> Here [migration.bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=551863#c7) I was answering to Santiago's question - configuration files are still there, so basically script works the same way. The only difference is that there is no need to provide them externally for each job. They are already inside JIPP's maven configuration.

But Frederic answered in that Migration Bug Report that it is not enough to simply remove the settings.xml:
>withCredentials([file(credentialsId: 'secret-subkeys.asc', variable:
'KEYRING')]) {
  sh 'gpg --batch --import "${KEYRING}"'
  sh 'for fpr in $(gpg --list-keys --with-colons  | awk -F: \'/fpr:/ {print $10}\' | sort -u); do echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key ${fpr} trust; done'
}
"--pinentry-mode loopback" needs to be added to the pom.xml as well.

Jenkinsfile
```diff
@@ -19,19 +19,13 @@ pipeline {
 				}
 			}
 			steps {
-				configFileProvider([
-					configFile(fileId: 'a79c37a7-bfd9-4699-8077-f992027b2c1b', targetLocation: '/home/jenkins/.m2/settings.xml'),
-					configFile(fileId: 'b78b4e15-215f-42bc-81ac-53cd3e2ef708', targetLocation: '/home/jenkins/.m2/')
```
mkarg: Don't we need the new GPG keyring import described by Frederic?

senivam:
Actually Frederic answered, that IF it is required to deploy to OSSRH/Maven Central, then described GPG keyring shall be added. 

The purpose of this PR was to fix current functionality (which is required by migration from old JIPP to new JIPP). 

Adding GPG config would provide new functionality to the build. This is not migration issue. Build will work without GPG, keeping already existed functionality.

If it is really required to add GPG it shall be another PR not related to migration.

Jenkinsfile
mkarg: > Adding GPG config would provide new functionality to the build. This is not migration issue. Build will work without GPG, keeping already existed functionality.

@senivam If you check the `pom.xml` file https://github.com/eclipse-ee4j/jaxrs-api/blob/61a230d5924e00714fb0799a4c16175a74c0dd29/jaxrs-api/pom.xml#L125 you will see that `maven-gpg-plugin` was used *already before the migration* to `sign` the JARs. So this is broken *by the migration*, hence should get fixed *by the same*  "fix-the-miration-PR".

senivam: yes, maven-gpg-plugin was used before, it is correct. However those to lines which are currently breaking the build are substitution for settings.xml and settings-security.xml which were provided inside job's configuration in previous JIPP. The current JIPP has mounted those files (settings.xml and settings-security.xml) just before the build (you can check it in log of failed builds after migration). So, there is no need to provide them inside job's configuration. Further more, there is no references to those configuration files in the actual configuration. That is why job keeps failing (this is visible in log output as well). There is no mention about GPG sign failure or sign padding failure (which would be sign of wrong GPG functioning). So, when those two lines will be removed from job's configuration, job will work.

senivam: btw in pom.xml gpg plugin is being activated when used with -Prelease profile. The job we are talking about performs nightly builds only so no profile release (thus no option -Prelease) is used when running the job.

mkarg: I do not agree that we are talking only about one particular job (nightly builds), as the word "nightly" was *first* mentioned by your comment directly above. Your PR is called "Migration of Jenkins files" (which includes the GPG keyring), and your initial comment says: "It's not required to use configuration files for Jenkins jobs after migration to ci.eclipse.org". If you like a second PR for the GPG fix, that's fine for me, but we should clearly rephrase *this* PR's title and description so reviewers do know that all you intend is to fix solely the *nightly* build.

senivam: where else is this Jenkins file used?

senivam: which contains stage:
`stage('Nightly Build') {`

mkarg: It needed some time to think about your last two comments, but now I see what you mean. GPG is not used by the *pipeline jobs* (hence `Jenkinsfile`) plays no role, while the jobs using GPG already perform the keyring import using their particular *static* configuration in Jenkins.

mkarg:
@senivam Thanks for fixing this! :+1:

senivam:
you are welcome :) If it is required to fix this for EE4J_8 branch as well, same PR shall be submitted there. However I'm not sure if EE4J_8 branch is still maintained...

mkarg:
> you are welcome :) If it is required to fix this for EE4J_8 branch as well, same PR shall be submitted there. However I'm not sure if EE4J_8 branch is still maintained...

That is a good question! I will ask the PMC about that, as it is "their" branch.

PR #828, opened 2019-12-31
Added API documentation for 2.1.6 to project page

chkal:
Currently, the "APIs" link on our [site](https://eclipse-ee4j.github.io/jaxrs-api/) results in a 404 error.

This PR adds a page for the API documentation (see `index.md`) and the API docs for the 2.1.6 release (the content of `jakarta.ws.rs-api-2.1.6-javadoc.jar` fetched from Maven Central).

You can see a preview of the page here: https://chkal.github.io/jaxrs-api/apidocs/

(Please note that our site uses absolute URLs and therefore clicking around in the preview will take you to the real site)

**I propose a fast-track review period of just one day as per our committer rules.**


mkarg:
Good catch, seems I missed that! This effectively leads to a *local JAR download*, which is better than nothing, but maybe it would be *even better* to simply fix the *broken* link to point to https://jakarta.ee/specifications/restful-ws/2.1/apidocs/overview-summary.html?

mkarg:
@chkal Seems there is something wrong: Github tells me this PR changes 331 files?!

chkal:
> This effectively leads to a _local JAR download_

Nope. It's the **content** of the JAR file. So it is unpacked and therefore viewable online via:

https://chkal.github.io/jaxrs-api/apidocs/2.1.6/

> but maybe it would be _even better_ to simply fix the _broken_ link to point to

To be honest, I wasn't aware that our API docs are available there. So yes, we could also link to this page.

On the other hand: I also see some benefits of having our API docs in our own `gh-pages` branch. Especially if we plan to publish API docs of our nightly builds.

> Github tells me this PR changes 331 files?!

Correct, because the PR contains the full javadocs for 2.1.6 as mentioned above.

mkarg:
Having a copy in `gh-pages` is a good idea, and indeed I am already working on a solution to have it *auto*-committed (just waiting for  [this Maven feature](https://github.com/apache/maven-scm-publish-plugin/commit/af43cb4bb9493fccb17a77d3da1a89240d202f3f) to be publicly available).

mkarg:
@chkal :birthday: As a birthday present, your PR is approved to get merges. :man_dancing:

PR #833, opened 2020-01-15
[#606] Clarify when to call onError vs onComplete

andymc12:
Should finish resolving issue #606.  This clarifies that `onError` and `onComplete` callbacks should be mutually exclusive.

As this is only clarifying behavior in javadoc, I'd like to fast-track this PR.

It should go into the 3.0 release (master) and the 3.1 release (3.1-SNAPSHOT).  If approved, I'll try to cherry pick this into master.  If that is locked down, then I'll open another PR for master.

@spericas, @asoldano, can you review and approve these changes?
@NicoNes, @reta, I cannot add you as reviewers, but would you also take a look to confirm your approval of these changes?

Thanks, Andy

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/sse/SseEventSource.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/sse/SseEventSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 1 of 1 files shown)


NicoNes:
Hi guys,

The changes looks good to me as well. 

However because of following javadoc:

    /**
     * Register {@link InboundSseEvent} and {@link Throwable} consumers and onComplete callback.
     * <p>
     * Event consumer is invoked once per each received event, {@code Throwable} consumer is invoked invoked upon a
     * unrecoverable error encountered by a {@link SseEventSource}, onComplete callback is invoked when there are no further
     * events to be received.
     *
     * @param onEvent event consumer.
     * @param onError error consumer.
     * @param onComplete onComplete handler.
     * @throws IllegalArgumentException when the any of the provided parameters is {@code null}.
     */
    void register(Consumer<InboundSseEvent> onEvent,
            Consumer<Throwable> onError,
            Runnable onComplete);

 it is not that clear that both method can't be invoked together. Actually, developpers can still believe that when uncoverable error occured `onError` will be invoked and since no more event will be received, then `onComplete` will be invoked too.
So to make it clear once and for all, what do you think of stating clearly in the doc that those methods are mutualy exclusive ?

andymc12:
@NicoNes agreed.  I'll update that too.  How about if I change this line:
`onComplete callback is invoked when there are no further events to be received`

to:

`onComplete callback is invoked after a successful connection and when there are no further events to be received. Note that the onComplete callback will not be invoked if the onError callback has been invoked.`

NicoNes:
@andymc12 clearer like that. Thanks !

andymc12:
@asoldano Can you review this PR?  We have three reviews, but since two are not officially JAX-RS committers, we need one more approval from a committer before I can merge.  Thanks!

Issue #835, opened 2020-01-23
Clarification: Can `Application.getSingletons()` return a CDI proxy instance?


andymc12:
In the last couple of weeks, we've had a few different users open issues as a result of some missing clarification in the spec around whether the objects returned from `getSingletons()` can be CDI instances or not.

For example:
```
public class MyApp extends Application {

    @Inject
    MyResource myResource;

    @Inject
    MyProvider myProvider;

    @Override
    public Set<Object> getSingletons() {
        return Stream.of(myResource, myProvider)
                      .collect(Collectors.toCollection(HashSet::new));
    }
}
```

or:
```
public class MyApp extends Application {

    @Inject
    MyResource myResource;

    @Inject
    MyProvider myProvider;

    @Override
    public Set<Object> getSingletons() {
        MyResource myResource = CDI.current().select(MyResource.class).get();
        MyProvider myProvider = CDI.current().select(MyProvider.class).get()
        return Stream.of(myResource, myProvider)
                      .collect(Collectors.toCollection(HashSet::new));
    }
}
```

From reading the spec today (assuming that `MyProvider` is application scoped - see section 11.2.3), it seems like this should be legal, but in practice it runs into problems with section 11.2.8, which says:
> * Field and property injection of JAX-RS resources MUST be performed prior to the container invoking any `@PostConstruct` annotated method.
> * Support for constructor injection of JAX-RS resources is OPTIONAL. Portable applications MUST instead use fields or bean properties in conjunction with a `@PostConstruct` annotated method. Implementations SHOULD warn users about use of non-portable constructor injection.
> * Implementations MUST NOT require use of `@Inject` or `@Resource` to trigger injection of JAX-RS annotated fields or properties. Implementations MAY support such usage but SHOULD warn users about non-portability.

The JAX-RS implementation can invoked the `getSingletons` method, but cannot guarantee that it will perform `@Context` injection of JAX-RS resources before the `@PostConstruct` method is invoked.  For example, if the scope (of the resource) is `@Dependent`, then the CDI bean is initialized when constructed, before passing the singletons set back to JAX-RS.  Other scopes don't initialize the bean until a method has been called on it - but applications that log the bean would then end up calling the `toString()` method which initializes it and invokes the `@PostConstruct` method - prior to JAX-RS injection.

[...]

PR #836, opened 2020-01-25
[#835] Clarify injection wrt `getSingletons`

andymc12:
This partially addresses issue #835, updating the spec to clarify that the CDI integration requirements do not apply to resources or providers returned from the `getSingletons` method.

This should also be pushed to 3.1-SNAPSHOT branch.

PR #840, opened 2020-01-28
[#835] Deprecate getSingletons

andymc12:
Resolves part 2 (of 2) of #835 - this deprecates the `getSingletons` method of `Application`.

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/Application.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/Application.java
@@ -75,7 +75,10 @@ public class Application {
      *
      * @return a set of root resource and provider instances. Returning {@code null} is equivalent to returning an empty
      * set.
+     * @deprecated Automatic discovery of resources and providers or the {@code getClasses} method is preferred over
+      * {@code getSingletons}.
      */
+    @Deprecated
     public Set<Object> getSingletons() {
         return Collections.emptySet();
     }


andymc12:
@mkarg 
> I thought actually we would advise using CDI instead, but you advise not using singletons?

I advise not using `getSingletons` as the registration mechanism for resources and providers.  The deprecation comment in the javadoc  is intended to tell users how they should replace code currently using that method.  Instead of using `getSingletons` they should return a class instance (using `getClasses`) or rely on automatic discovery (i.e. return an empty list from `getClasses` / `getSingletons` and rely on the implementation detecting `@Path` and `@Provider` annotations.  If CDI is available, then those resources and providers MUST (still) be managed by CDI.  

Since 3.1 does not yet _require_ CDI management, I tried to make the text somewhat neutral as to the lifecycle management.  That way the rest of the spec can provide details about when CDI is used or not.  Do you suggest different text in the javadoc?

mkarg:
I understand all you wrote, the JavaDoc change is clear. But how can someone declare a singleton then (without CDI)?

andymc12:
For providers, they would default to singletons.  For resources, you could still use the deprecated method, you could use `@Singleton` from the EJB spec, you could use sub-resource locators that return the same instance, etc.

I want to get to a CDI-only world too (wrt injection and lifecycle management), and I think deprecating this API is one step in that direction.  I wasn't planning on removing EJB support, mandating CDI usage, etc. in this PR.  I think that should come in a separate issue/PR.

spericas:
I have a sense most developers would understand the rationale for this deprecation. Not sure it needs more than what @andymc12 already provided.

mkarg:
> For providers, they would default to singletons. For resources, you could still use the deprecated method, you could use `@Singleton` from the EJB spec, you could use sub-resource locators that return the same instance, etc.
> 
> I want to get to a CDI-only world too (wrt injection and lifecycle management), and I think deprecating this API is one step in that direction. I wasn't planning on removing EJB support, mandating CDI usage, etc. in this PR. I think that should come in a separate issue/PR.

+1 as long as `@Singleton` is still supported.

jansupol:
@andymc12 in a pure Java SE environment, would the user have an option not to use a deprecated method that would make them mad for firing a warning?

You mention `javax.ejb.Singleton`, but does the spec say it should work? The Spec, mentions `Singleton` in few places. One is in Section 8.3, which describes EJB session beans (i.e. it works in the EJB container, which is not very commonly used these days), and another one is in Section 9.4, which I understand is `javax.inject.Singleton`. For the ambiguity, I always felt the customers preferred `getSingletons` over `@Singleton`. Maybe documentation of RestEasy is more descriptive, but in case that our understanding is different, maybe it should be cleared in the Spec.

Anyway, while I feel the javadoc mentioning the deprecation is ok (maybe with javadoc making more clear how to replace the `getSingleton` methods), is it good to use `@Deprecated`, which fires the warnings?

andymc12:
@jansupol 
> in a pure Java SE environment, would the user have an option not to use a deprecated method that would make them mad for firing a warning?

Yes and no.  Both CDI and EJBs can run in Java SE.  The CDI SE approach should work by design - I'm less sure about the embeddable EJB container.  Without those additional APIs, then no, I don't think there is a way for a root level resource to be marked as singleton/app-scoped in Java SE.

> You mention javax.ejb.Singleton, but does the spec say it should work?

In section 11.2.4, it says "In a product that supports EJBs, an implementation MUST support the use of stateless and singleton session beans as root resource classes, providers and Application subclasses."  It doesn't say that it must treat `@Singleton` as a per-app resource, but my understanding is that the EJB Container would then manage the lifecycle, and thus it would be per-app/singleton.  That is how we have it implemented in Liberty.

> The Spec, mentions Singleton in few places. One is in Section 8.3, which describes EJB session beans (i.e. it works in the EJB container, which is not very commonly used these days), and another one is in Section 9.4, which I understand is javax.inject.Singleton.

Section 8.3 is definitely specific to `javax.ejb.Singleton`, but section 9.4 just says `@Singleton` (no package).  That should also be clarified - but I agree with you that it ought to be `javax.injection.Singleton`.  Maybe this could be clarified in PR #836 ?

> For the ambiguity, I always felt the customers preferred getSingletons over @Singleton. Maybe documentation of RestEasy is more descriptive, but in case that our understanding is different, maybe it should be cleared in the Spec.

I can't say which of our customers prefer one approach over the other, but I suspect that part of the reason for that is that customers using `@Singleton` (either ejb or inject) just work.  We end up seeing more support cases where customers are doing weird things like returning pre-initialized CDI proxies in the `getSingletons` method - and this is why I'm in favor of deprecating and removing that method. :-)

But I agree with you that we should try to clarify it in the spec.  That might also go bettering in #836.  Wdyt?

> Anyway, while I feel the javadoc mentioning the deprecation is ok (maybe with javadoc making more clear how to replace the getSingleton methods), is it good to use @Deprecated, which fires the warnings?

Personally, I prefer the compiler warning, but I'm open to change if others prefer that we remove it.

Thanks for the feedback!

spericas:
@andymc12 Any reason not to merge this one?

andymc12:
@spericas Sorry, lost track of time.  By chance, do you know if there is a way to get notified when a PR has been open for two weeks?

I'll go ahead and merge now.  Thanks for the reminder.

PR #851, opened 2020-03-03
Added parent pom: org.eclipse.ee4j:project:1.0.6

Tomas-Kraus:
Removed POM parts already defined in parent:

1. release profile is redundant, parent defines oss-release profile with everything set properly to use gpg and jakarta.oss.sonatype.org
2.  repositories section for jakarta.oss.sonatype.org is defined in parent for both snapshots and staging builds
3. maven-deploy-plugin and maven-release-plugin are set properly in parent
4. distributionManagement is also defined in parent pom

Release is done in 2 steps:
1. mvn deploy -Poss-release,staging
2. release artifact from staging repository to Maven Central when all tests passed and all release related paperwork is done too
- parent pom contains all required setup

I can modify all release cycle related jobs to work properly with parent POM setup.

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -20,6 +20,12 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
+    <parent>
+        <groupId>org.eclipse.ee4j</groupId>
+        <artifactId>project</artifactId>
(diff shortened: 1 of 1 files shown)


arjantijms:
@Tomas-Kraus Thanks!

Tomas-Kraus:
@mkarg I know that you would like to be independent. But have 150 lines shorter pom.xml is not that bad. Also parent pom configuration was already tested on rest of the EE4J projects and we know that it works.
If something changes in EF infra/OSSRH again, we'll fix it just in a single file for whole EE4J space.
So yes, you can for sure live without it, but on the other hand it will make your life a bit more simple.

chkal:
@Tomas-Kraus You self-assigned this issue and we have enough +1 votes. Any reason you didn't merge it yet?

arjantijms:
@chkal with so many approvals and time having passed by, it can surely be merged. I'll press that scary button then ;)

PR #854, opened 2020-03-10
[#815] Indicate that JAXB is optional

### LINKED ISSUE #815: Make dependency on JAXB optional
There is a [very active thread on the Platform Dev mailing list](https://www.eclipse.org/lists/jakartaee-platform-dev/msg01095.html) where we are discussing whether to Add the JAXB Spec/API to Jakarta.  The latest reply from Markus to me indicated a possibility of marking the dependency on JAXB from JAX-RS as optional, but he would like an issue for further discussion with the JAX-RS development team.  Thus, this issue...

Just for some quick background...  JAXB moved from Java EE to Java SE several years ago.  The Java EE specifications still maintained a dependency on JAXB, but it was clearly now delivered by Java SE.  In Java SE 11, the JDK community decided to remove JAXB from their deliverable.  So, the question on the table is whether to just continue to have this dependency on JAXB in it's current state and the javax namespace, or whether there is a need to evolve this API which would then require a move to the jakarta namespace.

The vote is kind of split right now.  There is the one side that recognizes the dependency between JAX-RS and JAXB and, thus, want to move JAXB to Jakarta.  But, there's the other side that recognizes that JAXB probably does not need evolving or updating, so why move it to Jakarta and incur more work?

Anyway, that was more for the background story.  The true intent of this issue is to have the discussion as to whether JAXB could be made an optional dependency for JAX-RS.

andymc12:
Modifications to the spec doc and javadoc that indicate that JAXB is optional.  Where possible, I tried to keep the language consistent with JSON-P / JSON-B - i.e. if the product supports JSON-B, then it must have a built-in entity provider to handle JSON requests - likewise for JAXB / XML.  

This change is targeted for the `master` branch (3.0), and should also go into the `3.1-SNAPSHOT` branch.

Fixes #815

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
@@ -1,6 +1,6 @@
 ////
 *******************************************************************
-* Copyright (c) 2019 Eclipse Foundation
+* Copyright (c) 2019, 2020 Eclipse Foundation
 *
 * This specification document is made available under the terms
 * of the Eclipse Foundation Specification License v1.0, which is
@@ -189,6 +189,16 @@ matching `/json` or `/*+json`.
 Note that if JSON-B and JSON-P are both supported in the same
 environment, entity providers for JSON-B take precedence over those for
 JSON-P for all types except `JsonValue` and its sub-types.
+
+[[jaxb]]
+==== Jakarta XML Binding
+
+In a product that supports the Jakarta XML Binding API (JAXB)
+<<bib21>>, implementations MUST support entity providers for the type,
+`jakarta.xml.bind.JAXBElement`, and application-supplied JAXB classes
+in combination with the following XML-based media types:
+`text/xml` and `application/xml` and media types of the
+form `application/*+xml`.
 
 [[additional_reqs]]
 ==== Additional Requirements
--- a/jaxrs-spec/src/main/asciidoc/chapters/providers/_entity_providers.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/providers/_entity_providers.adoc
@@ -1,6 +1,6 @@
 ////
 *******************************************************************
-* Copyright (c) 2019 Eclipse Foundation
+* Copyright (c) 2019, 2020 Eclipse Foundation
 *
 * This specification document is made available under the terms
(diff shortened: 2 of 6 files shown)


jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
```diff
@@ -190,6 +190,16 @@ Note that if JSON-B and JSON-P are both supported in the same
...
 
+[[jaxb]]
+==== Jakarta XML Binding
+
+In a product that supports the Jakarta XML Binding API (JAXB)
+<<bib21>>, implementations MUST support entity providers for the type,
+`jakarta.xml.bind.JAXBElement`, and application-supplied JAXB classes
+in combination with the following XML-based media types:
+`text/xml` and `application/xml` and media types of the
+form `application/*+xml`.
```
kwsutter: Is this syntax correct?  Or, should it just be `application/+xml`?

jaxrs-spec/src/main/asciidoc/chapters/providers/_entity_providers.adoc
```diff
@@ -155,8 +152,8 @@ type combinations:
   boxing/unboxing conversion.
 
 Depending on the environment, the list of standard entity providers MUST
-also include those for JSON. For more information about these providers
-see <<jsonp>> and <<jsonb>>.
+also include those for JSON or XML. For more information about these providers
+see <<jsonp>>, <<jsonb>> and <<jaxb>>.
```
kwsutter: I like the oxford comma after `<<jsonb>>`, but maybe that's just me...  :-)

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
andymc12: It's a wildcard - it's for content types like `application/svg+xml` or `application/xslt+xml`.

andymc12:
@mkarg - you're right.  no fast track on this one.  Thanks for pointing it out.  You have requested changes, but I didn't see any comments indicating what you'd like to see changed.  

@kwsutter I don't have a strong opinion about the extra comma - I've seen it both ways (with and without a comma before the last object).  If you're ok with it, I'll leave it as-is unless there are other changes that need to go in. 

Thanks!

mkarg:
@andymc12 Why not merging this one?

PR #857, opened 2020-03-31
Mandate default ExceptionMapper

ronsigal:


andymc12:
I think the wording is fine, but I think that we might not be able to put this into the `master` branch (3.0) due to the Jakarta EE 9 restrictions of not changing behavior from EE 8 (other than the package name).  IMO, this should definitely go in the `3.1-SNAPSHOT` branch.  @ronsigal can you re-target this PR?

spericas:
@ronsigal I agree with @andymc12, we need to target this for 3.1-SNAPSHOT at this time.

ronsigal:
Replaced by https://github.com/eclipse-ee4j/jaxrs-api/pull/858

PR #876, opened 2020-06-07
Disable PDF version of the spec document

chkal:
Unfortunately the PDF version of the spec is currently broken, because `asciidoctorj-pdf` doesn't render math expressions correctly. See for example:

pdf

In #865 we fixed the HTML version, so the HTML output looks fine:

html

According to my research, this issue cannot be solved easily. Therefore, I think we should disable the PDF output for now. IMO it would be fine to just release the HTML version. Maybe we can fix this issue later on, but for the moment disabling the PDF seems to be the best option.

### DIFF
--- a/jaxrs-spec/pom.xml
+++ b/jaxrs-spec/pom.xml
@@ -106,6 +106,8 @@
                             </attributes>
                         </configuration>
                     </execution>
+                    <!-- Currently broken, because asciidoctorj-pdf doesn't support the 'stem' attribute for rendering math expressions -->
+                    <!--
                     <execution>
                         <id>asciidoc-to-pdf</id>
                         <phase>generate-resources</phase>
@@ -135,6 +137,7 @@
                             </attributes>
                         </configuration>
                     </execution>
+                    -->
                 </executions>
             </plugin>
         </plugins>


arjantijms:
If I’m not mistaken @mojavelinux is working on a renderer that creates the PDF from the HTML version, which may fix this.

mojavelinux:
Support for LaTeX expressions when converting to Asciidoctor PDF is provided by asciidoctor-mathematical (https://github.com/asciidoctor/asciidoctor-pdf#asciidoctor-mathematical). It can be a challenging gem to install because of its native dependencies, though it is provided by the Asciidoctor Docker container (https://github.com/asciidoctor/docker-asciidoctor).

Guillaume is the one working on the successor to Asciidoctor PDF, which uses the web stack to generate PDF from HTML. See https://github.com/mogztter/asciidoctor-web-pdf.

andymc12:
I tend to agree with Markus that we should not disable this.  Perhaps we could replace the text with an image (taken from the HTML output?) as a temporary workaround until a more permanent solution is found?

spericas:
What if we try a tool to convert HTML into PDF as an interim solution?

chkal:
@mojavelinux Thanks a lot for your comments. These references were very helpful.

@andymc12 I don't think that manually replacing the math expressions with images will work very well. Just have a look at the [3.0.0-M1 spec document ](https://jakarta.ee/specifications/restful-ws/3.0/restful-ws-spec-3.0-M1.pdf) (especially the `\mbox` occurrences on pages 29-32).

I spent some time trying to get better results with the `docker-asciidoctor` image and `asciidoctor-mathematical`. But there is still quite some broken output. Maybe the issue is with the document itself which (AFAIK) was converted from Latex with some kind of tooling (pandoc?).

However, I still think that having just the HTML without a PDF is better than having a broken PDF.

chkal:
If anyone wants to work on improving the PDF output, this is what I did:

```sh
cd ./jaxrs-spec/
docker run -it --rm -v $(pwd):/documents/ asciidoctor/docker-asciidoctor
cd /documents/

apk add ruby-dev build-base graphicsmagick graphicsmagick-dev
gem install prawn-gmagick prawn-svg

asciidoctor-pdf \
  -r asciidoctor-mathematical \
  -a mathematical-format=svg \
  -a imagesDir=/documents/src/main/asciidoc/images/ \
  -a pdf-stylesdir=/documents/src/theme/ \
  -a pdf-style=jakartaee \
  -D /documents/target/ \
  src/main/asciidoc/spec.adoc 
```

spericas:
@chkal Converting the HTML output to PDF using a browser (headless or otherwise) produces a decent output. Unfortunately, I'm not sure how to automate this using Maven. I propose that we leave the PDF generation as it is but manually override the output PDF file before submitting it for final publication. I can do that as part of the 3.0 release process.

arjantijms:
Retargeting this to 3.1, as 3.0 has been delivered.

spericas:
@chkal Should we close this PR? There isn't consensus to disable this AFAICT.

chkal:
I'm fine with closing it. However, please note that the PDF which is currently generated from the build is broken. We just should make sure that this PDF isn't published anywhere.

spericas:
@chkal Roger. Process is fairly manual so we can always control that.

PR #880, opened 2020-06-13
Fix for various spec document formatting issues

chkal:
I spent quite some time today and compared the current HTML version of the spec with the old JAX-RS 2.1 spec document. Especially the chapters containing many math expressions. I found many formatting issues, broken references and invalid AsciiDoc content, mostly caused by the automatic conversion from LaTeX to AsciiDoc.

This pull requests contains fixes for all the issues I discovered. Please note that I just reviewed a few sections (but IMO the most relevant ones, especially section 3.7). So there may be many other issues which nobody noticed so far.

I created PDF versions from the rendered HTML, so I could upload the resulting files to a PDF compare service to help you with your review. The diff is not perfect, but I guess it will be helpful to understand which parts were broken.

You can see the differences in the output here:

https://draftable.com/compare/SNCaeIcenRCw


spericas:
Thanks Christian!

chkal:
Thanks for reviewing!

PR #883, opened 2020-06-20
Updated references to Jakarta specs and adjusted status section

chkal:
Another pull request for cleaning up the spec document:

* References to other Jakarta specs have been updated to the jakarta namespace version (corresponding new major version)
* The "Status" section has been updated (spec version, mailing list, URLs, etc.)
* The reference to the "REST Wiki" has been removed (broken link, website seems to be gone)

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_bibliography.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_bibliography.adoc
@@ -32,7 +32,7 @@
 - [[[bib8,8]]]  Jakarta™ Servlet Specification, Version 4.0. Available at:
                https://jakarta.ee/specifications/servlet/4.0
 
-- [[[bib9,9]]]  Jakarta XML Web Services, Version 3.0
+- [[[bib9,9]]]  Jakarta™ XML Web Services, Version 3.0
                https://jakarta.ee/specifications/xml-web-services/3.0/.
 
 - [[[bib10,10]]]  S. Bradner. RFC 2119: Keywords for use in RFCs to Indicate Requirement Levels. RFC, IETF,
@@ -64,8 +64,8 @@
 - [[[bib19,19]]]  Jakarta™ JSON Binding Specification, Version 2.0. Available at:
                https://jakarta.ee/specifications/jsonb/2.0/
 
-- [[[bib20,20]]]  Jakarta Activation, Version 2.0
+- [[[bib20,20]]]  Jakarta™ Activation, Version 2.0
                https://jakarta.ee/specifications/activation/2.0/
 
-- [[[bib21,21]]]  Jakarta XML Binding, Version 3.0
+- [[[bib21,21]]]  Jakarta™ XML Binding, Version 3.0
                https://jakarta.ee/specifications/xml-binding/3.0/
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_bibliography.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_bibliography.adoc
@@ -14,8 +14,6 @@
 - [[[bib1,1]]] R. Fielding. Architectural Styles and the Design of Network-based Software Architectures. Ph.d
                dissertation, University of California, Irvine, 2000. See https://roy.gbiv.com/pubs/dissertation/top.htm.
 
-- [[[bib2,2]]] REST Wiki. Web site. See http://rest.blueoxen.net/cgi-bin/wiki.pl.
-
 - [[[bib3,3]]] Representational State Transfer. Web site, Wikipedia. See
                https://en.wikipedia.org/wiki/Representational State Transfer.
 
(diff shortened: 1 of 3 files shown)


jaxrs-spec/src/main/asciidoc/chapters/appendix/_bibliography.adoc
```diff
@@ -32,7 +32,7 @@
 - [[[bib8,8]]]  Jakarta™ Servlet Specification, Version 4.0. Available at:
                https://jakarta.ee/specifications/servlet/4.0
 
-- [[[bib9,9]]]  Jakarta XML Web Services, Version 3.0
+- [[[bib9,9]]]  Jakarta™ XML Web Services, Version 3.0
```
mkarg, line 35: According to this page https://jakarta.ee/ it has to be ® instead of ™, as Jakarta is a *registered* trademark, actually.

chkal: I added ™, because we already used it for some other entries in the bibliography section and because of the note in #804. Not sure which would we should use TBH.

mkarg: According to [WIPO](https://www3.wipo.int/branddb/en/) Jakarta is a *registered* trademark in the US and other countries, so it must bear the ® to clearly indiciate this.

jaxrs-spec/src/main/asciidoc/chapters/introduction/_status.adoc
```diff
@@ -11,20 +11,20 @@
...
 
 The corresponding Javadocs can be found online at:
 
-https://jax-rs.github.io/apidocs/2.1
+https://jakarta.ee/specifications/restful-ws/3.0/apidocs/
 
 The reference implementation can be obtained from:
 
-https://jersey.github.io
+https://eclipse-ee4j.github.io/jersey/
```
andymc12, line 25: Thanks for updating the link.  My comment here is not necessarily about your change, but about whether we should list "reference implementation" as such.  My understanding is that Jakarta is moving away from _reference_ implementation, and instead is moving toward _compatible_ implementations - similar to MicroProfile.  Should we turn this into a list of compatible implementations?  Or maybe just remove it since we don't know yet who will be implementing it?  Wdyt?

chkal: Good point. I also wondered if we should change the wording. However, I noticed that there are several wordings in the spec that need to be adjusted and therefore decided to do this in a subsequent pull request and/or discuss this issue on the mailing list first. For example:

- "During the course of this **JSR** we received..."
- "**Expert Group** Members"
- "The **reference implementation** can be obtained"

jaxrs-spec/src/main/asciidoc/chapters/appendix/_bibliography.adoc
andymc12: I chatted with @kwsutter about this.  He pointed me to this page: https://www.eclipse.org/legal/logo_guidelines.php

It says that we can use either one (® or ™).  Kevin also mentioned that we only need the trademark symbol  on the first use for a document or section. 

I don't have a strong preference either way on the ® vs ™ question.  One the one hand, it might be nice to be consistent with the jakarta.ee page - on the other hand, if it is already ™ elsewhere in the spec, I don't see enough justification to change it everywhere.  I guess my preference is for consistency in general - we shouldn't use ™ in some parts of the spec and ® in others.

Thanks @chkal!

jaxrs-spec/src/main/asciidoc/chapters/introduction/_status.adoc
spericas: I agree, we should clean this up. @chkal If you can create an issue with all these I can help.

jaxrs-spec/src/main/asciidoc/chapters/appendix/_bibliography.adoc
mkarg: Andy, there is a misunderstanding. The text says "the **appropriate** ™ or ® symbol". "Appropriate" does not imply we have a free choice. From a legal aspect. ™ *must* be used for *trademarks in general*, while ® *must* be used for *registered* trademarks. So in our case, it *must* be ®.

chkal: I'm fine with using ® instead of ™, especially for being consistent with the Jakarta website. I just update the pull request.

jaxrs-spec/src/main/asciidoc/chapters/introduction/_status.adoc
chkal: Ok, great! I'll file a separate issue for that.

chkal:
I just replaced Jakarta™ with Jakarta® as mentioned by @mkarg.

@andymc12 @spericas Could you resubmit your votes? Thanks!

PR #884, opened 2020-06-22
Make WebApplicationException serializable

brange:
By marking the Response object as transient, since Response isn't
serializable

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/WebApplicationException.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/WebApplicationException.java
@@ -32,7 +32,7 @@ import jakarta.ws.rs.core.Response;
 public class WebApplicationException extends RuntimeException {
 
     private static final long serialVersionUID = 8273970399584007146L;
-    private final Response response;
+    private final transient Response response;
 
     /**
      * Construct a new instance with a default HTTP status code of 500 and a default message generated from the HTTP status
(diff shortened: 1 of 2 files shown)


andymc12:
Hi @brange - I think that this is a good change, but I am not sure that we can put it into master, since we're not allowed to change the API in Jakarta EE 9. Can you open a new issue for this so that we can discuss it more? 

Thanks!

chkal:
I agree with @andymc12. Looks great, but unfortunately we will most likely have to put this into the 3.1 branch instead.

brange:
I understand, though I don't really see this as a "change to the API", it's more of a bug fix. 
`WebApplicationException` should be serializable since it extends RuntimeException (-> Exception -> Throwable implements Serializable).

I would understand if I changed the `Response` class to implement Serializable, that would be a change in the API.

Either way, It's fine by me if we change the merge target to the 3.1 branch.

chkal:
@brange I agree that this could be defined as a bug fix. However, the change will most likely break the signature tests and therefore the TCK. And for 3.0 this is unfortunately not allowed. :disappointed:

brange:
I understand. should we change the merge target to 3.1? Would that be ok?
I've a work around in my code, so I'm not really depending on this.

brange:
Changed merge target to 3.1-SNAPSHOT!

brange:
you're welcome!

PR #886, opened 2020-06-28
Updated wording of old JSR terms

chkal:
This PR replaces a few terms from the old JSR days with their Jakarta EE equivalent:

- expert group -> project team
- JSR -> specifcation
- reference implementation -> compatible implementation

There are still a few occurences of the term "expert group" and "JSR", but all of them refer to JAX-RS 2.1 and earlier, so IMO we should keep them this way.

BTW: We still need to add an up-to-date version of the project team, but this can be done in a subsequent PR.


jaxrs-spec/src/main/asciidoc/chapters/introduction/_project_team.adoc
```diff
@@ -8,13 +8,15 @@
...
-=== Expert Group Members
+[[project_team]]
+=== Project Team Members
 
-This specification is being developed as part of JSR 370 under the Java
-Community Process. It is the result of the collaborative work of the
-members of the JSR 370 Expert Group. The following are the present
-expert group members:
+This specification is being developed as part of Jakarta EE under the
+Jakarta EE Specification Process. It is the result of the collaborative work
```
andymc12, line 15:
```suggestion
This specification is being developed as part of Jakarta RESTful Web Services project under the
Jakarta EE Specification Process. It is the result of the collaborative work
```

I think it is fine as it is written, but thought it might be good to state the project name here since I think it maps more directly to the JSR, and it might sound better than having "Jakarta EE" twice in the same sentence.

chkal: Good idea! I just applied the suggested change!

chkal:
@andymc12 @spericas I just applied the suggested wording change. Could you cast your votes again? Thanks!

jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-1.1-release.adoc
```diff
@@ -14,7 +14,7 @@
 * <<status>>: Updated URLs to JSR pages, etc.
 * <<non_goals>>: Removed Client APIs as non-goal.
 * <<terminology>>: Added new terminology.
-* <<expert_group>>: List 2.0 expert group members.
+* <<project_team>>: List 2.0 expert group members.
```
mkarg, line 17: Why do the replace the term *here*? There never was a 2.0 project team, but only an expert group.

jaxrs-spec/src/main/asciidoc/chapters/introduction/_status.adoc
```diff
@@ -20,11 +20,11 @@ The corresponding Javadocs can be found online at:
 
 https://jakarta.ee/specifications/restful-ws/3.0/apidocs/
 
-The reference implementation can be obtained from:
+A compatible implementation can be obtained from:
```
mkarg, line 23: Maybe we should omit this, unless we are fair and list *all* compatible implementations?

jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-1.1-release.adoc
chkal: I renamed the "Expert Group Members" chapter to "Project Team Members", which makes sense IMO, as we are talking about a project team from now on. And the changelog simply refers to this chapter, which I think is acceptable, even if the reference also has the new name.

jaxrs-spec/src/main/asciidoc/chapters/introduction/_status.adoc
chkal: I would be fine with both.

jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-1.1-release.adoc
mkarg: I know that. My question was a philosophical one, whether it could be incorrect to change *historical* chapters like this one. Actually at time of writing *that* chapter, the topics wasn't titled "Project Team", so it feels strange that it refers to the *new* name now. I think it would be more correct that it navigates to "Project Team" but still *says* "Expert Group".

chkal: Got it. I agree that we could change the title of the link without changing the name of the section. However, this also feels a bit inconsistent.

Issue #745, opened 2019-02-22
Clarify more about SseBrodcaster.onClose() behavior


jimma:
SseBroadcaster.onClose() has this description: 
```
    /**
     * Register a listener, which will be called when the SSE event output has been closed (either by client closing the
     * connection or by calling {@link SseEventSink#close()} on the server side.
     * <p>
     * This operation is potentially slow, especially if large number of listeners get registered in the broadcaster. The
     * {@code SseBroadcaster} implementation is optimized to efficiently handle small amounts of concurrent listener
     * registrations and removals and large amounts of registered listener notifications.
     *
     * @param onClose consumer taking single parameter, a {@link SseEventSink}, which was closed.
     */
    void onClose(Consumer<SseEventSink> onClose);
```
If I undestandand correctly, from the javadoc,  these register consumers should be called when the SseEventSink is closed if this SseEventSink is in the broadcast target list.  Is this api intended to do nothing about SseBroadCaster.close() ?  Should these listeners registered in SseBroadcaster care  SseEventSink's close event ?

spericas:
> SseBroadcaster.onClose() has this description:
> 
> ```
>     /**
>      * Register a listener, which will be called when the SSE event output has been closed (either by client closing the
>      * connection or by calling {@link SseEventSink#close()} on the server side.
>      * <p>
>      * This operation is potentially slow, especially if large number of listeners get registered in the broadcaster. The
>      * {@code SseBroadcaster} implementation is optimized to efficiently handle small amounts of concurrent listener
>      * registrations and removals and large amounts of registered listener notifications.
>      *
>      * @param onClose consumer taking single parameter, a {@link SseEventSink}, which was closed.
>      */
>     void onClose(Consumer<SseEventSink> onClose);
> ```
> If I undestandand correctly, from the javadoc, these register consumers should be called when the SseEventSink is closed if this SseEventSink is in the broadcast target list. 

 Correct.

> Is this api intended to do nothing about SseBroadCaster.close() ? 

 No, there wouldn't necessarily be an `SseEventSink` to pass to those consumers right?

> Should these listeners registered in SseBroadcaster care SseEventSink's close event ?

 Are you asking whether this `onClose(Consumer<SseEventSink> onClose)` method makes any sense at all?

jimma:
> No, there wouldn't necessarily be an `SseEventSink` to pass to those consumers right?
Ah, right. 
[...]

PR #887, opened 2020-07-11
[745] Clarify more about SseBrodcaster.onClose() behavior

NicoNes:
Issue: https://github.com/eclipse-ee4j/jaxrs-api/issues/745

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/sse/SseBroadcaster.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/sse/SseBroadcaster.java
@@ -36,7 +36,7 @@ import java.util.function.Consumer;
 public interface SseBroadcaster extends AutoCloseable {
 
     /**
-     * Register a listener, which will be called when an exception was thrown by a given SSE event output when trying to
+     * Register a listener, which will be called when an exception is thrown by a given SSE event output when tries to
      * write to it or close it.
      * <p>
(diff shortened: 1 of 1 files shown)


jaxrs-api/src/main/java/jakarta/ws/rs/sse/SseBroadcaster.java
```diff
@@ -36,7 +36,7 @@
 public interface SseBroadcaster extends AutoCloseable {
 
     /**
-     * Register a listener, which will be called when an exception was thrown by a given SSE event output when trying to
+     * Register a listener, which will be called when an exception is thrown by a given SSE event output when tries to
```
andymc12, line 39:
```suggestion
     * Register a listener, which will be called when an exception is thrown by a given {@link SseEventSink} when this SseBroadcaster tries to
```

"when tries" isn't clear who/what is trying.  My suggestion is to use "when this SseBroadcaster tries" instead.

NicoNes: @andymc12  you're right it makes more sense like that.

Change done !

Thanks

NicoNes:
@andymc12 you're right it makes more sense like that.

Change done !

Thanks

chkal:
Thanks a lot @NicoNes

andymc12:
This PR has met all the required approvals and no negative votes. Merging on behalf of @NicoNes

PR #891, opened 2020-08-03
[#890] All array types to be used as parameters

### LINKED ISSUE #890: Support for arrays as parameters
Section 3.2 says the following:

> Valid parameter types for each of the above annotations are listed in
the corresponding Javadoc, however in general (excluding `@Context`) the
following types are supported:
> 1.  Types for which a `ParamConverter` is available via
a registered `ParamConverterProvider`. See Javadoc for these classes for
more information.
> 2.  Primitive types.
> 3.  Types that have a constructor that accepts a single
`String` argument.
> 4.  Types that have a static method named `valueOf` or
`fromString` with a single `String` argument that return an instance of
the type. If both methods are present then `valueOf` MUST be used unless
the type is an enum in which case `fromString` MUST be usedfootnote:[Due
to limitations of the built-in `valueOf` method that is part of all Java
enumerations, a `fromString` method is often defined by the enum
writers. Consequently, the `fromString` method is preferred when
available.].
> 5.  `List<T>`, `Set<T>`, or `SortedSet<T>`, where `T` satisfies
1,3 or 4 above.

It seems like we should support array types in places where we support `List` types (e.g. `@CookieParam`, `@FormParam`, `@HeaderParam`, `@MatrixParam`, and `@QueryParam` - `@BeanParam` and `@PathParam` do not support Lists, so should not support arrays either).

The MicroProfile Metrics community incorrectly assumed that arrays were already supported (see conversation here: https://groups.google.com/forum/#!topic/microprofile/yxoelXvk3o8 ).  

Arrays already seem to be supported in Apache CXF and RESTEasy. I have not tested Jersey, but I assume that they are not supported based on the MP Metrics conversation.

If nobody objects, I can open a PR to update the spec text and Javadoc.  Since this changes behavior, it would need to be released post-EE9.

andymc12:
Fixes #890 - updates to spec and javadoc enabling users to specify array types as parameters.

The javadoc contained this text: "The resulting collection is read-only."  I left it alone, but I'm not sure that I like it... We could remove that text or re-word it to something like "If it is a `List`, `Set` or `SortedSet` it will be read-only" to avoid any confusion about "read-only arrays". 
Also, I think "read-only" is not the right word here - it should probably be "immutable" or "unmodifiable". "Read-only" implies (to me) that the objects in the collection cannot be modified, but we cannot really guarantee that.  IIUC, the intent is to tell the user that they can't change things like the order of the collection - or the number of entries, etc. By Java Collections terminology, that would be "unmodifiable".  But even that restriction doesn't seem to make a lot of sense to me. If we lifted that restriction, users could remove query param objects from the collection and pass the modified collection to other methods (like listeners) that would then have fewer entries to filter through.  That would be an argument for just removing the sentence outright.  What do you all think?

Lastly, while updating the release notes page, I noticed that we still have a "since 2.2" release note page.  I'll plan to remove that (from master and 3.1-SNAPSHOT) in a future PR.

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/resources/_resource_field.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/resources/_resource_field.adoc
@@ -59,8 +59,8 @@ to limitations of the built-in `valueOf` method that is part of all Java
 enumerations, a `fromString` method is often defined by the enum
 writers. Consequently, the `fromString` method is preferred when
 available.].
-5.  `List<T>`, `Set<T>`, or `SortedSet<T>`, where `T` satisfies
-1,3 or 4 above.
+5.  `List<T>`, `Set<T>`, `SortedSet<T>` or an array `T[]`, where
+`T` satisfies 1, 3 or 4 above.
 
 The `DefaultValue` annotation may be used to supply a default value for
 some of the above, see the Javadoc for `DefaultValue` for usage details
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -35,8 +35,8 @@ import java.lang.annotation.Target;
  * example, {@link Integer#valueOf(String)})</li>
  * <li>Have a registered implementation of {@link jakarta.ws.rs.ext.ParamConverterProvider} JAX-RS extension SPI that
  * returns a {@link jakarta.ws.rs.ext.ParamConverter} instance capable of a "from string" conversion for the type.</li>
- * <li>Be {@code List<T>}, {@code Set<T>} or {@code SortedSet<T>}, where {@code T} satisfies 2, 3, 4 or 5 above. The
- * resulting collection is read-only.</li>
+ * <li>Be {@code List<T>}, {@code Set<T>}, {@code SortedSet<T>} or {@code T[]} array, where {@code T} satisfies 2, 3, 4
+ * or 5 above. The resulting collection is read-only.</li>
  * </ol>
  *
  * <p>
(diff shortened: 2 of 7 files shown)


sdaschner:
LGTM

jaxrs-spec/src/main/asciidoc/chapters/resources/_resource_field.adoc
```diff
@@ -59,8 +59,8 @@ to limitations of the built-in `valueOf` method that is part of all Java
 enumerations, a `fromString` method is often defined by the enum
 writers. Consequently, the `fromString` method is preferred when
 available.].
-5.  `List<T>`, `Set<T>`, or `SortedSet<T>`, where `T` satisfies
-1,3 or 4 above.
+5.  `List<T>`, `Set<T>`, `SortedSet<T>` or an array (i.e. `T[]`), where
```
spericas, line 62: Minor comment: I would drop the "i.e." part, and say as in the Javadocs "an array T[]" explicitly.

mkarg:
I do not think we should change the line about read-only. It does *not* imply read-only referenced elements, it only implies that the set of references cannot be changed. I do not see that anybody had a problem with that so far, so i wouldn't touch it. Also, arrays are not collections. For arrays, we should say that the resulting array MUST be read-only, so implementations will provide a copy to keep its internal consistency. This is essential to know for server programmers so they can rely on that mechanism and do not have to provide safety copies in all apps.

jaxrs-spec/src/main/asciidoc/chapters/resources/_resource_field.adoc
andymc12: Sounds good.  Fixed.

andymc12:
> I do not think we should change the line about read-only. It does not imply read-only referenced elements, it only implies that the set of references cannot be changed. I do not see that anybody had a problem with that so far, so i wouldn't touch it. Also, arrays are not collections. For arrays, we should say that the resulting array MUST be read-only, so implementations will provide a copy to keep its internal consistency. This is essential to know for server programmers so they can rely on that mechanism and do not have to provide safety copies in all apps.

That works for me.

> Can you please separate the "arrays" changes from the "exception mapper" changes into two distinct commits, so there is no risk of inadvertently cherry-picking each other when just one of both is wanted? Thanks. :-)

Good idea!  Done.

Please let me know if you see anything else that needs changing. Thanks!

andymc12:
> LGTM. We should create another issue to add a TCK for this.

@spericas Good idea, I opened https://github.com/eclipse-ee4j/jakartaee-tck/issues/413 to address this.

andymc12:
It's been two weeks and there are 5 +1 votes and no -1 votes.  I'll merge this.  Thanks all!

PR #893, opened 2020-08-13
Updating dependencies to use final version of JAXB and Activation

spericas:
Updating dependencies to use final version of JAXB and Activation from staging. We need -Pstaging. Will fast track.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ cache:
 
 install:
   - cd $TRAVIS_BUILD_DIR/jaxrs-api
-  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+  - mvn install -Pstaging -DskipTests=true -Dmaven.javadoc.skip=true -B -V
   - cd $TRAVIS_BUILD_DIR/examples
   - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
   - cd $TRAVIS_BUILD_DIR/jaxrs-spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
 
 script:
   - cd $TRAVIS_BUILD_DIR/jaxrs-api
-  - mvn verify -B
+  - mvn verify -Pstaging -B
   - cd $TRAVIS_BUILD_DIR/examples
   - mvn verify -B
   - cd $TRAVIS_BUILD_DIR/jaxrs-spec
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -182,6 +182,7 @@
                         </plugin>
                         <plugin>
                             <artifactId>maven-surefire-plugin</artifactId>
+                            <version>2.21.0</version>
                             <configuration>
                                 <argLine>--add-modules java.xml.bind --add-opens jakarta.ws.rs/jakarta.ws.rs.core=java.xml.bind</argLine>
                             </configuration>
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -538,9 +538,9 @@
         <spec.version>2.2</spec.version>
         <spec.version.revision /> <!-- e.g. (Rev a) -->
 
-        <jaxb.api.version>3.0.0-RC3</jaxb.api.version>
+        <jaxb.api.version>3.0.0</jaxb.api.version>
         <jaxb.impl.version>3.0.0-M4</jaxb.impl.version>
-        <activation.api.version>2.0.0-RC3</activation.api.version>
+        <activation.api.version>2.0.0</activation.api.version>
         <legal.doc.folder>${project.basedir}/..</legal.doc.folder>
     </properties>
 


spericas:
Any suggestions about those failures with JDK 9 and 10? Likely related to some plugins.

mkarg:
> Any suggestions about those failures with JDK 9 and 10? Likely related to some plugins.

I wonder who decided to use a *milestone* plugin? Not me. That's why I was against adopting the parent pom: It implis decisions out of *our* control. All was fine on v9 and v10 back in the gold old times of self-selected plugins. ;-)

There is a M5 meanwhile, maybe it solves the problem

andymc12:
Possibly this is the issue?
```
[INFO] -------------------------------------------------------
713[INFO]  T E S T S
714[INFO] -------------------------------------------------------
715[WARNING] Corrupted STDOUT by directly writing to native stream in forked JVM 1. See FAQ web page and the dump file /home/travis/build/eclipse-ee4j/jaxrs-api/jaxrs-api/target/surefire-reports/2020-08-13T20-22-40_337-jvmRun1.dumpstream
```

The unit tests could use some cleaning - I see unused imports and other compiler warnings - and some System.out statements that might be corrupting the output.  I'll submit a PR that cleans things up and see if that runs into the same problem.

chkal:
Basically looks good. But I guess we should merge #894 before to fix the test failures.

andymc12:
#894 is merged. Hopefully that resolves the test failures for this PR.

chkal:
@spericas Could you rebase this PR, so we get the fixes from #894?

spericas:
> @spericas Could you rebase this PR, so we get the fixes from #894?

It was actually a problem with Surefire, see commits.

jaxrs-api/pom.xml
```diff
@@ -182,6 +182,7 @@
                         </plugin>
                         <plugin>
                             <artifactId>maven-surefire-plugin</artifactId>
+                            <version>2.21.0</version>
```
mkarg, line 185: Why not 2.22.2?

spericas: Because it works just fine with 2.21.0 and haven't tried 2.22.2?

mkarg: I agree it was a really bad Idea to use a pre-release of Surefure, but as 2.22.2 is just a bug fix I assume it will work just as good. I would kindly ask you to actually try it out and use that one instead of needlessly falling back to an older version of that plugin, as it definitively *has* more bugs -- that's why dot-two was released. :-)

spericas: It's really unnecessary as this version works just fine for us and I don't want to delay this any longer. But I'll do it if this is a blocker for _you_.

spericas: OK, just for the record it _fails_ for me locally with 2.22.2. Others feel free to try. Marking this conversation as resolved.

mkarg:
> > @spericas Could you rebase this PR, so we get the fixes from #894?
> 
> It was actually a problem with Surefire, see commits.

Actually it was not, see https://github.com/spericas/jaxrs-api/pull/1. :-)

.travis.yml
```diff
@@ -21,15 +21,15 @@ cache:
 
 install:
   - cd $TRAVIS_BUILD_DIR/jaxrs-api
-  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+  - mvn install -Pstaging -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
mkarg, line 24: Why do we need this for *Travis*? IIUC this is only needed if we want the CI system to stage the build, but we use *Jenkins* for staging, not *Travis*.

jaxrs-api/pom.xml
mkarg, line 185: Please adopt my real solution instead, see https://github.com/spericas/jaxrs-api/pull/1.

.travis.yml
chkal: Doesn't this add the staging repo we need to fetch our dependencies which are only in staging yet? In this case we will need `-Pstaging`.

mkarg: @spericas Can you please clearly describe *in the PR description* **why** "We need -Pstaging."? Thanks. :-)

spericas: @mkarg It's in the title of the PR, our final dependencies are in staging until all EE 9 is ready. This is how all EE 9 API are building.

jaxrs-api/pom.xml
spericas: It's taking way too much time for a simple change. Let's get this in and then we can review your PR to improve this later.

spericas:
> > > @spericas Could you rebase this PR, so we get the fixes from #894?
> > 
> > 
> > It was actually a problem with Surefire, see commits.
> 
> Actually it was not, see [spericas#1](https://github.com/spericas/jaxrs-api/pull/1). :-)

Realized it was a PR on my branch, it has been merged. We need all approvals again.

.travis.yml
mkarg: You misunderstood the question. I didn't ask why **we** are in staging, but why you think you must add the `-P` flag to the **Travis** build script.  I understand it for **Jenkins**. Travis is not able to put anything into the staging area, nor did it fail without that flag. Travis just checks whether the build fails or not, and it does this pretty well without that flag.

jaxrs-api/pom.xml
mkarg: I will not accept a workaround getting in hastily once the actual solution is on the table -- it is away a single click, so why not doing it?

Issue #792, opened 2019-08-16
Clarify StatusInfo#equals


MarkusKull:
It is a known pitfall to compare Response#getStatusInfo with one of the Status-Enums as this will return unexpected results when reasonphrase differs. Happens especially with Tomcat8.5+.

See https://blog.zenika.com/2017/05/22/some-notes-about-jax-rs-http-statuses-and-tomcat/

Even if one wanted, this cannot be fixed by overriding `equals` because `Enum`-Baseclass of `Status` declares `equals` as final. Thus depending on order of comparison, it would still behave unexpectedly.

1) Please extend javadoc of `Response#getStatusInfo()` that the returned StatusType must not be  compared with a `Status` because this comparison may return unequal when reasonphrase differs; Instead the statuscodes should be compared.
2) Deprecate StatusInfo#equals with the same javadoc. 
```
    /**
     * @deprecated May return false when reasonPhrase differs. Instead compare statusCode directly.
     */
    @Deprecated
    @Override
    abstract boolean equals(Object obj);
```

mkarg:
I wonder why you not simply ask to make `equals` *work correctly*?

MarkusKull:
@mkarg Try it: "Cannot override the final method `equals` from `Enum<Status>`"

Overriding only `StatusInfo#equals` would work partly for that comparison-order, but not when users call `Status#equals`. For a transitive `equals`, more has to be done:
- change `Status` to an non-instantiable class instead of an enum and change former enum-members to constants of type `StatusType`. (Except for compatibility, one really ought do this because the current enum wrongly implies that there are never any other httpstatus)
- change `StatusInfo` from interface to final class

That would help insofar as one could now call `equals` transitively. Still, language lawyers will complain that equals only compares statusCode but not reasonPhrase. Will you then drop reasonPhrase because useless? But reasonPhrase is already used as seen in some jersey-issues. All in all, the compatibility hassles caused by such changes are grave.

mkarg:
I think it would be worth letting the committers discuss that. ;-)

unexist:
Maybe it is still worth pushing this issue.. ;)

chkal:
I agree that there seems to be way to fix `equals()` (or am I missing something).

Therefore, I'm fine with (1) and (2) proposed in the original issue description. Anyone wants to open a pull request against the `3.1-SNAPSHOT` branch?

PR #905, opened 2020-09-10
Clarification with reminder test for StatusInfo#equals

unexist:
See [#792]


jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
```diff
@@ -67,6 +67,11 @@ protected Response() {
 
     /**
      * Get the complete status information associated with the response.
+     * <p>
+     * Comparing {@link StatusType} directly might unexpectedly return {@code false}.
```
chkal, line 71:
```suggestion
     * Comparing {@link StatusType} directly might unexpectedly return {@code false} if only the reason phrase differs.
```

The note should also explain why this can happen.

jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
```diff
@@ -67,6 +67,11 @@ protected Response() {
...
+     *
+     * <b>Instead compare {@code statusCode} directly</b>.
```
chkal, line 73:
```suggestion
     * Instead compare {@code statusCode} directly.
```

IMO this shouldn't be bold.

jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
```diff
@@ -1131,6 +1136,13 @@ public ResponseBuilder status(final Status status) {
         public default Status toEnum() {
             return Status.fromStatusCode(getStatusCode());
         }
+
+        /**
+         * @deprecated May return false when reasonPhrase differs. <b>Instead compare {@code statusCode} directly.</b>
```
chkal, line 1141:
```suggestion
         * @deprecated May return false when reasonPhrase differs. Instead compare {@code statusCode} directly.
```

See above.

chkal (commented):
I added a few comments.

@eclipse-ee4j/ee4j-jaxrs-committers Other thoughts?

chkal (approved):

jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
```diff
@@ -1131,6 +1136,13 @@ public ResponseBuilder status(final Status status) {
         public default Status toEnum() {
             return Status.fromStatusCode(getStatusCode());
         }
+
+        /**
+         * @deprecated May return false when reasonPhrase differs. Instead compare {@code statusCode} directly.
+         */
+        @Deprecated
+        @Override
+        abstract boolean equals(Object obj);
```
andymc12, line 1145: I'm still mulling this over, but I think I don't like deprecating the `equals` method.  I like the clarification that users should not depend on `equals` in cases where the status code is the same, but the reason code might be different, but maybe instead of deprecating the `equals` method we could specify that an implementation must check statusCode, reasonPhrase.  Something like:
```
/**
 * Returns true only if this passed-in object is of {@code StatusType}, is non-null and contains the same
 * {@code statusCode}, {@code reasonPhrase}, and {@code family} as this {@code StatusType}.
 *
 * @param obj - the reference object with which to compare.
 * @return true if this object is the same as the obj argument; false otherwise.
 * @since 3.1
 */
abstract boolean equals(Object obj)
```

Wdyt?

chkal: Good point. Yeah, thinking about this more, exactly specifying how `equals()` should behave may be better than deprecating it. So +1 for your suggestion from my side.

chkal:
@unexist BTW: I just saw that the ECA check is failing. Could you double-check that you signed the ECA and added a signed-off header to your commit message as described here:

https://github.com/eclipse-ee4j/jaxrs-api/blob/master/CONTRIBUTING.md

jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
spericas: I don't like deprecating the method `equals` method either. However, I think there's still an issue if an operand is an instance of a class that implements `StatusType` and the other is one of the enum values, even if all the fields match. It would depends on which operand is left vs. right in `equals`. So I think it is possible for the condition as described in the newly proposed Javadoc above to be true and for the method to return false. I'd probably just add a clarification to `getStatusInfo`.

mkarg: I second that. We should not deprecate the equals method.

mkarg (changes requested):
Please do not deprecate the equals method but instead clearly specify it instead.

mkarg:
@unexist Can you please update this PR so it does not contain 38 commits, does not conflict with `master`, will build, and contains the requested changes? Thanks.

spericas:
Cleaning up old PR's. Please re-submit as necessary.

PR #908, opened 2020-09-21
[#907] Make XML Bind pkgs optional for JPMS/OSGi

### LINKED ISSUE #907: Declare XML Binding packages as optional in JPMS and OSGi
Per the [thread on the Jakarta Platform mail list](https://www.eclipse.org/lists/jakartaee-platform-dev/msg02182.html), we may need to make it more clear that Jakarta XML Binding is optional. We addressed this initially in issue #815 and PR #854  
However, we need to make it more clear for JPMS and OSGi users that these packages are optional.

Per Lukas's comment in the mail thread, this can be done by updating the pom.xml and module-info.java files:
> replace `requires transitive jakarta.xml.bind;` with `requires static 
jakarta.xml.bind;` in module-info[1]. That will do the trick for JPMS. 
Marking dependency on xml.bind-api optional in project's pom[2] should 
resolve the problem for OSGi.

andymc12:
Fixes #907.  The change to the surefire line is to ensure that Jakarta XML Binding APIs are present for unit testing. This will make the depending on XML Binding optional in both JPMS and OSGi environments.

I'd like to ask for fast-track approval on this, but I'm not sure if it qualifies or not.  The pom.xml change should qualify, but possibly the module-info.java file is considered part of the API?  I think we'll need this change in before we can release 3.0, so the sooner it gets in the better, but I don't want to thwart our process.

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -134,7 +134,7 @@
                         <plugin>
                             <artifactId>maven-surefire-plugin</artifactId>
                             <configuration>
-                                <argLine>--add-opens jakarta.ws.rs/jakarta.ws.rs.core=java.xml.bind</argLine>
+                                <argLine>--add-modules jakarta.xml.bind</argLine>
                             </configuration>
                         </plugin>
                     </plugins>
@@ -497,6 +497,7 @@
             <artifactId>jakarta.xml.bind-api</artifactId>
             <version>${jaxb.api.version}</version>
             <scope>provided</scope>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>jakarta.activation</groupId>
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -183,7 +183,7 @@
                         <plugin>
                             <artifactId>maven-surefire-plugin</artifactId>
                             <configuration>
-                                <argLine>--add-opens jakarta.ws.rs/jakarta.ws.rs.core=java.xml.bind</argLine>
+                                <argLine>--add-modules jakarta.xml.bind</argLine>
                             </configuration>
                         </plugin>
                     </plugins>
--- a/jaxrs-api/src/main/java/module-info.java
+++ b/jaxrs-api/src/main/java/module-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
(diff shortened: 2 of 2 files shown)


lukasj (approved):
LGTM. Thanks!

lukasj:
@andymc12 hm, what is the reason for having dependency on activation? Shouldn't that one be optional too or removed?

andymc12:
> hm, what is the reason for having dependency on activation? Shouldn't that one be optional too or removed?

Good question... I don't think we really need it.  I wonder if it was an indirect dependency from the past.  I removed the dependency and built the project without error, so we may be able to delete that dependency.  Anybody else know if we need Activation?

When I was searching through the code though I noticed that we got too aggressive in renaming `javax.*` to `jakarta.*` and changed a Java SE XML package in _entity_providers.adoc:
```
`jakarta.activation.DataSource`::
  All media types (`\*/*`).
`jakarta.xml.transform.Source`::
  XML types (`text/xml`, `application/xml` and media types of the form
  `application/*+xml`).
```

should be:

```
`jakarta.activation.DataSource`::
  All media types (`\*/*`).
`javax.xml.transform.Source`::
  XML types (`text/xml`, `application/xml` and media types of the form
  `application/*+xml`).
```

I'll open a separate issue/PR for that change.  Possibly the reference to `jakarta.activation.DataSource` is the reason we have the Activation dependency?  I wouldn't think that the API would need to depend on it though, but the implementation would.

chkal:
@andymc12 The CI build is failing now for JDK9+10. Maybe we need the same adjustments in the `jdk9-jdk10` profile?

lukasj:
@chkal do you (= the project, not "you" personally :-) ) really care about already generally unsupported JDKs? Wouldn't it be better do drop them instead? That could simplify things a bit and help avoiding similar issues in the future... Just asking...

spericas (approved):
LGTM

spericas:
> @chkal do you (= the project, not "you" personally :-) ) really care about already generally unsupported JDKs? Wouldn't it be better do drop them instead? That could simplify things a bit and help avoiding similar issues in the future... Just asking...

Personally I don't see the need to keep JDK 9 and 10 there. Seems everyone if going from 8 to 11, but others may have a different opinion.

andymc12:
Dang... I just updated the PR to fix JDK 9 and 10 (adding the same surefire option as 11+), which also removed the approvals.  

But I agree with Lukas and Santiago that we should pull the 9&10 out of the CI builds, but maybe we can do that in a separate issue/PR (I'm thinking that change is less time-sensitive than this one).

spericas:
@andymc12 We should absolutely fast track any change like this related to the EE 9 release

lukasj (approved):
LGTM

mkarg (commented):
I understand the need to fix it, but as I am not a JPMS expert, my vote is just +-0.

andymc12:
@chkal @spericas can you re-approve now that the Java 9/10 builds are passing?

spericas (approved):

chkal (approved):

jansupol (approved):

PR #910, opened 2020-09-21
[#909] Rename Java SE XML transform class in spec

### LINKED ISSUE #909: Rename `jakarta.xml.transform.Source` to `javax.xml.transform.Source` in _entity_providers.adoc
It looks like we inadvertently renamed a package that is still in the Java SE APIs. This should be changed back to `javax.xml.transform.Source`.

andymc12:
Fixes #909. Confirmed that this is a Java SE class: https://docs.oracle.com/en/java/javase/11/docs/api/java.xml/javax/xml/transform/Source.html

This too, really ought to be fast-tracked, but it is a spec change... I think we ought to add a clause for fast-tracking trivial bug fixes to avoid waiting two weeks for obvious fixes (i.e. fixing typos, etc.), but for now, I can wait.  :)

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/providers/_entity_providers.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/providers/_entity_providers.adoc
@@ -140,7 +140,7 @@ type combinations:
   All media types (`\*/*`).
 `jakarta.activation.DataSource`::
   All media types (`\*/*`).
-`jakarta.xml.transform.Source`::
+`javax.xml.transform.Source`::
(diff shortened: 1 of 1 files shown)


chkal (approved):
As is clearly a mistake from the namespace migration, I'm +1 for fast-tracking this one.

spericas (approved):

spericas:
@andymc12 Fast track +1

mkarg (approved):
This clearly was a mistake, so the change is not a real API change, but a bug fix. +1 to fast-track it.

PR #914, opened 2020-10-23
Additionally building on JDK 16

mkarg:
This PR provides all needed changes to also build on JDK 16:

- Travis CI uses Open JDK 16

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ jdk:
   - openjdk13
   - openjdk14
   - openjdk15
+  - openjdk16
 
 cache:
   directories:


chkal (approved):

andymc12 (approved):
Not a blocker by any means, but it is strange that JDK 16 is about 1 minute 40 seconds more than JDK 14 and 15.  We should probably also remove some of the out-of-service JDKs too - like 9, 10, 12, and 13.  But that shouldn't hold up this PR.  Thanks!

mkarg:
> Not a blocker by any means, but it is strange that JDK 16 is about 1 minute 40 seconds more than JDK 14 and 15. We should probably also remove some of the out-of-service JDKs too - like 9, 10, 12, and 13. But that shouldn't hold up this PR. Thanks!

The travis infrastructure runs thousands of builds parallel and is highly unreliable when it comes to timing. For example, it needed *hours* before this PR's tests actually started to run, while it is just a question of seconds usually. So I think we cannot rely on *any* timing-related information taken from Travis.

Regarding out-of-service, I am not against that idea, but I wonder who actually would have a benefit from that. As *only* Travis runs all these tests, and as Travis is free for us, and as Tests run in *parallel* anyways, who would really have a benefit from this?

mkarg:
**Thank you for approving this PR. I will wait with the merge to `master` until 3.0 is through the door *officially*.**

mkarg:
@spericas Can you please publish 3.0, so that `master` will be open for merges again? Thanks.

Issue #928, opened 2020-12-10
Missing API documentation for JAXRS-API 3.0 (jakartified)


senivam:
There are no API docs published for jakartified JAXRS API at https://eclipse-ee4j.github.io/jaxrs-api/apidocs.
The latest version shows only 2.1.6 API docs.

mkarg:
Can anybody remember they technical way how to put the current docs there?

mkarg:
> I guess it is the `gh-pages` branch of this repo?

Ah... right. You added it manually there back then. I will provide a PR doing the same for 3.0.0.

senivam:
this can be closed as fixed.

PR #929, opened 2020-12-16
[#928] Fixed: Missing API documentation for JAXRS-API 3.0 (jakartified)

mkarg:
Fixing #928.


andymc12:
Just a heads up, but I'm probably not going to review 334 file changes. :-)

Is there any way to review the "published" version of these changes - like maybe on a staging server or something?  That might be easier than reviewing raw HTML in GitHub.

senivam:
I think for the content which is basically a large web page this will not be a problem to fix something particular later. 

For now this can be merged as is to fix the whole documentation absence which is more serious issue I suppose.

andymc12 (approved):
It's a shame that there isn't an easier way... Suppose we make a change to fix a typo or clarify something in the Javadoc.  We have to make that change in the source code, but those changes won't be reflected in GH pages without another change.

Ideally, we could automate this process to make it easier for future releases.  In any case, I'll approve it, but as I mentioned, my review was not very thorough.  Like @senivam said, we can fix things later if need be.

Thanks @mkarg for putting this together.

spericas (approved):

senivam (approved):

PR #933, opened 2020-12-27
[Spec] Runtime Extensions

mkarg:
Core description of the `Extension` marker interface, as this is part of release 3.1 according to PR #751.

**The underlying Java code already exists in `master`. This PR *only documents* this already existing feature.**

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
@@ -29,9 +29,20 @@ all of the lifecycle options provided by that framework.
 ==== Automatic Discovery
 
 The annotation `@Provider` is used by a JAX-RS runtime to automatically
-discover provider classes via mechanisms such as class scanning. A
-JAX-RS implementation that supports automatic discovery of classes MUST
-process only those classes that are annotated with `@Provider`.
+discover provider classes via mechanisms such as class scanning.
+
+[[extensions]]
+Providers MAY implement the `Extension` interface to declare it as an
+extension of the runtime.
+Compatible products MUST automatically register a class in
+`jakarta.ws.rs.core.Configuration` runtime configurations if the
+class is a _provider class_ for the `Extension` _service type_ according
+to the rules set out in the description of `ServiceLoader`.
+
+A JAX-RS implementation that supports automatic discovery of classes MUST
+process only those classes that are annotated with `@Provider` or are
+registered via the `Extension` mechanism described above.
+
 
 [[provider_class_constructor]]
 ==== Constructors
--- a/jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
@@ -16,18 +16,9 @@ such as filtering requests, converting representations into Java
 objects, mapping exceptions to responses, etc. A provider can be either
 _pre-packaged_ in the JAX-RS runtime or supplied by an application. All
 _application-supplied_ providers implement interfaces in the JAX-RS API
-and MAY be annotated with `@Provider` for automatic discovery purposes;
-the integration of pre-packaged providers into the JAX-RS runtime is
-implementation dependent.
-
-[[extensions]]
-Providers MAY implement the `Extension` interface to declare it as an
-extension of the runtime.
-
-Compatible products MUST automatically register a class in
-`jakarta.ws.rs.core.Configuration` runtime configurations if the
-class is a _provider class_ for the `Extension` _service type_ according
-to the rules set out in the description of `ServiceLoader`.
+and MAY be annotated with `@Provider` or implement the `Extension`
+interface for automatic discovery purposes; the integration of
+pre-packaged providers into the JAX-RS runtime is implementation dependent.
 
 This chapter introduces some of the basic JAX-RS providers; other
 providers are introduced in Chapter <<client_api>> and Chapter
--- a/jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
@@ -20,6 +20,15 @@ and MAY be annotated with `@Provider` for automatic discovery purposes;
 the integration of pre-packaged providers into the JAX-RS runtime is
 implementation dependent.
 
+[[extensions]]
+Providers MAY be annotated with `@Extension` to declare it as an
+extension of the runtime.
+
+Compatible products MUST automatically register a class in
+`jakarta.ws.rs.core.Configuration` runtime configurations if the
+class is a _provider class_ for the extension _service type_ according
+to the rules set out in the description of `ServiceLoader`.
+
 This chapter introduces some of the basic JAX-RS providers; other
 providers are introduced in Chapter <<client_api>> and Chapter
 <<filters_and_interceptors>>.
--- a/jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
@@ -21,12 +21,12 @@ the integration of pre-packaged providers into the JAX-RS runtime is
 implementation dependent.
 
 [[extensions]]
-Providers MAY be annotated with `@Extension` to declare it as an
+Providers MAY implement the `Extension` interface to declare it as an
 extension of the runtime.
 
 Compatible products MUST automatically register a class in
 `jakarta.ws.rs.core.Configuration` runtime configurations if the
-class is a _provider class_ for the extension _service type_ according
+class is a _provider class_ for the `Extension` _service type_ according
 to the rules set out in the description of `ServiceLoader`.
 
 This chapter introduces some of the basic JAX-RS providers; other
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
@@ -12,7 +12,7 @@
 === Changes Since 3.0 Release
 
 * <<extensions>>: Added requirement that JAX-RS implementations MUST load
-`Extension`s using the `ServiceLoader` mechanism.
+`Extension` implementations using the `ServiceLoader` mechanism.
 * <<exceptionmapper>>: Added requirement that JAX-RS implementations have 
 default exception mappers.
 * <<resource_field>>: Array types may be specified for `@CookieParam`,
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
@@ -11,6 +11,8 @@
 [[changes-since-3.0-release]]
 === Changes Since 3.0 Release
 
+* <<extensions>>: Added requirement that JAX-RS implementations MUST load
+extensions using the `ServiceLoader` mechanism.
 * <<exceptionmapper>>: Added requirement that JAX-RS implementations have 
 default exception mappers.
 * <<resource_field>>: Array types may be specified for `@CookieParam`,
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
@@ -12,7 +12,7 @@
 === Changes Since 3.0 Release
 
 * <<extensions>>: Added requirement that JAX-RS implementations MUST load
-extensions using the `ServiceLoader` mechanism.
+`Extension`s using the `ServiceLoader` mechanism.
 * <<exceptionmapper>>: Added requirement that JAX-RS implementations have 
 default exception mappers.
 * <<resource_field>>: Array types may be specified for `@CookieParam`,


jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
```diff
@@ -20,6 +20,15 @@ and MAY be annotated with `@Provider` for automatic discovery purposes;
 the integration of pre-packaged providers into the JAX-RS runtime is
 implementation dependent.
 
+[[extensions]]
+Providers MAY be annotated with `@Extension` to declare it as an
```
chkal, line 24: This sounds like `Extension` is an annotation, but actually it is some kind of marker interface. Or am I missing something here?

jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
```diff
@@ -20,6 +20,15 @@ and MAY be annotated with `@Provider` for automatic discovery purposes;
...
+extension of the runtime.
+
+Compatible products MUST automatically register a class in
+`jakarta.ws.rs.core.Configuration` runtime configurations if the
+class is a _provider class_ for the extension _service type_ according
+to the rules set out in the description of `ServiceLoader`.
```
chkal, line 30: I'm not sure if I fully understand how this should be used in practice. If the developer creates something like this:

```java
public class MyFilter implements ContainerRequestFilter, Extension {
   // ...
}
```

Which of these files should the developer use for automatic registration?

  1. `META-INF/services/jakarta.ws.rs.ext.Extension`
  2. `META-INF/services/jakarta.ws.rs.container.ContainerRequestFilter`

I guess it should be (1)?

jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
mkarg: Good catch, seems I was drunk when writing the docs. I will fix this.

jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
mkarg: Good catch! In fact, I missed to emphasize the `Extension` word in line 29. I will fix that to make it clear that actually proposal 1 (`META-INF/services/jakarta.ws.rs.ext.Extension`) is the one to use.

jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
mkarg: Done :-)

jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
mkarg: Done :-)

spericas:
Thanks for the new version, the old one was very confusing --especially the use of an annotation. What's the original issue number that led to this PR?

spericas:
> Core description of `@Extension` as this is part of release 3.1.

There's no `@Extension` right?

mkarg:
> Thanks for the new version, the old one was very confusing --especially the use of an annotation. What's the original issue number that led to this PR?

Sorry for the confusion. The feature exists since PR #751, and is simply spoken a cross-vendor generalization of what Jersey does using its proprietary `ForcedAutoDiscoverable` interface.

mkarg:
> > Core description of `@Extension` as this is part of release 3.1.
> 
> There's no `@Extension` right?

Good catch. Forgot to also fix the PR description. Done. :-)

spericas:
> > Thanks for the new version, the old one was very confusing --especially the use of an annotation. What's the original issue number that led to this PR?
> 
> Sorry for the confusion. The feature exists since PR #751, and is simply spoken a cross-vendor generalization of what Jersey does using its proprietary `ForcedAutoDiscoverable` interface.

 Thanks. Right, I'm familiar with `AutoDiscoverable` in Jersey. The one thing that the original discussion did not fully address (or I missed that resolution) is the separation of client and server providers that is supported by `@ConstraintedTo`. Is that something we want to revisit with this new alternative of auto-registration of providers?

spericas:
To follow-up on my previous comment, I think the spec needs to be enhanced as follows:

- Clarify the role of `@ContrainedTo` and the new extensions. If a provider is loaded via the new extension mechanism, will a `@ContrainedTo` annotation be honored? I think it should. 
- We should define a property (just like Jersey does) to turn off this mechanism. The behavior of some apps may be unexpectedly changed by the auto-registration of some unexpected providers.

mkarg:
I do not see a need for a property to switch this mechanism *completely* off on the runtime level, but it might be a good idea that `Feature` vendors follow a "best practice" we define to always provide a per-feature property to switch *that feature* off. So new release will be auto-loaded, but application vendors can decide to *selectively* switch them off. No API or spec change needed, just a bit of communication on our web site. :-)

I do not see a need to explicitly describe the role of `@ContstrainedTo`, as I think it is rather *obvious* that it is respected even with `implements Extension`, neither am I against it. If there is a majority for adding some notes, I will be happy to write them down in this PR.

jansupol:
> I do not see a need for a property to switch this mechanism completely off

Lately, I have only seen too many huge apps (like 50MB+ EARs), each of which consists of multiple modules (multiple Applications). Each module is managed by a different team and they do not communicate with each other. I find it very likely that one team will introduce an Extension that will break the other team's module. The extension can do something simple, such as registering a higher priority MBR). Hunting where the new provider comes from can be very frustrating for the team. The property preventing from loading other team's  Extension seems to be a nice debugging option. Of course, it can be provided by each implementation independently, but I'd welcome an implementation-independent property. Or perhaps you can find a better way to prevent such a situation?

spericas:
> I do not see a need for a property to switch this mechanism _completely_ off on the runtime level, but it might be a good idea that `Feature` vendors follow a "best practice" we define to always provide a per-feature property to switch _that feature_ off.

 But this way of thinking defeats the point of vendor neutrality. If I need to use a vendor property to turn things off, then is this really that much better than letting vendors control the feature in the first place?

> 
> I do not see a need to explicitly describe the role of `@ContstrainedTo`, as I think it is rather _obvious_ that it is respected even with `implements Extension`, neither am I against it. If there is a majority for adding some notes, I will be happy to write them down in this PR.

 I disagree that this is obvious. We are introducing a new mechanism that (at least on the server side) is an alternative to an existing mechanism (`@Provider`) and is suited for environments where class/annotation scanning isn't practical. Therefore, it isn't clear at all the annotations such as `@ContrainedTo` are supposed to be honored IMO.

mkarg:
Okay, convinced. ;-)

mkarg:
> > I do not see a need for a property to switch this mechanism _completely_ off on the runtime level, but it might be a good idea that `Feature` vendors follow a "best practice" we define to always provide a per-feature property to switch _that feature_ off.
> But this way of thinking defeats the point of vendor neutrality. If I need to use a vendor property to turn things off, then is this really that much better than letting vendors control the feature in the first place?

You misunderstood that proposal. I did not propse vendor-specific options. I proposed `Feature`-specific options. So it is about the application vendor to decide, not the vendor.

mkarg:
* `@ConstrainedTo`: The JavaDocs already say: *"If a @ConstrainedTo annotation is not present on a JAX-RS provider type declaration, the declared provider may be used in any run-time context. If such a annotation is present, the JAX-RS runtime will enforce the specified usage restriction."*. I cannot see anything that would be unclear in that description, or might make somebody think this will *not* be true for extensions. So, I do not see how to make it any more clear than that. It makes no sense to extend it to "This is true for extensions, also." as why shouldn't it?

mkarg:
* In what *way* do you want to pass the new switch-off-property?
  * Java SE: `SeBootstrap.Configuration.Builder.property(...)` seems to be obvious.
  * others...?

mkarg:
Proposal: Can we please separate the discussion about adding / changing the underlying feature (e. g. by adding a new property) into a separate issue / PR, but concentrate the discussion in *this* PR around just documenting the feature *as it already exists* in `master`? My target with *this* PR solely is to get the documentation in `master` in sync with its actual Java code. All *further* ideas are really welcome, but please *first* let's sync `master`. Agreed?

**So I kindly ask everybody to vote on this *purely documenting* PR now!**

chkal (approved):
I'm fine with the current version of the pull request.

Regarding `@ContstrainedTo`: I agree that this annotation should be honored also for extensions. But I don't think that this needs to be mentioned in the new section. Especially because `@ContstrainedTo` is only mentioned in the "Summary of Annotations" appendix and nowhere else in the spec document. And the appendix states:

> Can be used to restrict the applicability of a provider to just the Client API or just the Server API. If omitted, a provider can be used in either context.

So IMO it is clear that this annotation applies to ALL providers, regardless of how they were registered.

chkal (approved):

jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
```diff
@@ -20,6 +20,15 @@ and MAY be annotated with `@Provider` for automatic discovery purposes;
 the integration of pre-packaged providers into the JAX-RS runtime is
 implementation dependent.
 
+[[extensions]]
+Providers MAY implement the `Extension` interface to declare it as an
+extension of the runtime.
+
+Compatible products MUST automatically register a class in
+`jakarta.ws.rs.core.Configuration` runtime configurations if the
+class is a _provider class_ for the `Extension` _service type_ according
```
ronsigal, line 29: I would say _service provider class_ to distinguish from JAX-RS providers.

ronsigal:
Section 4.1.1 "Automatic Discovery" says, "A JAX-RS implementation that supports automatic discovery of classes MUST process only those classes that are annotated with @Provider." How does that interact with the Extension mechanism? I would suggest that the discussion about Extensions be moved to the "Automatic Discovery" section,, where the interaction of the two mechanisms can be clarified.

jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
spericas: I agree

spericas:
> I'm fine with the current version of the pull request.
> 
> Regarding `@ContstrainedTo`: I agree that this annotation should be honored also for extensions. But I don't think that this needs to be mentioned in the new section. Especially because `@ContstrainedTo` is only mentioned in the "Summary of Annotations" appendix and nowhere else in the spec document.

 I'm sorry, but this should be the reason _why_ we should mention it and be explicit, not _why not_. It doesn't have good coverage beyond the Javadocs. To say that because we don't have good coverage now, we should continue to do so makes no sense to me. OTOH, we could do this in another PR, but it seemed relevant to this one.

mkarg:
> Section 4.1.1 "Automatic Discovery" says, "A JAX-RS implementation that supports automatic discovery of classes MUST process only those classes that are annotated with @Provider." How does that interact with the Extension mechanism? I would suggest that the discussion about Extensions be moved to the "Automatic Discovery" section, where the interaction of the two mechanisms can be clarified.

Both mechanisms are *alternatives* and that is exactly told already in the spec at exactly the location where providers are defined, so for me, it is already *unambiguous*. What else do you expect from a *spec* (it is not a tutorial)?

jaxrs-spec/src/main/asciidoc/chapters/providers/_providers.adoc
mkarg: Actually those *are* JAX-RS providers, so what problem do you see here?

ronsigal:
re: " it is already unambiguous"

When Section 4.1.1 says, "A JAX-RS implementation that supports automatic discovery of classes MUST process only those classes that are annotated with @Provider.", it doesn't explicitly restrict "MUST process only those classes" just to automatic discovery. To me, it could be interpreted to apply generally to the processing of providers. Anyway, when I read it, I thought, "Wait, does that apply to service loading?" I guess I would say that it's **potentially** ambiguous.

chkal:
I agree with @ronsigal that section 4.1.1 should be adjusted as well. Actually the text at the beginning of section 4 is just some kind of introduction. The discovery mechanisms themselves should perhaps go into 4.1.1. 

Maybe we should just add a minor change to the introduction text by changing it from:

> All application-supplied providers implement interfaces in the JAX-RS API and MAY be annotated with `@Provider` for automatic discovery purposes.

to:

> All application-supplied providers implement interfaces in the JAX-RS API and MAY be annotated with `@Provider` **or implement `Extension`** for automatic discovery purposes.

And then put the new text by @mkarg to section 4.1.1 instead and also adjust the text @ronsigal quoted. Actually section 4.1.1 is VERY short at the moment and I like the idea to enhance it.

mkarg:
In fact, it is not *or* but *and*: A provider can implement `Extension` *and* be annotated by `@Provider` at the same time.

chkal:
Agreed. Such aspects should be described as well.

mkarg:
@ronsigal @chkal Please send a PR to my fork that implements the changes you want.

chkal:
> @ronsigal @chkal Please send a PR to my fork that implements the changes you want.

Sorry for the delayed response! Sure, I hope to find some time tomorrow to create a corresponding PR.

chkal:
Sorry for the delay. I just opened this PR: https://github.com/mkarg/jaxrs-api/pull/7

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
```diff
@@ -29,9 +29,20 @@ all of the lifecycle options provided by that framework.
...
+[[extensions]]
+Providers MAY implement the `Extension` interface to declare it as an
+extension of the runtime.
+Compatible products MUST automatically register a class in
+`jakarta.ws.rs.core.Configuration` runtime configurations if the
+class is a _provider class_ for the `Extension` _service type_ according
+to the rules set out in the description of `ServiceLoader`.
+
+A JAX-RS implementation that supports automatic discovery of classes MUST
+process only those classes that are annotated with `@Provider` or are
```
mkarg, line 43: @chkal You added an *ONLY* restriction here, which in fact would prevent Jersey to further process its *proprietary* annotation. Is this really what you want? And is this agreed by @jansupol ?

jansupol: I believe two principles are mixed here together. The automatic discovery, i.e. classpath scanning, and JDK service loading mechanism. The chapter is about Automatic Discovery and it should not be mixed with the Extensions, which are not loaded by a classpath scanning. If the meaning would be as @mkarg suggests, any service loaded by an implementation via ServiceLoader would be in trouble. 

A new chapter regarding Extensions without a connection to the Automatic Discovery seems to be more appropriate, I.e. keep section 4.1.1 Autodiscovery as is, and add
```
==== Extensions

Providers MAY implement the `Extension` interface to declare it as an extension of the runtime. 
Compatible products MUST automatically register a class in `jakarta.ws.rs.core.Configuration` 
runtime configurations if the class is a _provider class_ for the `Extension` _service type_ according to 
the rules set out in the description of `ServiceLoader`.
```

Still, I believe the Extensions are going to be a source of trouble where one app will affect the second with their Extension and it should be possible to opt out the Extensions by an Application setting. The similar `AutoDiscoverable` feature in Jersey was meant for Jersey modules, and for the integration layers, but not for the end users.

jansupol: Even better yet, the Auto Discovery needs to be turned on (either by a missing Application or by returning an empty collection by both `Application.getClasses` and `Application.getSingletons`), and similarly, I'd prefer the Extension support to be turned on before it starts working, either by a missing Application or by annotating the existing Application with a new annotation.

mkarg: Opting out: We already discussed this and agreed that this a new opt-out property would be good, but this has nothing to this with *this* discussion, as it is solely about documenting the feature *as it already exists* in `master` and that everybody wanting new feature must file his own PR.

chkal: > @chkal You added an _ONLY_ restriction here, which in fact would prevent Jersey to further process its _proprietary_ annotation. Is this really what you want? And is this agreed by @jansupol ?

Good point. The `only` restriction was already present before, but just referred to `@Provider`. So I could argue that Jersey (and most other implementations) are already violating this part of the spec with their proprietary mechanism of registering providers. No?

> I believe two principles are mixed here together. The automatic discovery, i.e. classpath scanning, and JDK service loading mechanism.

IMO using `Configurable#register()` is *manual* and everything else (`@Provider` and `Extension`) is *automatic* discovery, as it doesn't require explicit calls to `register`.

However, I would also be fine with moving the `Extension` part to a separate section if everybody agrees that this would be better.

mkarg:
Guys, time is running out... I would kindly like to ask you to read this https://github.com/eclipse-ee4j/jaxrs-api/pull/947#issuecomment-808917797 comment and then go on with the code as-is: Let's document extensions, and merge the docs as-is, and **after that** continue **in separate threads** with the "alternative" idea of having auto-loaded application classes, and opting-out of auto-loading modules (if this still is needed after you all understood that it is a bad idea, as it would render Jersey modules unuseable).

anthonyvdotbe:
@mkarg #947 is what its title says: an alternative to #933/#751.
My main concern with the current solution is exemplified by the following scenario: say I have an EAR which contains: (1) multiple REST applications, all with an explicit `Application` subclass, and (2) a utility library (possibly as a (transitive) dependency) with all kinds of `Feature` & other JAX-RS interface implementations, some of which are annotated with `Extension`. Now when this EAR is deployed on a server with JAX-RS 3.0, the `Extension`-annotated implementations aren't registered. However, when the server is upgraded to JAX-RS 3.1, suddenly the `Extension`-annotated implementations get registered. So as I see it: (1) this solution is a backward-incompatible change, and (2) there's no way to opt-out of it.
It also seems inflexible to me: what if I have a `Feature` which some people want to deploy server-wide for all applications, while others want to embed it in their WAR for use by just their `Application`?
On a final note, even with this PR, I believe the solution is still underspecified.

(By the way, I think it would be a good rule to only allow approving PRs that contain both the required spec and API changes. Sometimes the spec changes raise concerns about the API changes & vice versa. When there's separate PRs, as in this case, it seems there's little to no room for altering the already-merged API/spec changes, even though that might be the right thing to do.)

mkarg:
As I said, there is a misunderstanding. `Extension` will not be part of an EAR or WAR, just like nobody would put a Jersey module into an EAR or WAR.

anthonyvdotbe:
But, unless I'm missing something, there's nothing that prevents it from (accidentally) happening either. In fact, in this regard it's a good thing that the interface for "runtime plugins" is currently implementation-specific: if the application server guarantees code portability (which it should) by strictly adhering to the Jakarta EE-specified behavior, it won't load runtime-specific plugins from within EARs/WARs.
Also, take e.g. [this example](https://github.com/eclipse-ee4j/jaxrs-api/pull/751#issuecomment-477711675): if I write a `Feature` which implements tracing, then some people will ask me to annotate it with `Extension`, so they can deploy it server-wide, while some people will ask me to do the opposite, so they can deploy it along with their application.

spericas:
> As I said, there is a misunderstanding. `Extension` will not be part of an EAR or WAR, just like nobody would put a Jersey module into an EAR or WAR.

You can't really prevent that, just like you can't really prevent someone from bundling a different implementation in a WAR/EAR.

chkal:
There are many comments in this PR and in #947, and it starts to get really difficult to follow all the thoughts. So I'll try to summarize my thoughts here.

In a nutshell the `Extension` interface mechanism and the alternative described in #947 are quite similar. Both rely on the `ServiceLoader` mechanism to load providers which is fine in my view. The main difference seems to be that #947 doesn't require an additional marker-like interface which seems like the simpler solution. But maybe I'm missing some downside of NOT having the `Extension` interface.

And yes, the `Extension` interface is already in `master` and is just lacking documentation in the spec itself. IIRC the original PR was created at a time we didn't have the spec documents in our repo yet. So we are in this weird state in which we should (of course) add the missing documentation in the spec but also have this (simpler?) alternative proposal on the table.

In my understanding both mechanisms share the same potential problems, which is that auto-loading of providers may be unexpected. However, the same downside applies to CDI extensions, `web-fragment.xml` files and others, so I'm not sure if this is really a downside in practice.

BTW: I see this mechanism and #947 primarily as a way for extension libraries deployed with the application to contribute providers to the runtime. So basically the primary use case is what @arjantijms describes [here](https://github.com/eclipse-ee4j/jaxrs-api/pull/947#issuecomment-789024808). And I also ran into a similar portability issue working on Eclipse Krazo.

If we want to move forward, it would be very valuable to have concrete pros and cons regarding the `Extension` mechanism vs #947.

mkarg:
Actually having an explicit marker interface is a big PRO of the existing solution, as it clearly defines an Jersey module as being a module of the runtime, in contrast to just some auto-loaded part of the application.

spericas:
@chkal Your summary is correct, they are very similar and there's no point in supporting both.

@mkarg I don't understand your comment or why _Jersey_ is relevant in this context. In general, the service loader mechanism in Java is used by libraries but nothing prevents applications to use it as well. Same for the proposal here.

anthonyvdotbe:
@chkal I'd propose to just consider/compare the actual proposed code changes, which are really small, rather than reading through all the comments (some of which are outdated/irrelevant/... anyway)

In short, I believe the advantages of #947 are that it: (1) is unambiguous, (2) doesn't require new API, (3) guarantees that an `Application` subclass retains full control, and as a consequence of that (4) is a backward compatible change.

mkarg:
> @chkal Your summary is correct, they are very similar and there's no point in supporting both.
> 
> @mkarg I don't understand your comment or why _Jersey_ is relevant in this context. In general, the service loader mechanism in Java is used by libraries but nothing prevents applications to use it as well. Same for the proposal here.

Jersey plays a role as its mechanism to load its modules (like JSON-B support for example) was the prototype for me when I designed `Extension`. What I wanted to reach when I proposed `Extension` was to get a cross-vendor module loading mechanism. If we not harm that target and let an application have control over what is loaded and what is not loaded, then this is definitively *not* what a I had in mind when I designed `Extension`.

spericas:
> > @chkal Your summary is correct, they are very similar and there's no point in supporting both.
> > @mkarg I don't understand your comment or why _Jersey_ is relevant in this context. In general, the service loader mechanism in Java is used by libraries but nothing prevents applications to use it as well. Same for the proposal here.
> 
> Jersey plays a role as its mechanism to load its modules (like JSON-B support for example) was the prototype for me when I designed `Extension`. What I wanted to reach when I proposed `Extension` was to get a cross-vendor module loading mechanism. If we not harm that target and let an application have control over what is loaded and what is not loaded, then this is definitively _not_ what a I had in mind when I designed `Extension`.

We seem to be returning back to the same points and going in circles. I thought we had established that nothing prevents an application to use _either_ proposal, so I don't understand why you keep bringing up this distinction. Even if there was a way to prevent that (don't know how) that's not necessarily a good thing.

mkarg:
> We seem to be returning back to the same points and going in circles. I thought we had established that nothing prevents an application to use _either_ proposal, so I don't understand why you keep bringing up this distinction. Even if there was a way to prevent that (don't know how) that's not necessarily a good thing.

I am bringing this up as the new proposal has no way to clearly mark a runtime module as a runtime module. In case a runtime needs to know this distinction (even simply for the case of logging "I loaded these runtime modules: ..." the new solution would not be sufficient. This has nothing to do with the problem of loading a module being part of an EAR, it has to do with the *intention* and *semantics* of `Extension`.

chkal:
> I am bringing this up as the new proposal has no way to clearly mark a runtime module as a runtime module. In case a runtime needs to know this distinction (even simply for the case of logging "I loaded these runtime modules: ..." the new solution would not be sufficient. This has nothing to do with the problem of loading a module being part of an EAR, it has to do with the _intention_ and _semantics_ of `Extension`.

As mentioned in #947 I saw the `Extension` interface primarily as a way to build portable libraries like Soteria, Mojarra, Payara JWT and Krazo. I and don't see a real difference between such libraries and a JAXB module for Jersey. They all contribute providers to the runtime. That's all according to my understanding.

spericas:
@mkarg Can we close this one?

mkarg:
Closing this MR as it had the intention to document a feature that meanwhile got replaced by https://github.com/eclipse-ee4j/jaxrs-api/pull/947.

PR #942, opened 2021-01-21
Add text markdown and text markdown type constants

mamadaliev:
Hello!
Could you add a new media type constant to this library?

```
text/markdown
```

Thanks.


mkarg:
I do not see a need to have this as a part of the standard. It is pretty simple to add this to the application itself by a single line of code.

PR #943, opened 2021-01-21
Add markdown text media type constant

mamadaliev:
Hello!
Could you add a new media type constant to this library?

```
text/markdown
```

Thanks.

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/MediaType.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/MediaType.java
@@ -160,6 +160,14 @@ public class MediaType {
      * A {@link MediaType} constant representing {@value #APPLICATION_JSON_PATCH_JSON} media type.
      */
     public static final MediaType APPLICATION_JSON_PATCH_JSON_TYPE = new MediaType("application", "json-patch+json");
+    /**
+     * A {@code String} constant representing {@value #TEXT_MARKDOWN} media type.
+     */
+    public final static String TEXT_MARKDOWN = "text/markdown";
+    /**
+     * A {@link MediaType} constant representing {@value #TEXT_MARKDOWN} media type.
+     */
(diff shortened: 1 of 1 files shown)


mkarg:
I do not see a need to have this as a part of the standard. It is pretty simple to add this to the application itself by a single line of code.

andymc12 (approved):
@mkarg are you opposed to the change (-1) or more just ambivalent (+0)?  Personally, I am not aware of any users requesting this constant (aside from present company), but MarkDown is fairly popular and could make a useful media type for RESTful resources.  I don't see any harm in adding the constant - and fwiw, there is an RFC for text/markdown at https://tools.ietf.org/html/rfc7763. +1 from me. Thanks for the suggestion @mamadaliev

jansupol:
For the sake of completeness, here is a [list of media types](https://www.iana.org/assignments/media-types/media-types.xhtml) and respective RFCs. Only a few are as popular as markdown. On the other hand, there are really many of them, to support them all. I'm +0.5 here :)

mkarg:
I am -1 for adding any popular media type to the specification. We should instead create a second JAR that is just an official library of extensions.

chkal:
I don't think that Markdown is served *directly* from a service very often without rendering it into HTML before.

+0 from my side.

mkarg (changes requested):
I am -1 for that change as it does *neither* solve anything that wouldn't be possible to solve by custom code in a single code line, *nor* is it related to REST, *nor* is this media frequently used by JAX-RS servers. I am +1 for JAX-RS project publishing an additional lib with commonly used extensions, that could provide this media type. So feel free to open another PR that introduces such a lib, and I will vote +1 for *that*.

spericas:
I'm +0 on this one, but definitely -1 to maintain a separate "library" for this purpose. Having another artifact to keep track as part of our process seems an unnecessary burden.

mkarg:
I am hereby closing this PR as it failed to reach the needed quorum.

PR #946, opened 2021-02-09
[944] Clarify HttpHeaders.getRequestHeader(String name) javadoc if header doesn't exist

### LINKED ISSUE #944: Clarification and test of `HttpHeaders.getRequestHeader(String)` when header does not exist
Per the [javadocs](https://jakarta.ee/specifications/restful-ws/3.0/apidocs/jakarta/ws/rs/core/httpheaders#getRequestHeader(java.lang.String)) for `HttpHeaders.getRequestHeader(String)`:
> List<String> getRequestHeader​(String name)
Get the values of a HTTP request header. The returned List is read-only. This is a shortcut for getRequestHeaders().get(name).
...
Returns:
a read-only list of header values.
...

There is a slight conflict in the text.  On the one hand, the description and "Returns:" section say that this method must return a read-only list.  On the other hand, the description says that this method is a shortcut for `getRequestHeaders().get(name)` which would return null if the specified key does not exist in the map.

I mention this because a user testing their app on CXF was checking for null, and then their app broke when it was getting an empty list in RESTEasy. For the short term, I've advised the user to check for null and for an empty list, but ideally this would be clarified and tested in the TCK.

IMO, either option (returning null or returning an empty list) is fine (as long as it is one or the other), but I tend to prefer returning null if the header key does not. If we wanted to be more Java 8-friendly, we could return an `Optional<List<String>>`, but that would be a breaking change...

I'll plan to code up PRs with the change to the javadoc and TCK.  If anybody has strong opinions about returning an empty list instead of null, please let me know.  Thanks!

andymc12:
Fixes #944 - updates the javadoc to indicate that `getRequestHeader(name)` will return null if the header does not exist.

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/HttpHeaders.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/HttpHeaders.java
@@ -34,11 +34,11 @@ public interface HttpHeaders {
 
     /**
      * Get the values of a HTTP request header if the header exists on the current request. The returned value will be
-     * a read-only List if the specified header exists or null if it does not. This is a shortcut for
+     * a read-only List if the specified header exists or {@code null} if it does not. This is a shortcut for
      * {@code getRequestHeaders().get(name)}.
      *
      * @param name the header name, case insensitive.
-     * @return a read-only list of header values if the specified header exists, otherwise null.
+     * @return a read-only list of header values if the specified header exists, otherwise {@code null}.
      * @throws java.lang.IllegalStateException if called outside the scope of a request.
      */
     public List<String> getRequestHeader(String name);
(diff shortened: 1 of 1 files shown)


spericas (approved):

andymc12:
@chkal @mkarg @arjantijms @asoldano are you available to review this change? (it's pretty small :) )

chkal (approved):
Sorry for the delay. LGTM.

But maybe we should use `{@code null}` instead. AFAIK this is how we do it in other cases as well (see [here](https://github.com/eclipse-ee4j/jaxrs-api/blob/2c481255003588caabd7bf40074538a7583d86ba/jaxrs-api/src/main/java/jakarta/ws/rs/core/Configurable.java#L154) for example). But that's not a blocker.

andymc12:
@chkal 
> But maybe we should use {@code null} instead. AFAIK this is how we do it in other cases as well (see here for example). But that's not a blocker.

I agree.  It's good to be consistent.  I've updated the changes to use `{@code null}` instead.  Would you and @spericas re-approve?

Thanks!

chkal (approved):

mkarg (approved):

spericas (approved):

PR #947, opened 2021-02-28
Alternative proposal for JAX-RS Extensions

anthonyvdotbe:
This PR reverts #751 and replaces #933 by doing the following instead:
* make this a spec-only change (as [already mentioned](https://github.com/eclipse-ee4j/jaxrs-api/pull/751#issuecomment-477789591), the `Extension` interface isn't indispensable)
* only require automatic registration of `Feature`s (which is basically what Jersey does with its `ForcedAutoDiscoverable`)
* explicitly specify the `ServiceLoader` invocations, to avoid any ambiguity & prevent issues in environments such as Jakarta EE (the goal being to avoid undesired automatic registration of `Feature`s in EARs/WARs)


mkarg (changes requested):
-1 as it won't be compatible with Java  SE 8

arjantijms:
+1 ServiceLoader [should work on Java SE 8,](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) even though I think we should move before long to Java SE 11 as a minimum.

mkarg:
> +1 ServiceLoader [should work on Java SE 8,](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) even though I think we should move before long to Java SE 11 as a minimum.

Certainly ServiceLoader will work, but not the proposal of only using it when using Module Layers.

anthonyvdotbe:
> Certainly ServiceLoader will work, but not the proposal of only using it when using Module Layers.

That's not what's proposed. On Java SE 8 the `if` clause is false by definition and so the `else` clause is used instead:
> else, all implementations discovered by `ServiceLoader.load​(Feature.class, Feature.class.getClassLoader())`

arjantijms:
I don't think it's proposed that it's only useable via module layers. It now says:

```
if the JAX-RS runtime was loaded in a `java.lang.ModuleLayer` L, all implementations discovered by `java.util.ServiceLoader.load(L, Feature.class)`
else, all implementations discovered by `ServiceLoader.load​(Feature.class, Feature.class.getClassLoader())`
```

The else case is important there.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
```diff
@@ -33,6 +33,13 @@ discover provider classes via mechanisms such as class scanning. A
 JAX-RS implementation that supports automatic discovery of classes MUST
 process only those classes that are annotated with `@Provider`.
 
+[[extensions]]
+==== Extensions
```
jansupol, line 37: I would not call these `Extensions`. The name would be easily messed up with CDI Extensions, especially in 4.0. I believe these are `Services`.

anthonyvdotbe: Good point, addressed in 30877eea03892c6fb5f53dee27a79ad9c7ce89f0

spericas:
I like this proposal, seems better integrated with our API and does not need an extra interface. Any reason why we shouldn't support this auto-loading mechanism for `DynamicFeature` as well.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
```diff
@@ -33,6 +33,13 @@ discover provider classes via mechanisms such as class scanning. A
 JAX-RS implementation that supports automatic discovery of classes MUST
 process only those classes that are annotated with `@Provider`.
 
+[[services]]
+==== Services
+An implementation MUST automatically register the following set of `jakarta.ws.rs.core.Feature` providers in `jakarta.ws.rs.core.Configuration` runtime configurations:
+
+* if the JAX-RS runtime was loaded in a `java.lang.ModuleLayer` L, all providers discovered by `java.util.ServiceLoader.load(L, Feature.class)`
+* else, all providers discovered by `ServiceLoader.load​(Feature.class, Feature.class.getClassLoader())`
```
andymc12, line 41: I think we would not want to use the Feature's classloader.  Instead, we should follow the pattern we use in the `FactoryFinder` classes where we first try the thread context classloader, then `Feature.class.getClassLoader()`, and then possibly a `jaxrs.properties` file. 

The trouble with just using the `Feature` class is that multi-app containers like Liberty and Wildfly might allow users to add their extensions/services/whatever-we-call-it to their app's classpath, but then it would not be found by the Feature's classloader.

andymc12 (changes requested):
Overall, I like this idea - using the ServiceLoader is a very common pattern and it works in Java 8 as well as with JPMS, but I think the text in the spec should indicate that it follows the FactoryFinder pattern for checking the `ServiceLoader` using the thread context classloader rather than (or before) the `Feature.class.getClassLoader()`.  Thanks!

mkarg:
> I like this proposal, seems better integrated with our API and does not need an extra interface. Any reason why we shouldn't support this auto-loading mechanism for `DynamicFeature` as well.

What I do not like with this proposal actually *is* that one cannot selectively decide that what an extension is (i. e. *shall* be autoloaded) and what *not* (i. e. *shall not*) get autoloaded. The idea of `Extension` explicitly was that it clearly provides this separation.

spericas:
> > I like this proposal, seems better integrated with our API and does not need an extra interface. Any reason why we shouldn't support this auto-loading mechanism for `DynamicFeature` as well.
> 
> What I do not like with this proposal actually _is_ that one cannot selectively decide that what an extension is (i. e. _shall_ be autoloaded) and what _not_ (i. e. _shall not_) get autoloaded. The idea of `Extension` explicitly was that it clearly provides this separation.

Yes, you can. As I understand it, you need specify services files for the auto-loading to kick in.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
spericas: That's a good point by @andymc12.

anthonyvdotbe: The rationale for just using `Feature`'s `ClassLoader` is to prevent undesired auto-loading of `Feature`s in EARs/WARs.

As I understood it, this feature is about extending the JAX-RS runtime in a portable way & without involvement of the application developer. So I'd argue `Feature`s must not be auto-loaded from developer-owned artifacts such as EARs/WARs.

Currently, developers have total control over what is and isn't loaded from their artifacts via `Application`. And if desired, they can use `ServiceLoader` to find & register additional `Feature`s themselves using the thread's context `ClassLoader`.

By requiring auto-loading via the thread's context `ClassLoader`, developers would loose a certain degree of control and it might lead to subtle, hard-to-debug issues (see also [this comment](https://github.com/eclipse-ee4j/jaxrs-api/pull/933#issuecomment-755449932) e.g.). In this case, developers would need a way to opt-out of this behavior, which would complicate the feature.

What do you think?

andymc12: >As I understood it, this feature is about extending the JAX-RS runtime in a portable way & without involvement of the application developer. So I'd argue Features must not be auto-loaded from developer-owned artifacts such as EARs/WARs.

Yes and no. In an embedded application scenario (like using Quarkus or a Java SE environment), everything would be on the same classpath so the JAX-RS implementation code wouldn't know whether a particular extension feature came from the application developer or as a product extension - it's just on the classpath.  But then if a developer did package their extension with the app and then ran it in multi-app container (like Wildfly/Liberty/etc.) they would see different results - so in effect, this approach isn't portable.

> Currently, developers have total control over what is and isn't loaded from their artifacts via Application. And if desired, they can use ServiceLoader to find & register additional Features themselves using the thread's context ClassLoader.

True, but that is not intuitive.  I don't think I've ever seen an example where a user used the ServiceLoader in the Application subclass.  Further, we are deprecating and (eventually) removing the `getSingletons` method from `Application`, which will make it that much harder for users to add providers via the ServiceLoader pattern - they would need to use `getClasses` which means using the ServiceLoader to return an instance of the provider, and the return the instance's class.  

> By requiring auto-loading via the thread's context ClassLoader, developers would loose a certain degree of control and it might lead to subtle, hard-to-debug issues (see also this comment e.g.). In this case, developers would need a way to opt-out of this behavior, which would complicate the feature.

I think this is a problem with the idea of extensions in general - regardless of which classloader is used. For example, a developer will run and (hopefully) pass tests on their application before turning it over to production. But if the production admins decides to add some new monitoring extension to the container, it might change the behavior of the app in unexpected ways.  I would think that if the extension were part of the application, it would actually make it easier to find and fix the behavior differences earlier in the development lifecycle.

But even aside from embedding the extensions in the application archives, these multi-app servers also allow apps to use shared libraries which is the deployment target of these extensions - and the shared libraries are accessed via the thread context classloader.

arjantijms: @anthonyvdotbe Are you referring to the situation where there's a jar in (say) an application server with a `META-INF/jakarta.ws.rs.core.Feature` that lists "foo.bar.Kaz", that there might be an EAR or WAR out there that contains a class that's by coincidence also called "foo.bar.Kaz", and that the REST runtime then loads the "foo.bar.Kaz" from the application archive instead of the application server archive?

arjantijms:
> Yes, you can. As I understand it, you need specify services files for the auto-loading to kick in.

Indeed, a `Feature` (dynamic or not) would never be loaded based on it being a `Feature` instance. Its fully qualified classname would have to appear explicitly in a `META-INF/jakarta.ws.rs.core.Feature` file.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
anthonyvdotbe: @arjantijms I'm referring to the concerns raised by @jansupol [here](https://github.com/eclipse-ee4j/jaxrs-api/pull/933#issuecomment-755449932) w.r.t. an EAR containing multiple modules developed by independent teams, where the module of one team could interfere with the modules of other teams. Or even just: a WAR, where one of the transitive dependencies could at any point add a `Feature` that has undesired (or even malicious) effects, even when an `Application` subclass is present.

anthonyvdotbe:
tl;dr I believe we should simply revert #751 and leave things as-is

As I see it, there's 2 user stories here:
1. as a systems integrator, I want to extend a JAX-RS implementation's set of pre-packaged providers in a portable way, so that [...]
2. as a Java SE application developer, I want to run my JAX-RS application without having to specify an `Application` subclass (i.e. I want some kind of automatic discovery), so that [...]

At first sight, it seems both user stories can be solved at once by employing a `ServiceLoader`-based mechanism.

However, on the one hand, for (1):
* applications must be oblivious of, and unable to interfere with, this mechanism
* there is no need for a way to opt-out
* supporting `Feature` is sufficient

On the other hand, for (2):
* applications are the intended target of this mechanism
* there must be a way to opt-out. Moreover, the presence of an `Application` subclass must cause the mechanism to be disabled
* every component interface should be supported

One way out, might be to specify the mechanism like follows:
> On runtime initialization, the implementation discovers and registers all providers of X, where X is [...], as follows:

```java
if(!ServiceLoader.load(Application.class).iterator().hasNext()) {
    // register all of ServiceLoader.load(X.class);
}
if(!ServiceLoader.load(Application.class, X.class.getClassLoader()).iterator().hasNext()) {
    // register all of ServiceLoader.load(X.class, X.class.getClassLoader());
}
```

The purpose of the `On runtime initialization` clause, is to guarantee that this discovery is done just once. So in a Jakarta EE environment, this must be equivalent to doing it on server startup, rather than once for each deployed application.

So while I think the above mechanism would actually work as desired in all environments, I believe we should simply revert #751 and leave things as-is for the following reasons:

W.r.t. (1):
* the spec already says `the integration of pre-packaged providers into the JAX-RS runtime is implementation dependent`, so I'd argue that extending the set of pre-packaged providers should, by definition, be implementation-dependent as well
* in most cases (e.g. in the context of an application server), the JAX-RS runtime is known and unlikely to change (i.e. there's no actual need for portability)
* the actual provider implementations are generic/reusable anyway. Only the means of having the providers automatically registered by the runtime are implementation-dependent
* this can be standardized across implementations, without burdening the spec with this (in other words: I believe "all implementations have this integration point" is, on its own, insufficient motivation for standardization)

W.r.t. (2):
* once CDI becomes mandatory, CDI can/should be used
* in the meantime, developers can already use `ServiceLoader` themselves in their `Application` subclass. When using Java SE 9+, this is trivial to do in `getClasses` without having to instantiate providers: `ServiceLoader.load(X.class).stream().map(Provider::type)`.

In summary, I propose to simply remove all commits but 80106d16a11ee681af9881343a7c9d73e437c36f from my PR and merge like that. (Alternatively, in case there's consensus that this feature must be part of 3.1, I hold that the chosen solution must allow for an `Application` subclass to retain complete control over what is and isn't registered. In this case, I propose to adopt the mechanism outlined above.)

arjantijms:
> in most cases (e.g. in the context of an application server), the JAX-RS runtime is known and unlikely to change (i.e. there's no actual need for portability)

You're forgetting the case of portable libraries like Soteria, Mojarra, Payara JWT, etc that are supposed to be integrated with different stacks. Having no portable APIs for this integration makes things quite complicated for these.

Look at this one for example: 

https://github.com/payara/Payara/tree/master/appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth

I was able to code this up using only portable APIs. Almost. CDI and Security provide the portable APIs, Jakarta Rest does not. So this one needed to use Jersey specific code: https://github.com/payara/Payara/blob/master/appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/jaxrs/RolesAllowedAutoDiscoverable.java

With a portable API in place, such libs do not need to be tied to a specific application server implementation. Better yet, with such APIs application servers can far more easily support multiple implementations of components, which is the issue I'm facing with Piranha Cloud.

anthonyvdotbe:
> You're forgetting the case of portable libraries like Soteria, Mojarra, Payara JWT, etc that are supposed to be integrated with different stacks. Having no portable APIs for this integration makes things quite complicated for these.

Point taken, I hadn't considered that case. What do you think of the alternative solution I mentioned? Or waiting for mandated CDI?

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
spericas: I thought the main use case for these extensions is precisely that in which the features are _not_ part of an application (much like CDI extensions). If it is part of an application, it can be added without any auto-loading mechanism. The corresponding facility in Jersey is mostly used by frameworks that sit on top of Jakarta REST. Thus, this needs to follow the `FactoryFinder` approach as pointed out by @andymc12 to make sure it runs predictably in all environments.

I also think the proposal should handle `DynamicFeature` or explain why it is left out at least.

spericas:
> > You're forgetting the case of portable libraries like Soteria, Mojarra, Payara JWT, etc that are supposed to be integrated with different stacks. Having no portable APIs for this integration makes things quite complicated for these.
> 
> Point taken, I hadn't considered that case. What do you think of the alternative solution I mentioned? Or waiting for mandated CDI?

In my view, this is the primary use case for this auto-loading mechanism --everything else is gravy ;)

anthonyvdotbe:
Ok, I've had another go at it: `DynamicFeature` is now included as well. It deviates from the `FactoryFinder` algorithm in a few ways, but I think that's logical:
* it has a condition on the equivalent `Application` invocation not finding anything, in order to avoid undesired auto-loading in places where an `Application` is in control
* it always searches with both `ClassLoader`s, unlike `FactoryFinder` which returns the first result, to ensure predictability
* it doesn't consider `jaxrs.properties` since, if both `ClassLoader`s used in the previous steps already discoverd an `Application` provider, there's no obvious `ClassLoader` to use for loading the specified providers

mkarg:
Thinking about this PR I have the feeling that I need to better explain what `Extension` actually is good for. I do not see it in competition with this PR.

Jersey, and possibly more JAX-RS implementations, are modular systems. There is a core implementation that does the boilerplate plumbing, and there are (lots!) of separate modules ("Jersey plugins") that you can simply put on the Class-Path to enable them. These modules are not part of the application, they are part of the runtime! For example, support for JAXB or JSONB are not in the Jersey core, they are separate modules. If you want to have a Jersey runtime that supports JAXB or JSONB, you put that modules on the Class-Path and that's it. It has *nothing* to do with auto-loading application components!

What my target with the (actually already accepted and found in `master` branch) interface `Extension` was *solely* was to allow writing cross-vendor runtime modules. This means, take the JAXB or JSON modules for Jersey, remove their current proprietary marker interface, and add the cross-vendor `Extension` interface. So the target is: Be able to *extend* any JAX-RS runtime in a cross-vendor way. That's why I called it `Extension`, that's why I wanted it to be enabled by default.

In the current discussion about the *documentation* of `Extension` and about *this* "alternative" we need to really care about this original intention, to not turn the idea into something different - in particular not to turn it into an idea how to auto-load application components, and in particular not to drop the original idea of loading *runtime plugins*.

Having said that, looking at the pace we do, I urge you to separate the idea of `Extension` (i. e. cross-vendor runtime plugins) from the idea of auto-loading application components (which seems to be the idea of *this* PR). Thanks.

spericas:
> Thinking about this PR I have the feeling that I need to better explain what `Extension` actually is good for. I do not see it in competition with this PR.
> 
> Jersey, and possibly more JAX-RS implementations, are modular systems. There is a core implementation that does the boilerplate plumbing, and there are (lots!) of separate modules ("Jersey plugins") that you can simply put on the Class-Path to enable them. These modules are not part of the application, they are part of the runtime! For example, support for JAXB or JSONB are not in the Jersey core, they are separate modules. If you want to have a Jersey runtime that supports JAXB or JSONB, you put that modules on the Class-Path and that's it. It has _nothing_ to do with auto-loading application components!
> 
> What my target with the (actually already accepted and found in `master` branch) interface `Extension` was _solely_ was to allow writing cross-vendor runtime modules. This means, take the JAXB or JSON modules for Jersey, remove their current proprietary marker interface, and add the cross-vendor `Extension` interface. So the target is: Be able to _extend_ any JAX-RS runtime in a cross-vendor way. That's why I called it `Extension`, that's why I wanted it to be enabled by default.
> 
> In the current discussion about the _documentation_ of `Extension` and about _this_ "alternative" we need to really care about this original intention, to not turn the idea into something different - in particular not to turn it into an idea how to auto-load application components, and in particular not to drop the original idea of loading _runtime plugins_.
> 
> Having said that, looking at the pace we do, I urge you to separate the idea of `Extension` (i. e. cross-vendor runtime plugins) from the idea of auto-loading application components (which seems to be the idea of _this_ PR). Thanks.

I really don't understand your point. Could you explain why the proposal in this PR does not allow the creation of what you called _cross-vendor runtime modules_. This auto-loading mechanism can be used for both applications and vendor modules, but it is much more useful for the latter since these vendor modules don't have full control of the "flow". 

The proposal in this PR is superior to the one we had agreed on before, and we should always strive for the better.

spericas (approved):

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
```diff
@@ -33,6 +33,14 @@ discover provider classes via mechanisms such as class scanning. A
 JAX-RS implementation that supports automatic discovery of classes MUST
 process only those classes that are annotated with `@Provider`.
 
+[[services]]
+==== Services
+On runtime initialization, an implementation MUST discover and register certain service providers in `jakarta.ws.rs.core.Configuration` runtime configurations.
+For each service X, where X is `jakarta.ws.rs.core.Feature` or `jakarta.ws.rs.container.DynamicFeature`, the set of providers to be registered MUST be determined as follows:
+
+* first, if `ServiceLoader.load(Application.class)` returns no results, register all providers discovered with `ServiceLoader.load(X.class)`
+* additionally, if `ServiceLoader.load(Application.class, X.class.getClassLoader())` returns no results, register all providers discovered with `ServiceLoader.load(X.class, X.class.getClassLoader())`
```
andymc12, line 42: We should not use the `ServiceLoader` to determine the existence of an `Application` subclass.  This would only work if the JAX-RS archive contained a `META-INF/services/jakarta.ws.rs.core.Application` file that specified a subclass (or if there was a Java 11+ module that specified the application subclass).  Instead, I think this line should just mention the existence of an `Application` subclass - but it should also specify that the services should be registered if `Application`'s `getClasses` and `getSingletons` methods return null or empty sets. Many JAX-RS apps contains an `Application` subclass solely for the purpose of specifying an `@ApplicationPath` or providing additional properties, etc.  

Maybe we could change these lines to something like:
```suggestion
If an `Application` subclass exists in the application, and that subclass returns a non-empty collection for either the `getClasses` or `getSingletons` method, then the implementation MUST NOT register any providers specified via the service loader mechanism. If no `Application` subclass exists, or if the subclass exists but does not specify the set of classes or singletons to use for the application, then the implementation MUST register all providers discovered with `ServiceLoader.load(X.class)`.
```

andymc12 (changes requested):
Thanks for the changes so far.  I added a concern/suggestion inline.  It also might be nice to have more background text in the spec document that explains the use case for these service extensions - and maybe an example too.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
anthonyvdotbe: Thanks, those are good points. I've added a note about what it's for. The spec itself is updated in a slightly different way as suggested though, to allow for the case
> when essentially portable features are provided via jar files that are put in an internal server path (e.g. glassfish/modules in Payara).

dixit [arjantijms](https://github.com/eclipse-ee4j/jaxrs-api/pull/751#issuecomment-477816397)

What do you think of this version?

spericas: @andymc12 I'm a bit confused by your last suggestion (and the new changes). If my application depends on REST library "Foo" and my application includes an `Application` subclass that returns non-empty collections from at least one of its method, how can library "Foo" extend the Jakarta REST runtime to function properly? I may have missed some earlier discussion about this.

andymc12: @spericas My thinking was that if the `Application` subclass returns a non-empty collection of providers, then we would not want to add any additional providers - even those that are part of the extension JARs.  The original text excluded the providers in the extension JARs if an `Application` subclass was present - so my comment was that some apps will contain an `Application` subclass but only to provide the `@ApplicationPath` and not to specify a strict collection of providers - in those cases, I think we want the providers in the extension JARs to be auto-registered.

chkal: I'm not convinced that implementing `getClasses` or `getSingletons` should disable extension discovery for the classloader of the `Application` subclass. 

And to be honest: I'm not sure if we really need a way to opt out extensions from beeing processed at all. If an extension is on your classpath, why not enabing it? If somebody puts an extension into an location where it is visible to all applications deployed to an app server or puts it as a library into an EAR where it is visible to all contained modules, it is hopefully because of some good reason. Also, the same potential "problem" exists for CDI extensions as well. At least I don't know of any mechanism to opt out a CDI extension which is on the classpath of the app.

In my optionion it should be the responsibility of the extension developer to provide way to disable the extension although it is visible to a JAX-RS application. And this is actually super easy to do:

```java
public class MyExtension implements Feature {

  @Override
  public boolean configure( FeatureContext context ) {
    if( context.getConfiguration().getProperty( "foo.bar.disable" ) != null ) {
      context.register( MyProvider1.class );
      context.register( MyProvider2.class );
      return true;
    }
    return false;
  }

}
```

So IMO extensions should always be processes, whether or not `getClasses` or `getSingletons` is implemented.

mkarg: Look at this case: An application explicitly implements `getClasses` to provide some custom filter. It *only* declares that filter, as it *only* provides this filter. The application it depends on JAXB, which so far was mandatory on JAX-RS. So it *does not* register it explicitly, but just relies on it "being there somehow". JAXB is not mandatory anymore, so the depoloyer (not the application programmer) adds an `Extension` module which shall get auto-loaded to provide it. This case makes clear, that it **must not** be up to the **application** to explicitly provide JAXB, as it is the **deployer's** choice. Also the case makes clear, that implementing `getClasses` **must not** disable auto-loading of the extension, as it would effectively **break** the application. So am I strictly **pro `Extension`**, strictly **-1 for disabling auto-loading of `Extension`, and +-0 for having some other opt-out mechanism.

mkarg: (Maybe we should start discussing *separate* Classloaders, one for application-provided providers, and one for bundled *`Extension`s*?

anthonyvdotbe: Suppose we have `foo.jar`, which contains `Feature` X as a service. Then with this PR in its current state: if `foo.jar` is deployed on an internal server path (e.g. `glassfish/modules` in Payara), it will be picked up by the JAX-RS implementation. This is currently not possible to do in a portable way and, as I understand it, the main use case. If `foo.jar` is part of a WAR, behavior matches the [existing discovery mechanism](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#servlet) and therefore guarantees backward compatibility.

@mkarg I didn't follow the history there, but the current construction w.r.t. JAXB is unconventional to say the least. Anyway, since the related API is deprecated, I'd expect for the application developer to migrate away from it, and for the deployer to add any missing JARs on an internal server path in the meantime

mkarg: There is nothing unconventional about this use of JAXB. It is how Jersey does it, and Jersey is anything but unconventional.

anthonyvdotbe: (Off-topic: I was referring to the fact that the JAXB clause in the module descriptor is `requires static`, even though it's still depended on by exported API)

mkarg: JAXB was just an arbitrary example and was not meant literally. Just pick *any* Jersey module to get the point.

anthonyvdotbe: I thought I got the point and replied to it, but let me try again: either an application developer develops against the spec, and bundles everything that isn't required by the spec in their application. Or they agree with their server administators/deployers about a set of additional features that will be provided to their application. In the latter case, if the JAX-RS implementation doesn't provide all of the agreed-upon features, the server administrator/deployer simply adds the required features on an internal server path, as is supported by this PR, and that is that.

mkarg: And exactly this scenario is what I try to solve with standardized `Extension` modules: You are not bound to the vendor-specific way of supplying the modules "in some internal server path" but can rely upon the JAX-RS standard, hence it will work exactly the same with all implementations, even in the cloud where you cannot talk to the administrator at all.

anthonyvdotbe: But in that case, the application developer shouldn't make any assumptions about what is and isn't available, and just develop against the spec, no? I must be missing something, so will you please give a concrete example, where it's neither acceptable for the application developer to bundle the `Feature`, nor possible for the deployer to add it to the server? (PS: I've edited the comment above to remove "internal server path", because it's irrelevant: every Jakarta EE server has its own way to extend it, but if you want to standardize that, you should do so at the Platform level, not for each specification individually)

mkarg: Think of an ISV that provides an EAR to the public. The ISV cannot know whether or not your application server provides support for JAXB (again, just an example), as there are cloud providers on the market that provide strict JAX-RS compliant service offerings, and others that provide several optional modules ontop. So the deployer (= customer of the ISV) can choose to either deploy to a cloud service officially supporting JAXB (= using a highly optimized and perfectly tuned implementation of that cloud provider), or he can add a slow-but-cross-vendor-runnable JAXB module into the EAR (just drop it in *there*) that *he* (not the appliction vendor) selected from the free market.

anthonyvdotbe: Thanks for the example. So IIUC, your point is that the deployer should be able to drop a JAR into the EAR, right? But if so, I still see several ways out. The ISV could e.g. ...
* use `ServiceLoader` inside the `Application` subclass to discover additional `Features` itself
* avoid adding an `Application` subclass that overrides `getClasses` and/or `getSingletons`
* provide different builds of its EAR for different deployment environments
* simply state `our EAR is only certified against these Jakarta EE implementations` (just as it could state that it requires Java SE 11 e.g.)
* provide Docker images where everything's already bundled up and the application server has already been optimized for their EAR

mkarg: * You are mixing application vendor and runtime vendor here. My target is to have a clear separation of the choice of the application vendor ("I need JAXB support") from the choice of the deployer (= customer) ("I fulfil this wish using this cross-vendor runtime module from Apache). The idea of Java EE ever was to have this clear separation, and my invention of the `Extension` interface reflects exactly this separation.
* It is (since ever) the choice of the application vendor whether to implementation `Application` or not. EARs are not provided by vendors, they are provided by deployers.
* Having different pre-bundled EARs completely undermines the idea of vendor-independence and WORA that is fundamental to JAX-RS: If we agree to that, we can drop the idea of common standard at all but just use native Jersey API. * JAX-RS is a standalone API which must work the same within or without Jakarta EE, as there are JAX-RS compliant cloud offerings which never will be Jakarta EE compliant by intention.
* Forcing people into usage of Docker mixes up vendor-decisions, deployer-decisions and provider-decisions and completely undermines the idea of a common Java-based standard, hence is out of scope of JAX-RS. We cannot answer the request to standardize the existing vendor-specific flags that identifies JAX-RS runtime modules by saying "use Docker" as there are people out there not using Docker at all.

To sum up, you proposed a lot of workarounds, but none of your answers solves the actual request: Standardizing the existing vendor-specific runtime module identifiers in a cross-vendor JAX-RS standard way.

anthonyvdotbe: * all I'm saying is: if desired, the ISV can accommodate deployers that want to inject additional `Features` in their EAR, despite the presence of the ISV's `Application` subclass
* in your example you say "an ISV that provides an EAR" and "the deployer (= customer of the ISV)", but now you say "EARs are not provided by vendors, they are provided by deployers."? Anyway, yes, it's up to the ISV to decide, and I gave an example of how to accommodate deployers in both situations
* what I meant is: an EAR with `jaxb.jar` (which has a service provider for `Feature` that provides JAXB support), and an EAR without `jaxb.jar`. The code in `jaxb.jar` itself would be pure JAX-RS
* as an ISV, I can state whatever I want, so I don't get your point
* I gave 4 alternative options, so I don't understand how that is forcing anything. Moreover, I don't get the point: in this scenario, the ISV sells a "solution", and is both application developer & deployer at the same time. And when building its Docker image, they can use the functionality provided by this PR to extend the server runtime in a standardized way, independently of their EAR

> To sum up, you proposed a lot of workarounds, but none of your answers solves the actual request: Standardizing the existing vendor-specific runtime module identifiers in a cross-vendor JAX-RS standard way.

Maybe none of my answers do, but I believe my PR does.

chkal: This discussion is getting a bit hard to follow, but I want to comment on a few aspects:

> > use ServiceLoader inside the Application subclass to discover additional Features itself

> You are mixing application vendor and runtime vendor here. My target is to have a clear separation of the choice of the application vendor ("I need JAXB support") from the choice of the deployer (= customer) ("I fulfil this wish using this cross-vendor runtime module from Apache")

My thoughts for this concrete example: As JAXB isn't required for Jakarta EE anymore, an ISV should not have JAXB as a requirement. I agree with @anthonyvdotbe that in this case the ISV should simply provide a fallback for environments where JAXB is not available, which should be easy by registering a provider with a low priority.

> EARs are not provided by vendors, they are provided by deployers.

IMO maybe in an ideal world, but not in the real world. 

> * Having different pre-bundled EARs completely undermines the idea of vendor-independence and WORA that is fundamental to JAX-RS

I guess the real problem is that with optional features WORA doesn't really work. And to be honest: If a deployer has to "repackage" an ISV application to add support for something like JAXB, this isn't WORA anymore.

I would also like to comment on this reply by @mkarg:

> Look at this case: An application explicitly implements getClasses to provide some custom filter. It only declares that filter, as it only provides this filter. The application it depends on JAXB, which so far was mandatory on JAX-RS. So it does not register it explicitly, but just relies on it "being there somehow". JAXB is not mandatory anymore, so the depoloyer (not the application programmer) adds an Extension module which shall get auto-loaded to provide it. This case makes clear, that it must not be up to the application to explicitly provide JAXB, as it is the deployer's choice. 

I don't agree to the *must not*. Both ways should be possible. Either the deployer adds JAXB support (by adding it to the EAR or whatever) OR the application could include JAXB support itself (either by adding some kind of fallback provider OR by registering a provider with higher priority that is always preferred.

> Also the case makes clear, that implementing getClasses must not disable auto-loading of the extension, as it would effectively break the application. 

Please note that according to the current proposal implementing `getClasses` will only disable auto-discovery of extensions in the application's classloader. So according to my understanding extensions installed in the app server and extensions provided via EARs are still loaded.

> So am I strictly pro Extension, strictly **-1 for disabling auto-loading of Extension, and +-0 for having some other opt-out mechanism.

I still don't understand the benefit of having this additional `Extension` interface TBH. And I still don't see a need for distinguishing between runtime extensions and other extensions. I still see the mechanism we are discussing here primarily for portable libraries like Soteria, Mojarra, Payara JWT and Krazo.

mkarg: In the real world, there *are* scenarios where application vendor, deployer, and vendor of the module are *distrinct* entities that solely trust on the fact that there is a **standard** way to identify and deploy runtime modules, and that way is part of the JAX-RS standard, not outside of it, not vendor-specific, not application-bound. The need for a separate interface is that in future we *might* want to have methods in that feature, allowing runtimes to call on these modules (like for initialization purposed etc). And yes, this *mostly* is needed for portable modules, certainly, as this is *the reason* **why I invented `Extension` interface**.

mkarg: @chkal Again, please note that "JAXB" was just an example, please replace it by "any runtime module". I already answered this several times.

chkal: > In the real world, there _are_ scenarios where application vendor, deployer, and vendor of the module are _distrinct_ entities that solely trust on the fact that there is a **standard** way to identify and deploy runtime modules, and that way is part of the JAX-RS standard, not outside of it, not vendor-specific, not application-bound.

Agreed. And if we decide to merge this PR, it should support this scenario. Although I still don't see a real difference between a "runtime module" and a "portable library". In my view it is essentially the same. But maybe I'm missing something.

> The need for a separate interface is that in future we _might_ want to have methods in that feature, allowing runtimes to call on these modules (like for initialization purposed etc). And yes, this _mostly_ is needed for portable modules, certainly, as this is _the reason_ **why I invented `Extension` interface**.

So you say there is no need for the `Extension` interface *today* and you would like to keep it, so we can add methods later? I agree that we *may* have a need for additional methods later on, but wouldn't it make more sense to add such methods to the `Feature`/`DynamicFeature` interfaces instead? Basically a `Feature` is a set of providers. And as mentioned before, I don't see a need to distinguish between "runtime modules" and "portable libraries" which are both just sets of providers. That's what I like about this proposal: An extension is just a `Feature` (or `DynamicFeature`) which is loaded automatically and registers a list of providers.

mkarg: Yes you are missing something indeed: Runtime modules might exist that provide implementations *not intended* for applications. For example, a runtime module might exist *in future* that implements parts of RuntimeDelegate (i. e. RuntimeDelegate uses customizable factories). Such a factory will never be part of an application neither will it be a feature, and such a module must be loaded *before* the runtime is even trying to scan the application. So there is a clear separation between runtime modules and application components *in theory*. Yes, we did not plan to cover these cases *now*, but it makes sense to start *somewhere*, so we should keep the API open for such enhancements that go much deeper into that separation *in future*. As a result, we *must* keep the interface, and we *must* keep the separation between extensions (runtime modules) and features (application components). Extensions *can* be features (and as of today possibly will be in *many* cases), but they *not necessarily* are.

This is why I always said, extensions are something *different* to application components providers since I invented the interface months (years) ago, and this is why *this PR* is definitively **not** *an alterantive* to `Extension` as it does not even try to cover the sole use case I invented that interface for!

anthonyvdotbe:
@chkal @spericas @andymc12 @arjantijms @jansupol will you please (re-)review this PR? I believe all of your concerns have been addressed, so I'd like to solicit further feedback and/or get approval. Thanks in advance.

In short:
* the (primary) use case is extending a JAX-RS implementation in a portable way (see e.g. [this explanation](https://github.com/eclipse-ee4j/jaxrs-api/pull/947#issuecomment-789024808) by Arjan)
* the goal is to be backward-compatible, which is why this PR specifies what happens in the presence of an `Application` subclass for which `getClasses` and/or `getSingletons` return a non-empty collection (note that the influence of an `Application` subclass is limited to the scope of its `ClassLoader`)

mkarg:
My vote still is -1 as you still remove `Extension`, for the reasons explained in the above discussion with Christian.

anthonyvdotbe:
> My vote still is -1 as you still remove `Extension`, for the reasons explained in the above discussion with Christian.

I respect that, but making this a spec-only change was one of the very reasons to open this PR:
> * make this a spec-only change (as [already mentioned](https://github.com/eclipse-ee4j/jaxrs-api/pull/751#issuecomment-477789591), the `Extension` interface isn't indispensable)

In your [last comment](https://github.com/eclipse-ee4j/jaxrs-api/pull/947#discussion_r611778241) you conclude "we *must* keep the interface". However, after reading the discussion and the reasons given, my conclusion is the exact opposite.

So let's agree to disagree, and try to build consensus with the other members around our respective PRs.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
jansupol: > For example, a runtime module might exist in future that implements parts of RuntimeDelegate (i. e. RuntimeDelegate uses customizable factories). Such a factory will never be part of an application neither will it be a feature, and such a module must be loaded before the runtime is even trying to scan the application.

This is not a good example indeed. RuntimeDelegate implementation is strictly connected to the Jakarta REST implementation and as such, it should be the vendor's responsibility to get the mentioned RuntimeDelegate part onboard. 

I understand the API as an instrument for customers to use. For vendors, we shall define the requirements and demands, not Extensions they are free to use. It is more the customers who will use the Extension interface (or this counterproposal should it be approved) primarily, the vendors likely stay with the module loading mechanism they already have.

spericas:
> My vote still is -1 as you still remove `Extension`, for the reasons explained in the above discussion with Christian.

 As I said before, I do not understand your insistence to keep the interface. As far as I can tell, nobody else wants to keep it either; hope you reconsider and not block this PR.

spericas:
> @chkal @spericas @andymc12 @arjantijms @jansupol will you please (re-)review this PR? I believe all of your concerns have been addressed, so I'd like to solicit further feedback and/or get approval. Thanks in advance.
> 
> In short:
> 
> * the (primary) use case is extending a JAX-RS implementation in a portable way (see e.g. [this explanation](https://github.com/eclipse-ee4j/jaxrs-api/pull/947#issuecomment-789024808) by Arjan)
> * the goal is to be backward-compatible, which is why this PR specifies what happens in the presence of an `Application` subclass for which `getClasses` and/or `getSingletons` return a non-empty collection (note that the influence of an `Application` subclass is limited to the scope of its `ClassLoader`)

I don't think this last bullet point is a requirement for backward compatibility: no existing application or library today would have a services file as defined here that would cause something unexpected to be loaded. If a newer version of a library supports this, its documentation should clarify the new behavior IMO.

anthonyvdotbe:
> I don't think this last bullet point is a requirement for backward compatibility: no existing application or library today would have a services file as defined here that would cause something unexpected to be loaded. If a newer version of a library supports this, its documentation should clarify the new behavior IMO.

Indeed, at this point it's very unlikely that existing applications/libraries define services (though I can imagine that people who write `module-info.java` files might add `provides` clauses as a matter of form/documentation).

The starting point is that I want to maintain the authoritativeness of an `Application` subclass in terms of what gets loaded and what not. The [specification](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#servlet) a.o. says:
> When an Application subclass is present in the archive [...] If either getClasses or getSingletons returns a non-empty collection then only those classes or singletons returned MUST be included in the published JAX-RS application.

Now consider the following sequence of events: `foo.war` uses JAX-RS 3.0/Jakarta EE 9 + `bar.jar` (and contains an `Application` subclass which ...); `bar.jar` releases a new version which a.o. defines services; `foo.war` upgrades its dependency on `bar.jar`; `foo.war` upgrades to JAX-RS 4.0/Jakarta EE 10.
So when `foo.war` upgrades to Jakarta EE 10, I argue that it should load the same `Features` as it did before.

Aside 1: the spec doesn't delimit the authoritativeness of an `Application` subclass more precisely than "the archive" (correct me if I'm wrong, I haven't read the whole spec yet). Now since application servers typically (always?) use 1 `ClassLoader` per WAR, and the `ServiceLoader` mechanism takes a `ClassLoader` argument as well, it seemed logical to use the `Application` subclass's `ClassLoader` as a universally-applicable demarcation line.

Aside 2: `bar.jar` is likely to specify most, if not all, of its `Features` as services, because most `Features` make equally much sense as a "runtime extension" as they do as an "application component" (i.e. some people will want to deploy `bar.jar` server-wide, while other people will want to bundle it in `foo.war` as an application-specific dependency).

Now if this PR wouldn't differentiate between the presence/absence of an `Application` subclass which ..., then in the scenario above, all `Features` defined as services in `bar.jar` would get loaded, *and* there wouldn't be a way for `foo.war` to opt-out of it.

So this is what I mean with "backward-compatible" in this case.

spericas:
> 
> > When an Application subclass is present in the archive [...] If either getClasses or getSingletons returns a non-empty collection then only those classes or singletons returned MUST be included in the published JAX-RS application.

 Right, but the main use case we are going after here is that of libraries that need to extend the JAX-RS _runtime_ to work properly (e.g. by adding a new filter). What comprises an application isn't really changing --unless the application uses this very same mechanism which is possible but not as interesting.

> 
> Now consider the following sequence of events: `foo.war` uses JAX-RS 3.0/Jakarta EE 9 + `bar.jar` (and contains an `Application` subclass which ...); `bar.jar` releases a new version which a.o. defines services; `foo.war` upgrades its dependency on `bar.jar`; `foo.war` upgrades to JAX-RS 4.0/Jakarta EE 10.
> So when `foo.war` upgrades to Jakarta EE 10, I argue that it should load the same `Features` as it did before.

 This example is too contrived. If `bar.jar` defines a new version that includes services, it should state that it needs JAX-RS 4.0 to run properly (how will it run otherwise?). Intermediate mix and matches don't make sense.

anthonyvdotbe:
>> When an Application subclass is present in the archive [...] If either getClasses or getSingletons returns a non-empty collection then only those classes or singletons returned MUST be included in the published JAX-RS application.
> 
> Right, but the main use case we are going after here is that of libraries that need to extend the JAX-RS _runtime_ to work properly (e.g. by adding a new filter). What comprises an application isn't really changing --unless the application uses this very same mechanism which is possible but not as interesting.

Agreed, except for the "which is possible" part. My POV is that an `Application` subclass should be able to retain full control.

>> Now consider the following sequence of events: `foo.war` uses JAX-RS 3.0/Jakarta EE 9 + `bar.jar` (and contains an `Application` subclass which ...); `bar.jar` releases a new version which a.o. defines services; `foo.war` upgrades its dependency on `bar.jar`; `foo.war` upgrades to JAX-RS 4.0/Jakarta EE 10.
>> So when `foo.war` upgrades to Jakarta EE 10, I argue that it should load the same `Features` as it did before.
> 
> This example is too contrived. If `bar.jar` defines a new version that includes services, it should state that it needs JAX-RS 4.0 to run properly (how will it run otherwise?). Intermediate mix and matches don't make sense.

In this example, say `bar.jar` is a domain-specific library, which also contains a `Feature` X to integrate with JAX-RS. So there will be JAX-RS applications which register X, and others which don't (e.g. because they need/register only a subset of X's providers manually in their `Application` subclass). But all applications use the domain-specific APIs that `bar.jar` provides, and want to stay up-to-date with `bar.jar`.
Now if JAX-RS 3.1 is released, some users will ask `bar.jar` to define X as a service, so that they can deploy it on their server and have it picked up by the runtime from there for all of their applications. So `bar.jar` would release a new version to do so, but `Feature` X would still be equally usable as an application dependency, just like before. So all users would upgrade to the new version of `bar.jar`, but only some users would upgrade to JAX-RS 4.0 and deploy `bar.jar` on the server. Now when later in the future other users, which explicitly don't register X, upgrade to JAX-RS 4.0, X will suddenly be registered. And this is the scenario I want to prevent.

Hope this helps to clarify my rationale. If not, will you please explain which exact changes you'd like to see the current PR and why?

spericas:
> > > When an Application subclass is present in the archive [...] If either getClasses or getSingletons returns a non-empty collection then only those classes or singletons returned MUST be included in the published JAX-RS application.
> > 
> > 
> > Right, but the main use case we are going after here is that of libraries that need to extend the JAX-RS _runtime_ to work properly (e.g. by adding a new filter). What comprises an application isn't really changing --unless the application uses this very same mechanism which is possible but not as interesting.
> 
> Agreed, except for the "which is possible" part. My POV is that an `Application` subclass should be able to retain full control.

 Let's flip this around. Suppose I have an application that does not define an `Application` subclass and is already using the services mechanism described in this PR with some library X. For whatever reason, a later version of the application adds a new `Application` subclass: the moment it does that library X stops working and so does the app since it depends on it. The developer then scratches his/her head for days ...

 Giving full control to the application is not necessarily a bad idea; however, the way it is proposed here is IMO. For example, Jersey actually supports the use case you have in mind, but using system properties to disable service loading. If we want to include a proposal along those lines, that would be fine. 

> 
> Hope this helps to clarify my rationale. If not, will you please explain which exact changes you'd like to see the current PR and why?

 See above. I'd be nice to hear other opinions as well.

anthonyvdotbe:
> Let's flip this around. Suppose I have an application that does not define an `Application` subclass and is already using the services mechanism described in this PR with some library X. For whatever reason, a later version of the application adds a new `Application` subclass: the moment it does that library X stops working and so does the app since it depends on it. The developer then scratches his/her head for days ...

In that case the developer would have to override `getClasses` (for the services mechanism to stop working). But to implement `getClasses`, they'd have to explicitly determine which classes to register, so I don't see the issue? Or am I missing something?

Also note that an `Application` subclass can still trivially do the service loading explicitly if desired, by simply adding the result of `ServiceLoader.load(Feature.class).stream().map(Provider::type)` (and same for `DynamicFeature`) to the result of `getClasses`.

> Giving full control to the application is not necessarily a bad idea; however, the way it is proposed here is IMO. For example, Jersey actually supports the use case you have in mind, but using system properties to disable service loading. If we want to include a proposal along those lines, that would be fine.

So IIUC, it's simply a matter of what should be the default: I'm proposing to have the service loading disabled, with the `Application` subclass being able to do it explicitly if desired. Whereas you're proposing to have the service loading enabled, with a way to disable it using system properties.

spericas:
> > Let's flip this around. Suppose I have an application that does not define an `Application` subclass and is already using the services mechanism described in this PR with some library X. For whatever reason, a later version of the application adds a new `Application` subclass: the moment it does that library X stops working and so does the app since it depends on it. The developer then scratches his/her head for days ...
> 
> In that case the developer would have to override `getClasses` (for the services mechanism to stop working). But to implement `getClasses`, they'd have to explicitly determine which classes to register, so I don't see the issue? Or am I missing something?

 Yes, you are, that they need to know how or what library `X` is registering for it to run (note that most of the time library `X` is a third-party dependency). That is why we need this feature, for libraries to install what they need without the application necessarily knowing how to do this. We use the equivalent feature of Jersey all the time.

> 
> Also note that an `Application` subclass can still trivially do the service loading explicitly if desired, by simply adding the result of `ServiceLoader.load(Feature.class).stream().map(Provider::type)` (and same for `DynamicFeature`) to the result of `getClasses`.

 See above.

> 
> > Giving full control to the application is not necessarily a bad idea; however, the way it is proposed here is IMO. For example, Jersey actually supports the use case you have in mind, but using system properties to disable service loading. If we want to include a proposal along those lines, that would be fine.
> 
> So IIUC, it's simply a matter of what should be the default: I'm proposing to have the service loading disabled, with the `Application` subclass being able to do it explicitly if desired. Whereas you're proposing to have the service loading enabled, with a way to disable it using system properties.

 As explained above, the difference is that if my app depends on library `X`, `Y` and `Z`, I don't need to understand their inner workings to define my `getClasses` but I can, if I want to, disable all their service loading capabilities for whatever reason necessary.

andymc12:
Regarding the opt-out mechanism... I like @spericas's idea of using a system property - or application property (i.e. `Application#getProperties()`) to control whether the implementation should auto-discover and register service extensions.  Something like `jakarta.ws.rs.autoDiscoverFeatureExtensions=true|false` (with the default being true) or maybe allow users to disable specific features with something like `jakarta.ws.rs.disableFeatureExtension=<comma-separated list of feature classes or * for all>`.  This should avoid the scenario where a user implements `Application#getClasses` and all of sudden things like Open Telemetry or MP Metrics stops working.  But it still gives app developers and deployers a chance to opt-out if they want.

anthonyvdotbe:
Thanks @spericas, it clicked!

I've updated the PR, so any feedback is welcome.

The opt-out mechanism is an application property because, in the context of an application server, I believe it ought to be possible for different WARs to make different choices. The property is simply named `...loadServices` because it's reminiscent of `ServiceLoader`, and if in the future more kinds of providers are to be discovered, we wouldn't need to introduce another property.

Disabling specific `Features` isn't supported at the moment. However, if this is desired, I'd propose not to introduce a property for that, but instead add the following method to `Application`:

```java
public boolean filterService(Class<?> service) {
    return true;
}
```

which a JAX-RS implementation would use like: 

```java
Application app = ...;
List<Class<?>> filteredFeatures = ServiceLoader.load(Feature.class).stream().map(Provider::type).filter(app::filterService).toList();
```

> It also might be nice to have more background text in the spec document that explains the use case for these service extensions - and maybe an example too.

@andymc12 I've updated the note (credit to @spericas, as the first part is heavily inspired by his last comment). Will you please let me know if it's sufficient as background text? Also, is there a specific way to distinguish informative text from normative text in the specification?

mkarg:
> > My vote still is -1 as you still remove `Extension`, for the reasons explained in the above discussion with Christian.
> 
> As I said before, I do not understand your insistence to keep the interface. As far as I can tell, nobody else wants to keep it either; hope you reconsider and not block this PR.

My opinion is unchanged and the reasons are laid out.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
mkarg: I invented the `Extension` interface solely for implementation by vendors not users -- by runtime providers and by third party providers of cross-vendor runtime modules, actually, independent of what anybody *thinks* or *assumes*. Example: Application declares to use JSON-B, and deployer decides whether to use original Jersey-JSON-B-media-module or let Jersey use Johnzon-JSONB cross-vendor module. This is neither the application's choice not the runtime vendor's choice.

mkarg:
> So let's agree to disagree, and try to build consensus with the other members around our respective PRs.

Sorry to say, but the rules of this project are not to find concensus, but to have a majority and ZERO vetos.

andymc12 (dismissed):
LGTM - thanks @anthonyvdotbe!

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
```diff
@@ -35,13 +35,14 @@ process only those classes that are annotated with `@Provider`.
 
 [[services]]
 ==== Services
-On runtime initialization, an implementation MUST discover and register certain service providers in `jakarta.ws.rs.core.Configuration` runtime configurations.
-For each service X, where X is `jakarta.ws.rs.core.Feature` or `jakarta.ws.rs.container.DynamicFeature`, the set of providers to be registered MUST be determined by the method invocations below. However, when an `Application` subclass is present, and either `getClasses` or `getSingletons` returns a non-empty collection, those invocations for which the (possibly implicit) `ClassLoader` argument equals the `Application` subclass's `ClassLoader` MUST be skipped.
+On runtime initialization, an implementation MUST discover and register certain service providers in `jakarta.ws.rs.core.Configuration` runtime configurations, if and only if service loading is enabled.
+Service loading is enabled, except when there's an `Application` subclass whose `getProperties` method returns a map which contains the key `jakarta.ws.rs.loadServices` with a value of `Boolean.FALSE`.
```
spericas, line 39: Historically, the spec document has been written without using contractions such as "there's". Could we use "there is" here?

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
```diff
@@ -35,13 +35,14 @@ process only those classes that are annotated with `@Provider`.
...
+For each service X, where X is `jakarta.ws.rs.core.Feature` or `jakarta.ws.rs.container.DynamicFeature`, the set of providers to be registered MUST be determined as follows:
 
 * first, register all providers discovered with `ServiceLoader.load(X.class)`
 * additionally, register all providers discovered with `ServiceLoader.load(X.class, X.class.getClassLoader())`
 
-Note that the primary target audience for this, are system integrators rather than application developers. For example, it allows to write portable libraries which integrate several Jakarta EE specifications without using any implementation-specific code.
+Note that service loading allows libraries to be self-sufficient: it allows them to register what they need with the JAX-RS implementation, without intervention from the application. From the viewpoint of an application developer, this means that they do not need to understand the inner workings of their dependent libraries when overriding `getClasses`. As another example, it allows system integrators to extend the JAX-RS implementation (e.g. to provide integrations with other Jakarta EE specifications) in a completely portable manner.
```
spericas, line 45: Is there a need for a "," in this sentence?

spericas:
> Thanks @spericas, it clicked!
> 
> I've updated the PR, so any feedback is welcome.
> 
> The opt-out mechanism is an application property because, in the context of an application server, I believe it ought to be possible for different WARs to make different choices. The property is simply named `...loadServices` because it's reminiscent of `ServiceLoader`, and if in the future more kinds of providers are to be discovered, we wouldn't need to introduce another property.
> 
> Disabling specific `Features` isn't supported at the moment. 

 I don't think we need this for a first version; a nice to have IMO.

> ```java
> public boolean filterService(Class<?> service) {
>     return true;
> }
> ```

 I suspect `Application` will change somewhat in 4.0, so if we can avoid changes now, it would be better.

> @andymc12 I've updated the note (credit to @spericas, as the first part is heavily inspired by his last comment). Will you please let me know if it's sufficient as background text? Also, is there a specific way to distinguish informative text from normative text in the specification?

 Unfortunately, we don't have that sort of separation, which basically makes everything that isn't an example _normative_.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
anthonyvdotbe: The comma in "implementation, without intervention from the application" has been removed. There's a comma in the following 2 sentences as well, but I personally find these helpful. I've also reformatted the section to be in line with the rest of the file.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
anthonyvdotbe: Sure, done.

spericas (approved):
LGTM

jansupol:
> Sorry to say, but the rules of this project are not to find concensus, but to have THREE +1 and ZERO vetos.
Then I am sorry @mkarg, but your [Extension PR](https://github.com/eclipse-ee4j/jaxrs-api/pull/751) did not match this criterion.

If the PR was meant for the vendors, it was not clear, as this PR proves, and I am +1 for this PR.

jansupol (approved):

mkarg:
> > Sorry to say, but the rules of this project are not to find concensus, but to have THREE +1 and ZERO vetos.
> 
> Then I am sorry @mkarg, but your [Extension PR](https://github.com/eclipse-ee4j/jaxrs-api/pull/751) did not match this criterion.
> If the PR was meant for the vendors, it was not clear, as this PR proves, and I am +1 for this PR.

This is wrong as my own +1 was implied when opening the PR (this is how Github works).

spericas:
> > So let's agree to disagree, and try to build consensus with the other members around our respective PRs.
> 
> Sorry to say, but the rules of this project are not to find concensus, but to have THREE +1 and ZERO vetos.

You have requested a change related to Java 8 compatibility. It has been cleared out that this is taken care by in the PR. Please review your change request.

andymc12 (approved):
LGTM - thanks!

mkarg:
> > > So let's agree to disagree, and try to build consensus with the other members around our respective PRs.
> > 
> > 
> > Sorry to say, but the rules of this project are not to find concensus, but to have THREE +1 and ZERO vetos.
> 
> You have requested a change related to Java 8 compatibility. It has been cleared out that this is taken care by in the PR. Please review your change request.

I will re-review this PR soon, no hurry. :-)

chkal (approved):
Looks good to me! 

@anthonyvdotbe: Thanks for your work on this.

A little more explanation: I think that this PR is a simpler version of what `Extension` provides, and therefore I prefer it over `Extension`. I always thought that `Extension` was primarily for writing portable JAX-RS libraries (like Soteria, Mojarra, Payara JWT, Krazo, etc). But as @mkarg mentioned [here](https://github.com/eclipse-ee4j/jaxrs-api/pull/947#discussion_r614266313), it is more meant to be implemented only by vendors and not by users, for which I don't really see a need. What is really missing today, is a way for libraries to contribute JAX-RS provides in a portable way, which will work fine with the mechanism contained in the PR.

mkarg (changes requested):
While I still am not convinced that the `Extensions` interface wouldn't be the superior approach (as it allows us to add methods to it giving access into the runtime implementation itself, hence becoming "real" runtime plugins instead of just "automatically loading providers"), I will accept **if** there is a *majority* of committers that actively support the request to *remove* the effectively already concluded feature. Having said that, I am voting -1 **just for the time being** as it shouldn't be enough that *some* committers are +1 in a case of **feature dropping**. Due to that, and to not losing more time than necessary, I agree to give up my veto as soon as all major vendors are +1 for this PR. Hence, I am explicitly asking @ronsigal and @arjantijms to cast a vote. As soon as both vote +1, I will turn my vote into a 0, effectively making the way free for a merge. Sorry for this barrier, but again, we are **dropping** something here (not just adding something), so I want to be sure that Red Hat and others are good with that, too, not just IBM and Oracle.

spericas:
> While I still am not convinced that the `Extensions` interface wouldn't be the superior approach (as it allows us to add methods to it giving access into the runtime implementation itself, hence becoming "real" runtime plugins instead of just "automatically loading providers"), I will accept **if** there is a _majority_ of committers that actively support the request to _remove_ the effectively already concluded feature. Having said that, I am voting -1 **just for the time being** as it shouldn't be enough that _some_ committers are +1 in a case of **feature dropping**. Due to that, and to not losing more time than necessary, I agree to give up my veto as soon as all major vendors are +1 for this PR. Hence, I am explicitly asking @ronsigal and @arjantijms to cast a vote. As soon as both vote +1, I will turn my vote into a 0, effectively making the way free for a merge. Sorry for this barrier, but again, we are **dropping** something here (not just adding something), so I want to be sure that Red Hat and others are good with that, too, not just IBM and Oracle.

Just to be clear, the fact that we are removing an interface, doesn't mean that we are _dropping anything_. In fact, we've improved the original proposal and _simplified_ the API by removing what was unnecessary without affecting the "feature" that we intend to support.

mkarg:
> Just to be clear, the fact that we are removing an interface, doesn't mean that we are _dropping anything_. In fact, we've improved the original proposal and _simplified_ the API by removing what was unnecessary without affecting the "feature" that we intend to support.

This is what nobody seems to understand: We *do* drop a feature, as the `Extension` interface was intended by me *by intentention* (not just as a technical necessarity) to detect the difference between application-provided providers and runtime-module provided components. That difference *is* effectively dropped with this PR.

As I said, I am not standing in the way to giving up *that*, if at least two more vendors are plus one besides IBM and Oracle.

mkarg:
@arjantijms @ronsigal Please review this PR and cast your vote. Thanks.

spericas:
> > Just to be clear, the fact that we are removing an interface, doesn't mean that we are _dropping anything_. In fact, we've improved the original proposal and _simplified_ the API by removing what was unnecessary without affecting the "feature" that we intend to support.
> 
> This is what nobody seems to understand: We _do_ drop a feature, as the `Extension` interface was intended by me _by intentention_ (not just as a technical necessarity) to detect the difference between application-provided providers and runtime-module provided components. That difference _is_ effectively dropped with this PR.

 As it was explained a few times, there's absolutely no way in practice to prevent applications from using the `Extension` interface, thus making the distinction mostly academic.

> 
> As I said, I am not standing in the way to giving up _that_, if at least two more vendors are plus one besides IBM and Oracle.

 More reviews are always better, but I don't think the approvals are in any way related to vendor preference.

mkarg:
> As it was explained a few times, there's absolutely no way in practice to prevent applications from using the `Extension` interface, thus making the distinction mostly academic.

Sure, but this does not change my opinion about the idea behind the interface, still doesn't make *this* PR an "alternative" to *what I wanted the interface for*, and does not change my opinion about the vote on *this* PR. 

> > As I said, I am not standing in the way to giving up _that_, if at least two more vendors are plus one besides IBM and Oracle.
> More reviews are always better, but I don't think the approvals are in any way related to vendor preference.

I abstain from commenting on that, given the fact that *most* of the committers in this project do *not* have their status due to the Eclipse rules (i. e. because they actively committed to this project), but only due to their past or current employment with a particular company. As a result, I personally care a lot about asking committers from *all* vendors to cast a vote.

arjantijms (approved):

spericas:
> so I want to be sure that Red Hat and others are good with that, too, not just IBM and Oracle.

@mkarg See new approvals.

mkarg:
Thank you for your vote, @arjantijms. Can you please cast your vote, @ronsigal?

mkarg (commented):
I think others had lots of time to cast their vote meanwhile. As nobody else voted -1, I hereby replace my -1 by 0 as announced. Thanks everybody for the vivid discussion and the patience. While I did not get completely what I originally wanted when I invented the `Extensions` interface, I do think that automatic loading of feature providers is a great feature even on its own.

mkarg (approved):

mkarg (commented):
voting 0

spericas:
Merging based on all approvals

mkarg:
Congratulations everyone! This is a nice feature!

mkarg:
Is there a publicly available implementation of this already?

PR #948, opened 2021-03-04
[418] multipart/form-data support

### LINKED ISSUE #418: Support for MultiPart forms
The spec currently provides no specific support for multpart/form. While it is possible of course to have a MBR for the format and some utility libraries provide this (e.g. Apache Clerezza jaxrs.utils) many implemententations provide a tighter integration with annotation supporting the fields of MultiPart Forms (jersey with @FormDataParam and CXF with @Multipart) code taht relies on these extension is no longer portable.

The Jax-rs spec should standartize support for this.
#### Affected Versions
[2.0]

andymc12:
Fixes #418 

Includes new API, spec text and some simple examples.


mkarg:
@andymc12 Thank you for contributing this! Unfortunately I am busy with JavaLand organization recently, so the next two weeks I won't have time to actually review this code. Sorry for that. I pretend to have a look at it after JavaLand! :-)

andymc12:
@mkarg - I hope JavaLand went well - it looks like it was well attended with lots of notable speakers!  Hopefully you have recovered and can review this PR?

@chkal @spericas - could you take a look as well?  Thanks!

mkarg:
> @mkarg - I hope JavaLand went well - it looks like it was well attended with lots of notable speakers! Hopefully you have recovered and can review this PR?

Thanks, Andy, yes, we had approx. 1500 online attendees and a quite shiny lineup of notable speakers. Despite all hopes, I am still busy today and tomorrow, so if things run well, will look into this PR on Sunday. Stay tuned! :-)

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Returns the filename of this part. This will be the "filename" attribute of
+     * the {@code Content-Disposition} header for this part. A filename is not
+     * required in a part, so if a filename is not present it will return
+     * {@code Optional.empty()}
+     * 
+     * @return an {@code Optional<String>} indicating the filename if present
+     */
+    Optional<String> getFileName();
```
mkarg, line 80: Is using Optional a good idea? What is the difference between Optional.empty() and null?

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
```diff
@@ -126,3 +129,58 @@ the media type `application/xml` given its higher qs-value. Note that
...
+
+    @POST
+    @Consumes(MediaType.MULTIPART_FORM_DATA)
+    public Response postWidget(@FormParam("part1Name") String part1, 
+                               @FormParam("part2Name") InputStream part2,
+                               @FormParam("part3Name") Part part3) {...}
+}
+----
+
+The only way to access the headers for a particular part is to use the `Part`
```
mkarg, line 182: Why don't we allow to invoke parameter handlers?

mkarg (commented):
I'm done with an initial preview. Kudos, this is great work! But I have some questions, see inlined. Also I wonder why we *force* the use of streams? I understand that streams have benefits, by why not *also* allows parameter conversion or using `Entity` classes?

examples/src/main/java/jaxrs/examples/multipart/MultipartClient.java
```diff
@@ -0,0 +1,56 @@
...
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.Part;
+
+public class MultipartClient {
+
+    public boolean sendPdfs(File dir) throws IOException {
+        List<Part> parts = new ArrayList<>();
+        for (File f : dir.listFiles()) {
+            parts.add(Part.newBuilder(f.getName()).entityStream(f.getName(), new FileInputStream(f))
```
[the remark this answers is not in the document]
andymc12: Agreed - will fix.

examples/src/main/java/jaxrs/examples/multipart/MultipartClient.java
```diff
@@ -0,0 +1,56 @@
...
+    public List<File> retrievePdfs(String remoteDirName) throws IOException {
+        Client client = ClientBuilder.newClient();
+        WebTarget target = client.target("http://localhost:9080/multipart").queryParam("dirName", remoteDirName);
+        Response response = target.request(MediaType.MULTIPART_FORM_DATA).get();
+        List<Part> parts = response.readEntity(new GenericType<List<Part>>(){});
+        List<File> files = new ArrayList<>(parts.size());
+        for (Part p : parts) {
+            File f = new File(p.getFileName().orElse(p.getName()));
+            f.createNewFile();
+            Files.copy(p.getEntityStream(), f.toPath());
```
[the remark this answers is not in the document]
andymc12: Sounds good.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
```diff
@@ -0,0 +1,195 @@
...
+import jakarta.ws.rs.core.MultivaluedMap;
+
+/**
+ * A {@code Part} is one part of a multipart entity. As defined in
+ * <a href="https://tools.ietf.org/html/rfc7578">RFC 7578</a>, a multipart
+ * request or response must have a content type of "multipart/form-data" with a
+ * {@code boundary} parameter indicating where one part ends the next may begin.
+ * <p>
+ * Multipart entities may be received in a resource method as a collection of
+ * parts (ex: {@code List<Part>}) or as a form parameter (ex:
```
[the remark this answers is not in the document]
andymc12: I've seen and used both, but I don't mind using "e.g." to be consistent with other parts of the project.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
```diff
@@ -0,0 +1,195 @@
...
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.MultivaluedMap;
+
+/**
+ * A {@code Part} is one part of a multipart entity. As defined in
+ * <a href="https://tools.ietf.org/html/rfc7578">RFC 7578</a>, a multipart
+ * request or response must have a content type of "multipart/form-data" with a
+ * {@code boundary} parameter indicating where one part ends the next may begin.
+ * <p>
+ * Multipart entities may be received in a resource method as a collection of
```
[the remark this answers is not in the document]
andymc12: I would expect that the user would get an error because there is no suitable ParamConverter provider to convert the query/matrix param into a `Part`.  Users or vendors could build their own ParamConverter providers to make that work, but it doesn't seem like that would be a very common use case.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
andymc12: I prefer `Optional` because it aligns with the RFC that the file name attribute is not required. It's also "showing my support for modern Java." ;-)

In my example, I think it makes it more readable to detect when the file name attribute exists or not - i.e. `String filename = part.getFileName().orElse(part.getName() + ".txt");`

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
andymc12: Sorry, I don't understand your question.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
```diff
@@ -0,0 +1,195 @@
...
+ */
+public interface Part {
+
+    /**
+     * Creates a new {@code Part.Builder} instance.
+     * 
+     * @param partName name of the part to create within the multipart entity
+     * @return {@link Builder} for building new {@link Part} instances
+     */
+    static Builder newBuilder(String partName) {
```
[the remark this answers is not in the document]
andymc12: Good point.  I've also seen `with...` when creating builders - i.e. `Part.withName(String)`.  From a language perspective, I like "withName" better than "fromName" since the builder will ultimately create a `Part` object _with_ the specified name, but I'm fine with either.  Wdyt?

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
```diff
@@ -0,0 +1,195 @@
...
+    Optional<String> getFileName();
+
+    /**
+     * Returns the input stream for this part. This is the content body of the part
+     * and is accessed as a stream to avoid loading potentially large amounts of
+     * data into the heap.
+     * 
+     * @return an {@code InputStream} representing the content of this part
+     */
+    InputStream getEntityStream();
```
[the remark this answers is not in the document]
andymc12: Sounds good - will change.

examples/src/main/java/jaxrs/examples/multipart/MultipartClient.java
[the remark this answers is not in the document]
andymc12: Yeah, that's a good question - here are a few thoughts (assuming that we change `newBuilder(...)` to `withName(...)`:
(1) We could add a `withFileName(String filename)` method that would set both the name and filename to the passed in String value.
(2) We could add a `withName(String name, boolean sameAsFileName)` - and when `sameAsFileName` is true, it would be equivalent to calling `withName(name).fileName(name)`
(3) Some other convention?

I think I lean towards (1).

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
```diff
@@ -126,3 +129,58 @@ the media type `application/xml` given its higher qs-value. Note that
...
+type. The `InputStream` and `String` parameter types will only provide the
+content of the part. Note that parts of a multipart entity can be quite large,
+so care should be taken when using `String` parameter types as that will load
+the entire content of the part into the Java heap.
```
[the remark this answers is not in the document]
andymc12: >What is the character encoding used for the String case?

My interpretation of the RFC is that the encoding should be UTF-8 unless specified otherwise.  The RFC says that the encoding/charset can be specified per-part via the part's `Content-Type` header.  It also mentions a "hidden" part with the reserved name of "\_charset\_" and whose body would specify the charset for all other parts. 

I think I like the per-part approach better - and it is easy enough to specify with the APIs in this PR.

The "hidden" _charset_ part could also work with the current APIs, but would be a little clunky... maybe we want to create a quick helper method to generate a default charset part? Wdyt?

>  Should we support byte[]?

🤷  maybe??  I was hesitant to support `String` because it forces the content to be loaded into the heap - and that could be a vulnerability without proper safeguards.  I ended up pulling in `String` because RESTEasy supports it (I think CXF might also), so I figured users must want that - maybe for secured environment where they know that that parts are small enough to handle (btw, both RESTEasy and CXF has config settings to limit the size of these attachments which should reduce the vulnerability).

If we support `byte[]` do you think we should drop support for `String`? That would put the encoding burden on the user - and they wouldn't be able to read the part's headers to see if it had a non-default encoding...  Wdyt?

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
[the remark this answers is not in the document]
spericas: Yes, that seems fine.

examples/src/main/java/jaxrs/examples/multipart/MultipartClient.java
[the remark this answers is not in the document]
spericas: Yes, definitely (1).

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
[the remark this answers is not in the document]
spericas: Yes, on UTF-8 and charset in `Content-Type`. The hidden part option seems like an _ugly hack_ to me. You're right about the issue with `charset` and `byte[]`, but it also applies to the `InputStream` alternative. It's simple to convert an `InputStream` to a `byte[]`, so the latter may not be really needed.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
spericas: Yes, `Optional<?>` would have been used earlier in the API had it been available.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
```diff
@@ -0,0 +1,195 @@
+/*******************************************************************
+* Copyright (c) 2021 Eclipse Foundation
+*
+* This specification document is made available under the terms
+* of the Eclipse Foundation Specification License v1.0, which is
+* available at https://www.eclipse.org/legal/efsl.php.
+*******************************************************************/
+package jakarta.ws.rs.ext;
```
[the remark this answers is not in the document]
spericas: I think using `ext` is fine.

andymc12:
Hi @mkarg @spericas - thanks for all of the feedback - excellent suggestions!  I think I have addressed all of your concerns. Would you mind taking another look and approving if you think it's ready to go in?  Thanks again.

mkarg:
> Hi @mkarg @spericas - thanks for all of the feedback - excellent suggestions! I think I have addressed all of your concerns. Would you mind taking another look and approving if you think it's ready to go in? Thanks again.

@andymc12 Excellent PR, but please let's have Easter first. :-)

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
```diff
@@ -0,0 +1,219 @@
...
+     * Returns the input stream for this part. This is the content body of the part
+     * and is accessed as a stream to avoid loading potentially large amounts of
+     * data into the heap.
+     * 
+     * It is the responsibility of the calling code to close this stream after
+     * receiving it.
+     * 
+     * @return an {@code InputStream} representing the content of this part
+     */
+    InputStream getContent();
```
chkal, line 109: I wonder if it is really sufficient that the content is only available as an `InputStream`. What about adding something like:

    <T> T getContent(Class<T> type);

This method could use the available MBRs to convert the content to all kind of types like `byte[]`, `java.io.Reader`, `javax.json.JsonObject`, etc. Or am I missing something. AFAIK Jersey supports this as well.

Of course the argument that `InputStream` prevents loading much data into the heap is valid, but that's actually the case for all resources methods which consume the request body without using an `InputStream`, no?

Also, maybe `getEntity()` would be a better name for the method?

chkal (commented):
@andymc12 Thank you so much for working on this. See my comment in the source.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
mkarg: In fact we cannot name it `getEntity()` as the entity is the container holding *all* parts.

chkal: Well, the term *entity* isn't strictly defined in the spec, AFAIK. But a few related terms are:

There is the *request entity* and the *response entity* for which there can only be at most one per request/response. Then we have the *entity parameter* which basically allows the resource method to access the *request entity* created from the *request entity body*.

For multipart requests we also have a single *request entity*. But this entity actually represents one or more *parts* and each one is quite similar to a *request entity* in a sense that we have a body and a media type. Maybe we could call it *part entity* or something like that in the API docs.

mkarg: Actually the term "entity" is pretty clearly defined by the HTTP spec, but not even mentioned by the specs for multiform, so we should definitively *not* mix things here, but stick with the term "content" which the multiform specs actually comes up with.

chkal: > Actually the term "entity" is pretty clearly defined by the HTTP spec

Thanks for the hint. I wasn't aware that the term "entity" is actually from the HTTP spec. However, I still like the idea that a part also has an "entity" because it is actually a very similar concept. Jersey uses the term entity as well in the [BodyPart](https://eclipse-ee4j.github.io/jersey.github.io/apidocs/latest/jersey/org/glassfish/jersey/media/multipart/BodyPart.html) class. However, I would love to hear other thoughts as well.

spericas: Calling the content of a part an "entity" is confusing IMO for the reasons already given here.

chkal: Ok. Beside the name change suggested, what about my other comment?

> I wonder if it is really sufficient that the content is only available as an `InputStream`. What about adding something like:
> 
> ```
> <T> T getContent(Class<T> type);
> ```
> 
> This method could use the available MBRs to convert the content to all kind of types like `byte[]`, `java.io.Reader`, `javax.json.JsonObject`, etc. Or am I missing something. AFAIK Jersey supports this as well.
> 
> Of course the argument that `InputStream` prevents loading much data into the heap is valid, but that's actually the case for all resources methods which consume the request body without using an `InputStream`, no?

mkarg: I second that. InputStream is fine, but it should not be mandatory -- even if the multiforms specs only talk about "files".

andymc12: Yeah, I think that could work - and I think Santiago mentioned something similar.  Would we also need a separate method to handle `GenericType` parameters to convert to generic types? Something like this? `<T> T getContent(GenericType<T> type);`

I'll go ahead and start coding up both methods - but let me know if the generic type approach is unnecessary (I think we'd need it though to account for type erasure).

andymc12: Additional `getContent` methods are in.  Please take a look and let me know if there's anything in those methods that I should change.  Thanks again!

andymc12:
@spericas @mkarg @chkal Thanks for the great review comments so far!  I think I've addressed all concerns and suggestions.  Would you mind taking another look and approving?  Thanks!

spericas (approved):
LGTM

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
```diff
@@ -126,3 +129,62 @@ the media type `application/xml` given its higher qs-value. Note that
...
+        }
+        return Response.ok().build();
+    }
+
+}
+----
+
+Another approach is to use `@FormParam` parameters where the value in the
+annotation corresponds to the name of the part. The parameter type may be
+a `jakarta.ws.rs.ext.Part`, a `java.io.InputStream` or a `String`. Here is
```
chkal, line 166: I wonder if we should really restrict the type to `Part`, `InputStream` and `String` as the API now provides `Part.getContent(Class<T>)` which effectively allows using MBRs to get other types. Why not support other types as well? Wouldn't this be more consistent with the `Part` API and also easy to do from an implementation perspective?

If we decide to do so, the wording in the next paragraph will need adjustments as well.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
chkal: Thanks a lot @andymc12. And sorry for my delayed response.

> Would we also need a separate method to handle `GenericType` parameters to convert to generic types? Something like this? `<T> T getContent(GenericType<T> type);`

Good question. Not sure TBH. But I guess we would need it to supported more advanced use cases like nested multipart entities!? For example:

```java
List<Part> children = part.getContent( new GenericType<List<Part>>() {} );
```

chkal (commented):
@andymc12 Thanks a lot for integrating the latest changes. This looks awesome. I'm basically +1 for this, but added one comment. It would be great to hear your thoughts.

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
spericas: I was satisfied with the original types supported for a first version, but since we added `getContent(Class)`, it seems reasonable to assume other types should be allowed here as well as @chkal points out.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Part.java
spericas: Are nested multipart's even a thing? Seems like a can of worms. Let's keep things simple for a first version of the API.

chkal: No, nested multiparts doesn't make much sense. This was just an example for using generics with `getContent`.

But @andymc12 already added `T getContent(GenericType<T> type)` below, so I guess the issue regarding support for generic types is already resolved.

andymc12: Right, I added the `GenericType` support more for collection use cases than for nested multiparts.  So instead of @chkal's example, I would use an example like this:
```
List<String> lines = part.getContent(new GenericType<List<String>>(){});
```

or:
```
Set<Widget> orderForm = part.getContent(new GenericType<Set<Widget>>(){});
```

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
andymc12: Agreed.  I'll add that shortly.

andymc12:
Added support for `@FormParam("partName") SomeObject param` in the spec text.  

@chkal, @spericas, @mkarg, others - would you take another look and (re-)approve?  Thanks!

spericas (approved):

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
```diff
@@ -126,3 +129,65 @@ the media type `application/xml` given its higher qs-value. Note that
...
+    }
+
+}
+----
+
+Another approach is to use `@FormParam` parameters where the value in the
+annotation corresponds to the name of the part. The parameter type may be
+a `jakarta.ws.rs.ext.Part`, a `java.io.InputStream`, a `String`, or any type
+that can be converted to using a registered `MessageBodyReader` or
+`ParamConverter`. Here is
```
chkal, line 168: Sorry. One last question. 

The API docs of `Part#getContent(Class|GenericType)` only mentions `MessageBodyReader`, but here we allow `ParamConverter` as well.

I'm wondering if `ParamConverter` makes sense here. If we allow both, we will most likely have to define which one of the two mechanisms is preferred over the other. And `ParamConverter` uses a `String` as the input and in the next paragraph we explicitly warn about `String` being not optimal as it requires to read the full entity into the heap.

andymc12: My thinking here was that `ParamConverter` providers are already responsible for converting Strings to parameter types.  The only real difference here (with `@FormParam`) is that we have an `InputStream` instead of a a `String` - but since we allow the implementation to convert the `InputStream` to a `String` for `@FormParam(...) String`, it seemed intuitive (to me, at least :) ) that the implementation should first check that it can convert to the parameter type using registered MBRs, and if it finds none, then it could check to see if a `ParamConverter` is registered for the parameter type.  If so, then the implementation would convert the part's `InputStream` to `String` and then invoke the `ParamConverter`.  

For `Part#getContent(...)`, I don't think it makes sense to use `ParamConverter` because at this point, it's not a parameter; the parameter was `Part`.  A MBR does make sense because it is reading the message body of the individual part to the specified class or `GenericType`.

andymc12: BTW, I appreciate the thoroughness of your review.  I'd certainly rather spend time up front to get it right rather than trying to patch something that was broken from the start.  :)

jansupol:
@andymc12 `Part` seems to be quite a general name. Possibly, a part of anything else can arise and the name would clash. Would it be better to tell Part of what this class is? `EntityPart` perhaps?

mkarg:
Speaking in terms of Java, it would be `Entity.Part`, as it is not to be seen as an *alternative* to an entity, but as a *part* of it.

andymc12:
@jansupol @mkarg 
> @andymc12 Part seems to be quite a general name. Possibly, a part of anything else can arise and the name would clash. Would it be better to tell Part of what this class is? EntityPart perhaps?

> Speaking in terms of Java, it would be Entity.Part, as it is not to be seen as an alternative to an entity, but as a part of it.

I see that `Part` is used in the servlet spec (for pretty much the same thing as we're using it for here) and in the Java/Jakarta Mail spec.  I think it's probably rare enough of a usage that it probably won't clash (and worst case, the package name isn't super long), but I'm not averse to renaming it.

I like Markus's idea of `Entity.Part`, but the `Entity` class is in the `client` package.  Since this class can be used on both the client and server, I'd like to avoid confusion and would probably go with `EntityPart` in the `ext` package.

andymc12:
Renamed `Part` to `EntityPart`.

@spericas @chkal @mkarg @jansupol could you take another look and approve?  Thanks again!

mkarg:
> @jansupol @mkarg
> 
> > @andymc12 Part seems to be quite a general name. Possibly, a part of anything else can arise and the name would clash. Would it be better to tell Part of what this class is? EntityPart perhaps?
> 
> > Speaking in terms of Java, it would be Entity.Part, as it is not to be seen as an alternative to an entity, but as a part of it.
> 
> I see that `Part` is used in the servlet spec (for pretty much the same thing as we're using it for here) and in the Java/Jakarta Mail spec. I think it's probably rare enough of a usage that it probably won't clash (and worst case, the package name isn't super long), but I'm not averse to renaming it.
> 
> I like Markus's idea of `Entity.Part`, but the `Entity` class is in the `client` package. Since this class can be used on both the client and server, I'd like to avoid confusion and would probably go with `EntityPart` in the `ext` package.

Another idea would be to actually use `Entity.Part`, but to refactor it in a future release so it is getting deprecated in favor of the same class found in the `core` package. That would be straightforward.

spericas:
I'm fine with either name; name clashes won't be that common IMO. The PR LGTM.

spericas (dismissed):

mkarg:
@ronsigal @jansupol @deki WDYT?

chkal:
I'm fine with `EntityPart` as well. :+1:

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
chkal: @andymc12 Thanks for your explanation and sorry for the delayed reply.

After thinking about this more, it somehow makes sense to allow `ParamConverter` in case of `@FormParam`, as it is also used for standard `@FormParam` cases. However, I also think it could be a bit weird for the user that `@FormParam` and `getContent` may behave differently.

So basically we have three options:

  1. `@FormParam` uses the MBR and falls back to a `ParamConverter` (if available), but `getContent()` only uses the MBR.
  2. Both `@FormParam` and `getContent()` use the MBR and fall back to `ParamConverter`.
  3. Both `@FormParam` and `getContent()` only use the MBR.

To be honest, I'm not sure yet which option I personally like most.

@eclipse-ee4j/ee4j-jaxrs-committers Other opinions?

deki (approved):

jaxrs-api/src/main/java/jakarta/ws/rs/ext/EntityPart.java
```diff
@@ -0,0 +1,274 @@
...
+         * 
+         * The {@code InputStream} will be closed by the implementation code after
+         * sending the multipart data. Closing the stream before it is sent could
+         * result in unexpected behavior.
+         * 
+         * @param content {@code InputStream} of the content of this part
+         * @return the updated builder
+         * @throws IllegalArgumentException if {@code content} is {@code null}
+         */
+        public Builder content(InputStream content) throws IllegalArgumentException;
```
pdudits, line 254: I find allowing only `InputStream` based content too limiting. Content of a part can quite often be just a String -- a form field in usual multipart HTML form, or the user might want to map an in-memory object to a JSON or other format and send that. Allowing just `InputStream` would make these usacases too cumbersome:

```
String userName;
EntityPart userNamePart = EntityPart.withName("userName")
    .mediaType("text/plain;charset=UTF-8")
    .content(new ByteArrayInputStream(
        userName.getBytes(StandardCharsets.UTF_8))
    .build();
```

The API here should allow full usage of `MessageBodyWriter` capabilities of JAX-RS with InputStream being just a special case:

```
String userName;
EntityPart userNamePart = EntityPart.withName("userName")
    .mediaType("text/plain;charset=UTF-8")
    .content(userName)
   .build();
```

```
JsonObject metrics;
EntityPart metricsPart = EntityPart.withFileName("metrics.json")
    .mediaType(MediaType.APPLICATION_JSON)
    .content(metrics)
    .build();
````

File: jaxrs-api/src/main/java/jakarta/ws/rs/ext/EntityPart.java [hunk 15]
```diff
@@ -0,0 +1,274 @@
+/*******************************************************************
+* Copyright (c) 2021 Eclipse Foundation
+*
+* This specification document is made available under the terms
+* of the Eclipse Foundation Specification License v1.0, which is
+* available at https://www.eclipse.org/legal/efsl.php.
+*******************************************************************/
+package jakarta.ws.rs.ext;
```
pdudits, line 8: I believe this belongs into `.core` package rather than `.ext`, because it models a concept from RFC rather than part of processing pipieline, which is what the rest of the classes in this package represemt.

spericas: It's a good observation, but likely other things will go into `ext` (extensions) going forward. There's something to be said about keeping the core as small as possible. I believe @andymc12 (correct me if I'm wrong) considered a new package/module, but did not seem worth it.

mkarg: In fact I need to second @pdudits here. This shouldn't be `ext`, it should be `core`.

andymc12: I don't have a strong opinion either way, so I went ahead and moved `EntityPart` to the `core` package.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/EntityPart.java
andymc12: Thanks for the suggestion!  I added these two new methods:
```
public <T> Builder content(T content, Class<? extends T> type) throws IllegalArgumentException;
public <T> Builder content(T content, GenericType<T> type) throws IllegalArgumentException;
```

The type parameter should be required in order to find an applicable MBW.  For example, you might have a MBW that implements `MessageBodyWriter<MyInterface>`, but pass in an instance of `MyImpl` - or perhaps a dynamic proxy of `MyInterface`.

andymc12:
@spericas @chkal @mkarg @jansupol @deki @pdudits I've made the changes that @pdudits suggested.  Would you please take another look and approve?  Thanks!

chkal (approved):
Thanks for adjusting the PR according to the latest comments.

+1 from my side. 

However, it would be great if the other committers could have a look at [this comment](https://github.com/eclipse-ee4j/jaxrs-api/pull/948#discussion_r619644176) and share some thoughts.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/EntityPart.java
pdudits: It wouldn't hurt to serve the most usual cause of `content(entity, entity.getClass())` with a (default) method.

On the other hand, if we look at `Response.ResponseBuilder`, it offers following methods for respone content (entity):

```
        public abstract ResponseBuilder entity(Object entity);
        public abstract ResponseBuilder entity(Object entity, Annotation[] annotations);
```

Where handling of generic types is done by utilizing `GenericEntity<T>`. I believe similar content methods would provide least surprise as EntityPart would work exactly like response entity. These methods also add option for adding  explicit annotations for `MessageBodyWriter`, which would require even more overloads in current proposal to cover.

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
pdudits: Documentation of `ParamConverter` says it applies to parameters injected via (among other annotations) `@FormParam`.:

> Defines a contract for a delegate responsible for converting between a String form of a message parameter value and the corresponding custom Java type T. Conversion of message parameter values injected via `@PathParam`, `@QueryParam`, `@MatrixParam`, `@FormParam`, `@CookieParam` and `@HeaderParam` is supported. 

I am not aware of an existing API method that would utilize converters. Therefore to maintain consistency, I'd suggest staying with option 1.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/EntityPart.java
andymc12: Sounds good.  I added a default method `content(Object)` that delegates to `content(o, o.getClass())`.  Thanks for the suggestion!

andymc12:
I added one more change - adding default methods where applicable to make life more convenient for users.  @spericas @chkal @mkarg @jansupol @deki @pdudits - would you mind taking another look?  Thanks!

chkal (dismissed):
Looks good. Thanks a lot.

But I'm still hoping to get more feedback about [my comment](https://github.com/eclipse-ee4j/jaxrs-api/pull/948#discussion_r619644176) above.

jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
mkarg: > 1. `@FormParam` uses the MBR and falls back to a `ParamConverter` (if available), but `getContent()` only uses the MBR.
> 2. Both `@FormParam` and `getContent()` use the MBR and fall back to `ParamConverter`.
> 3. Both `@FormParam` and `getContent()` only use the MBR.

`getContent()` is not intended for parameters, so it shouldn't use `ParamConverter`.

My preference would be 1.

Actually after some thinking I wonder why we want `@FormParam` to use MBR, actually, as it is a parameter.

chkal: In my view `@FormParam` on types other than `EntityPart` is basically a convenient way to parse and access part data.

So instead of doing this:

```java
public Response someMethod( List<EntityPart> parts ) {

  JsonObject data = parts.stream()
    .filter( p -> p.getId().equals("file") )
    .findFirst()
    .map( p -> p.getContent( JsonObject.class ) )
    .orElse( null );

   // ...

}
```

or instead of this:

```java
public Response someMethod( @FormParam("file") EntityPart part ) {

  JsonObject data = part.getContent( JsonObject.class );
  // ...

}
```


You can simply do this:

```java
public Response someMethod( @FormParam("file") JsonObject data ) {
  // ...
}
```

And therefore it would make sense that using `@FormParam` on some specific type `T` is the same as calling `EntityPart.getContent(T.class)`.

However, I agree that this may be problematic given the originally intended usage of `ParamConverter` and `MessageBodyReader`.

spericas:
@andymc12 The latest changes look good. I think at some point we need to answer @chkal 's question about `ParamConverter`'s but, honestly, I'm not sure which option is best at this time. Perhaps we need more time using the API to determine that.

spericas (dismissed):

chkal:
> @andymc12 The latest changes look good. I think at some point we need to answer @chkal 's question about `ParamConverter`'s but, honestly, I'm not sure which option is best at this time. Perhaps we need more time using the API to determine that.

Agreed on this one. Maybe we should only allow `@FormParam` for `EntityPart` for now? We can still add support for other types later on if we want.

This would mean that code block 1 and 2 from [my comment](https://github.com/eclipse-ee4j/jaxrs-api/pull/948#discussion_r625209128) would work fine. Just code block 3 won't be possible. But IMO that would be acceptable.

andymc12:
@chkal @spericas 

> > @andymc12 The latest changes look good. I think at some point we need to answer @chkal 's question about ParamConverter's but, honestly, I'm not sure which option is best at this time. Perhaps we need more time using the API to determine that.

> Agreed on this one. Maybe we should only allow @FormParam for EntityPart for now? We can still add support for other types later on if we want.
> 

What if we also left in the ability to automatically convert to `InputStream` and `String`? i.e.
```
public Response someMethod( @FormParam("file") InputStream data ) {
  // ...
}
```

and

```
public Response someMethod( @FormParam("file") String data ) {
  // ...
}
```

This should avoid the confusion with ParamConverters, but still allow some of the more common use cases - like writing a part to disk or reading it directly as a String.  

> This would mean that code block 1 and 2 from my comment would work fine. Just code block 3 won't be possible. But IMO that would be acceptable.

That should be fine.  The only difference between your 2nd block and 3rd is one line of code.  I can live with that! :-)

spericas:
> @chkal @spericas
> 
> > > @andymc12 The latest changes look good. I think at some point we need to answer @chkal 's question about ParamConverter's but, honestly, I'm not sure which option is best at this time. Perhaps we need more time using the API to determine that.
> 
> > Agreed on this one. Maybe we should only allow @FormParam for EntityPart for now? We can still add support for other types later on if we want.
> 
> What if we also left in the ability to automatically convert to `InputStream` and `String`? i.e.
> 
> ```
> public Response someMethod( @FormParam("file") InputStream data ) {
>   // ...
> }
> ```
> 
> and
> 
> ```
> public Response someMethod( @FormParam("file") String data ) {
>   // ...
> }
> ```
> 
> This should avoid the confusion with ParamConverters, but still allow some of the more common use cases - like writing a part to disk or reading it directly as a String.
> 
> > This would mean that code block 1 and 2 from my comment would work fine. Just code block 3 won't be possible. But IMO that would be acceptable.
> 
> That should be fine. The only difference between your 2nd block and 3rd is one line of code. I can live with that! :-)

 This seems like a reasonable solution at this time; getting the param converters in the mix seems reasonable but also adds additional complexity.

andymc12:
The latest commit should address @chkal's question - basically allowing code blocks 1 & 2, but not 3.  It also allows `String` and `InputStream` as `@FormParam` types.  I believe that this was the last concern.  

@spericas @chkal @mkarg @pdudits @jansupol @deki - thanks for all of the discussion and reviews!  Can I ask that you take another look and approve?

chkal (approved):
Looks good!

@andymc12 Thanks a lot for all the adjustments. :-)

andymc12:
@spericas @mkarg @pdudits @jansupol @deki - I think we're all happy with these changes (if not, please let me know).  If one of you could approve this PR, I'll go ahead and merge.  Thanks!

deki (approved):

spericas (approved):

kalgon:
Hi, I have a simple remark: can't `EntityPart` use `javax.ws.rs.core.HttpHeaders` ?

andymc12:
> Hi, I have a simple remark: can't EntityPart use javax.ws.rs.core.HttpHeaders ?

Not directly, but you can specify headers when building an `EntityPart` or read them when consuming one - e.g.:
```
EntityPart spanishPart = EntityPart.withName("spanishData").header(HttpHeaders.CONTENT_LANGUAGE, "es")...
```

or

```
String language = entityPart.getHeader(HttpHeaders.CONTENT_LANGUAGE); 
assertEquals("es", language);
```

mkarg:
@andymc12 Great feature! Thank you for driving this!

Issue #970, opened 2021-05-28
@Consumes incorrectly handles requests with no content type header / potential security risk


likandoe:
Given the following definition:

@consumes(MediaType.APPLICATION_JSON)
@path("myresource")
@produces(MediaType.APPLICATION_JSON)
public class MyResource {

@Consumes(MediaType.APPLICATION_JSON)
@POST
@Produces(MediaType.APPLICATION_JSON)
public String login() {
    return "\n--Login!\n";
}
}

A request to "/myresource" with no 'content-type' header is accepted, login() is executed and a 200 OK response is returned. This behavior was observed with two different jaxrs-api implementations, so I assume this is a jaxrs-api issue (I also sent this issue to Jersey and was told to come here).

$ curl -v -X POST localhost:8080/myresource

Connected to localhost (127.0.0.1) port 8080 (#0)
POST /myresource HTTP/1.1
Host: localhost:8080
User-Agent: curl/7.64.0
Accept: /

< HTTP/1.1 200 OK
< Content-Type: application/json
< Content-Length: 10
<

--Login!

IMHO, the correct behavior would be to return "415 Unsupported Media Type" because the method and class are clearly annotated to handle only 'application/json'. A request with a missing 'content-type' header should not be interpreted as a valid/acceptable media type, it should be handled as an invalid media type because the lack of a content type header should be equivalent to null and therefore it is not included in the list of acceptable media types indicated by the developer (only 'application/json' in this case).

The spec mentions:

"3.5 also mentions the following:
An implementation MUST NOT invoke a method whose effective value of @consumes does not match the request Content-Type header.""

A missing content-type header (I'm referring to an HTTP Request that does not have a 'Content-Type' http header) cannot match any media type specified by the dev. However, currently, a missing content-type header is effectively matched to */*, which is not a valid content-type and it's not what the dev specified.

IMHO Even if some section of the Spec says that an HTTP Request without no content-type header is assumed to be */*, this is not correct IMHO, MIME sniffing is a security concern and secure applications are always encouraged to disable MIME sniffing (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) (side note: Is 'Content-type: */*' valid at all? 'application/*' etc yes, but '*/*' ? have you ever seen an app using it?)

[...]

PR #972, opened 2021-06-03
Clarify matching rules 

spericas:
Clarify matching rules in the case where Content-Type and/or Accept headers are absent. Prompted by discussion in issue #970 and some potential ambiguity in implementations.

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
@@ -26,3 +26,5 @@ default exception mappers.
 against the *base* URI, not against the *request* URI anymore.
 * <<context-injection>>: New section that mentions removal of `@Context`
 injection support in future versions.
+* <<declaring_method_capabilities>>: Clarified resoure matching when `Content-Type`
+or `Accept` are missing in request.
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
@@ -26,5 +26,5 @@ default exception mappers.
 against the *base* URI, not against the *request* URI anymore.
 * <<context-injection>>: New section that mentions removal of `@Context`
 injection support in future versions.
-* <<declaring_method_capabilities>>: Clarified resoure matching when `Content-Type`
+* <<declaring_method_capabilities>>: Clarified resource matching when `Content-Type`
 or `Accept` are missing in request.
--- a/jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/resources/_declaring_method_capabilities.adoc
@@ -67,9 +67,14 @@ using the `WidgetProvider` class (see <<entity_providers>> for
 more information on `MessageBodyReader`).
 
 An implementation MUST NOT invoke a method whose effective value of
-`@Produces` does not match the request `Accept` header. An
-implementation MUST NOT invoke a method whose effective value of
+`@Produces` does not match the request `Accept` header.
+If no `Accept` header is present in the request, it is assumed that
+the client accepts any media type or `\*/*`.
+
+An implementation MUST NOT invoke a method whose effective value of
 `@Consumes` does not match the request `Content-Type` header.
+If no `Content-Type` header is present in the request, the default
+media type is assumed to be `application/octet-stream`.
 
 [[selecting_from_multiple_media_types]]
 ==== Selecting from multiple media types
--- a/jaxrs-spec/src/main/asciidoc/chapters/resources/_mapping_requests_to_java_methods.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/resources/_mapping_requests_to_java_methods.adoc
@@ -214,9 +214,10 @@ For convenience, we defined latexmath:[$p \ge {\perp}$] for every media
 type latexmath:[$p$].
 +
 Given these definitions, we can now sort latexmath:[$M$] in descending
-order based on latexmath:[$\ge$] as followsfootnote:[If any of these
-types or sets of types are unspecified, latexmath:[$\mbox{*/*}$] and
-latexmath:[$\mbox{\{*/*\}}$] are assumed.]:
+order based on latexmath:[$\ge$] as followsfootnote:[If the request content type
+is missing, it is assumed to be `application/octet-stream`. If any of the other
+types or sets of types here are unspecified, then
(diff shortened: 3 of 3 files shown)


jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
```diff
@@ -26,3 +26,5 @@ default exception mappers.
 against the *base* URI, not against the *request* URI anymore.
 * <<context-injection>>: New section that mentions removal of `@Context`
 injection support in future versions.
+* <<declaring_method_capabilities>>: Clarified resoure matching when `Content-Type`
```
andymc12, line 29:
```suggestion
* <<declaring_method_capabilities>>: Clarified resource matching when `Content-Type`
```

andymc12 (commented):
Just a typo in the release notes page, but otherwise, LGTM.

andymc12 (approved):
Looks good - thanks @spericas!

chkal (dismissed):
LGTM! Thanks!

jansupol:
I believe 
`If no Content-Type header is present in the request, the default media type is assumed to be application/octet-stream` clashes with [note 7](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#_footnotedef_7): (Section 3.7.2.3.b, describing C_M): `If any of these types or sets of types are unspecified, */* and {*/*} are assumed`.

jansupol (changes requested):
-1 until the clash gets resolved. To me, this sounds like a backward incompatible change, not a clarification. This can be fixed in Algorithm 3.7.3, by making note 7 more clear (or someone can prove me being wrong), but excluding the Consumes (C_M) from note 7 is a backward incompatibility, which I am not sure should be brought in to a minor version of the Spec.

jansupol:
Please do not take me wrong, I agree with clarification, I am just pointing at the processing issues.

andymc12:
@jansupol I think that is a good catch, but I think the conflict may have already been in the spec. In [section 4.2.1](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#message_body_reader), it states that:
> 1. Obtain the media type of the request. If the request does not contain a `Content-Type` header then use `application/octet-stream`.

It's likely a stretch, but it could be that note 7 is referring to the set of applicable methods.  In that case, I think everything is fine.  The rule has always been that if a resource class/method is not annotated with `@Consumes`, then it is expected that the resource method consumes `*/*`.  

So it would resolve something like this - say we have these resource methods:
```
@POST
@Consumes("application/octet-stream")
@Path("1")
Response consumesOctetStream(SomeObject obj) {...}

@POST
@Path("2")
Response noConsumes(SomeObject obj) {...}

@POST
@Consumes("application/xml")
@Path("3")
Response consumesXml(SomeObject obj) {...}
```

If I send a POST to `/1` with no `Content-Type` header, the runtime should match it to the `consumesOctetStream` method, and process the entity body as `application/octet-stream`.

Likewise, if I send POST to `/2` with no `Content-Type` header, the runtime should match it to the `noConsumes` method, and process the entity body as `application/octet-stream`. Unlike `/1`, this method will also be matched to requests with any value for the `Content-Type` header.

If I send a POST to `/3` with no `Content-Type` header, the runtime should respond with a `415` because the missing `Content-Type` header is assumed to be `application/octet-stream` which is incompatible with `application/xml`. 

Does that make sense?  If I'm misunderstanding it, then I think we probably do need more clarification.  Even so, it might be worth clarifying that Note 7 is referring the `M` - the set of methods, not the `Content-Type` or `Accept` headers.

spericas:
> -1 until the clash gets resolved. To me, this sounds like a backward incompatible change, not a clarification. This can be fixed in Algorithm 3.7.3, by making note 7 more clear (or someone can prove me being wrong), but excluding the Consumes (C_M) from note 7 is a backward incompatibility, which I am not sure should be brought in to a minor version of the Spec.

The footnote refers to the `@Consumes` and `@Produces` parts, which may be omitted and default to `*/*` as stated elsewhere (and in our API code). It doesn't make sense to talk about wildcards as possible values for `Content-Type`. Will update the footnote to make this clear.

spericas:
> @jansupol I think that is a good catch, but I think the conflict may have already been in the spec. In [section 4.2.1](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#message_body_reader), it states that:
> 
> > 1. Obtain the media type of the request. If the request does not contain a `Content-Type` header then use `application/octet-stream`.
> 
> It's likely a stretch, but it could be that note 7 is referring to the set of applicable methods. In that case, I think everything is fine. The rule has always been that if a resource class/method is not annotated with `@Consumes`, then it is expected that the resource method consumes `*/*`.

 Yes, see my comment. Section 4.2.1 should now align better with this new clarification that we are adding. Of course, this section is about MBR's not matching.

> 
> So it would resolve something like this - say we have these resource methods:
> 
> ```
> @POST
> @Consumes("application/octet-stream")
> @Path("1")
> Response consumesOctetStream(SomeObject obj) {...}
> 
> @POST
> @Path("2")
> Response noConsumes(SomeObject obj) {...}
> 
> @POST
> @Consumes("application/xml")
> @Path("3")
> Response consumesXml(SomeObject obj) {...}
> ```
> 
> If I send a POST to `/1` with no `Content-Type` header, the runtime should match it to the `consumesOctetStream` method, and process the entity body as `application/octet-stream`.

 Exactly.

> 
> Likewise, if I send POST to `/2` with no `Content-Type` header, the runtime should match it to the `noConsumes` method, and process the entity body as `application/octet-stream`. Unlike `/1`, this method will also be matched to requests with any value for the `Content-Type` header.
> 
> If I send a POST to `/3` with no `Content-Type` header, the runtime should respond with a `415` because the missing `Content-Type` header is assumed to be `application/octet-stream` which is incompatible with `application/xml`.
> 

 Right.

jansupol:
@andymc12 What you say makes perfect sense, but the note is better to be clarified, too, if it is to be understood as you say. Section 4.2.1 discusses the message body provider selection, and Note 7 refers to resource method matching, which is strictly speaking not the same task.

spericas:
> > -1 until the clash gets resolved. To me, this sounds like a backward incompatible change, not a clarification. This can be fixed in Algorithm 3.7.3, by making note 7 more clear (or someone can prove me being wrong), but excluding the Consumes (C_M) from note 7 is a backward incompatibility, which I am not sure should be brought in to a minor version of the Spec.
> 
> The footnote refers to the `@Consumes` and `@Produces` parts, which may be omitted and default to `*/*` as stated elsewhere (and in our API code). It doesn't make sense to talk about wildcards as possible values for `Content-Type`. Will update the footnote to make this clear.

Footnote has now been updated as well.

andymc12 (approved):
Changes lgtm.  Thanks!

jansupol (approved):
Thanks!

mkarg (approved):

chkal (approved):

PR #973, opened 2021-06-07
[742] Mandate usage of JSON-B instance provided by user's ContextResolver<Jsonb>

### LINKED ISSUE #742: Mandate usage of JSON-B instance provided by user's ContextResolver<Jsonb>
Glassfish/Jersey contains a nifty feature that allows a user to specify a `ContextResolver<Jsonb>` provider.  The returned instance of `Jsonb` is then used by Glassfish to serialize and deserialize JSON objects.  That way a customer could, for instance, serialize private fields.

While this feature exists in the RI, it is not mandated by the spec, and so other JAX-RS implementations (CXF, RestEasy - and the app servers that consume them including Liberty, Thorntail/Wildfly, TomEE, etc.) do not implement this feature.

Adam Bien wrote a popular blog post on the topic here:
http://www.adam-bien.com/roller/abien/entry/private_fields_serialization_with_json

And that has caused some confusion when users discover that this feature is not portable, such as:
https://github.com/OpenLiberty/open-liberty/issues/5903

As @WhiteCat22 mentioned in his comments, there is a requirement in the spec to support a JAXBContext returned from a ContextResolver, but no such requirement for a Jsonb instance.

Adding this requirement will enhance portability.

mkarg:
Adding minimum spec text to enforce respect of `ContextResolver<Jsonb>` as commonly discussed in #742 as an initial proposal.

Please discuss and vote. :-)

Closes #742

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-3.0-release.adoc
@@ -11,6 +11,9 @@
 [[changes-since-3.0-release]]
 === Changes Since 3.0 Release
 
+* <<standard_entity_providers>>: JSON-B entity providers MUST favor
+`Jsonb` instances provided by `ContextResolver<Jsonb>` over their own
+default context.
 * <<consuming_multipart_formdata>>: Added portable API for handling
 multipart/form-data.
 * <<services>>: Added requirement that JAX-RS implementations MUST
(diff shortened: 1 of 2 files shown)


andymc12 (approved):
LGTM - thanks!

chkal (approved):
Looks great! Thanks!

spericas (approved):
This is clearly useful, but we should keep in mind that `ContextResolver` is likely going away in the next major release.

mkarg:
> This is clearly useful, but we should keep in mind that `ContextResolver` is likely going away in the next major release.

Sure, but it might need months to years until all vendor make the step from 3.1 to 4.0, so in the meantime application programmers will be happy to gain this "temporary" solution.

andymc12:
> Sure, but it might need months to years until all vendor make the step from 3.1 to 4.0, so in the meantime application programmers will be happy to gain this "temporary" solution.

+1 - and I think most of the implementations already do this - this PR is really just making it official.

PR #987, opened 2021-06-23
Remove out-of-support JDKs from CI

andymc12:
With the current LTS strategy, only Java SE versions 8, 11, and 16 are currently supported (and 16 will soon give way to 17 - 16 is not LTS, while 17 is).  As such, I'd like to reduce CI cycles by removing out-of-support JDKs from the CI pipeline.

This isn't an API/spec/TCK change, so it should be fast-tracked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,9 @@ language: java
 
 jdk:
   - oraclejdk8
-  - oraclejdk9
   - oraclejdk11
   - openjdk8
-  - openjdk9
-  - openjdk10
   - openjdk11
-  - openjdk12
-  - openjdk13
-  - openjdk14
-  - openjdk15
   - openjdk16
 
 cache:


mkarg (commented):
I understand the target and will not stand in the way of this PR, but on the other hand keeping this JDKs in the CI loop doesn't do an actual harm while there might be machines running them still, despite what is in LTS or not. So my vote is "+-0".

chkal (approved):
I'm fine with removing non-LTS. Especially because of the Travis quota. We can still create a Jenkins job to run the build against all other JDK releases if we want to. This should be easy to do and would allow us to catch incompatibles.

spericas (approved):
Thanks

andymc12:
Hmm... I can't merge because the Travis CI is hung.  I find it ironic that a PR that would reduce the Travis CI cost is prevented from merging because of the Travis CI cost.  :)

mkarg:
> Hmm... I can't merge because the Travis CI is hung. I find it ironic that a PR that would reduce the Travis CI cost is prevented from merging because of the Travis CI cost. :)

Actually there is no Travis cost. Travis is used on a free tier. The problem is that every user has 10.000 free credits per month, which would be plenty enough for us, but the EF admins decided that ALL projects of the EF share a SINGLE user. If they would simply use one user per project (as offered by Travis) there would be no problem at all.

PR #976, opened 2021-06-13
[742] TCK Tests for ContextResolver<Jsonb>

mkarg:
This PR is amending the our new TCK by tests for `ContextResolver<Jsonb>`.

See https://github.com/eclipse-ee4j/jaxrs-api/pull/973 for the according spec change.

Issue #10, opened 2009-03-10
Improve handling of trailing slash


glassfishrobot:
As per the current spec trailing slashes are ignored for the selection of root
resources and resource methods.

This is odd, because it gives effectively two names to the HTTP-resources made
available but it also has a significant practical disadvantage: A resource
method that generates relative links has to have the UriInfo injected and check
for the presence/absence of trailing slash and either adapt the relative links
or send a redirect. It would be good if implementations would free applications
from this burden.

See also: mail-thread from 2009-01-06 on users@jsr311.dev.java.net "trailing slash"
#### Environment
Operating System: All
Platform: All
#### Affected Versions
[2.1]

PR #988, opened 2021-06-24
Initial commit of arquillian-based TCK

andymc12:
This is a work in progress.  The plan is to create a TCK artifact using JUnit 5 and Arquillian that vendors can run to verify compliance. At this point, it only includes a very basic test case.  Once this is working, then I'd like to start migrating tests from the platform TCK here.

### DIFF
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -117,10 +117,5 @@
             <version>${arquillian-bom.version}</version>
         </dependency>
 
-        <dependency>
-            <groupId>com.squareup.okhttp3</groupId>
-            <artifactId>okhttp</artifactId>
-            <version>4.9.1</version>
-        </dependency>
     </dependencies>
 </project>
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -31,6 +31,8 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
+        <arquillian-bom.version>1.7.0.Alpha9</arquillian-bom.version>
+        <junit-jupiter.version>5.7.2</junit-jupiter.version>
     </properties>
 
     <organization>
@@ -78,6 +80,18 @@
         <tag>HEAD</tag>
     </scm>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>5.7.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
     <dependencies>
         <dependency>
             <groupId>jakarta.ws.rs</groupId>
@@ -95,6 +109,18 @@
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-library</artifactId>
             <version>[2.2, 2.3-A00)</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.arquillian.junit5</groupId>
+            <artifactId>arquillian-junit5-container</artifactId>
+            <version>${arquillian-bom.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>4.9.1</version>
         </dependency>
     </dependencies>
 </project>
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -31,8 +31,9 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
-        <arquillian-bom.version>1.7.0.Alpha9</arquillian-bom.version>
+        <arquillian-bom.version>1.7.0.Alpha10</arquillian-bom.version>
         <junit-jupiter.version>5.7.2</junit-jupiter.version>
+        <hamcrest-version>2.2</hamcrest-version>
     </properties>
 
     <organization>
@@ -85,7 +86,7 @@
             <dependency>
                 <groupId>org.junit</groupId>
                 <artifactId>junit-bom</artifactId>
-                <version>5.7.0</version>
+                <version>${junit-jupiter.version}</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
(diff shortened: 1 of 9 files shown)


jaxrs-tck/pom.xml
```diff
@@ -78,6 +80,18 @@
         <tag>HEAD</tag>
     </scm>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>5.7.0</version>
```
mkarg, line 88: Do you deliberately have different version numbers in lines 35 and 88?

jaxrs-tck/pom.xml
```diff
@@ -96,5 +110,17 @@
...
         </dependency>
+
+        <dependency>
+            <groupId>org.jboss.arquillian.junit5</groupId>
+            <artifactId>arquillian-junit5-container</artifactId>
+            <version>${arquillian-bom.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
```
mkarg, line 121: We might need a "uses" CQ for this. What is it doing?

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/Util.java
```diff
@@ -0,0 +1,13 @@
+package jakarta.ws.rs.tck;
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+public class Util {
+    public static final int HTTP_PORT = Integer.getInteger("http.port");
```
mkarg, line 7: We should default to zero, which makes JRE discover a free port.

mkarg (commented):
Cool stuff! 8-)

mkarg:
Good starting point, but how to see this code in action? Can you put an example ontop that is actually executable using `mvn verify`? Thanks! :-)

jaxrs-tck/pom.xml
andymc12: No, I copied that from other TCK examples.  I think we'll probably want to use the latest level, but we've got to get it working first. :)

jaxrs-tck/pom.xml
andymc12: It's just a simple HTTP client.

> Good starting point, but how to see this code in action? Can you put an example ontop that is actually executable using `mvn verify`? Thanks! :-)

I have this configured to run with Open Liberty and GlassFish (depending on profile, `-Parq-liberty-managed` or `-Parq-glassfish-managed`)  at https://github.com/andymc12/jakartaRest31TCK-openliberty 

The problem is that although the test framework claims that the tests run and pass, they aren't actually executed... I'm investigating this issue further with some folks over slack at: https://eclipsefoundationhq.slack.com/archives/C0131MLD538/p1624542758005000

If we don't make progress soon, I may give TestNG a try.

jaxrs-tck/pom.xml
```diff
@@ -31,6 +31,8 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
+        <arquillian-bom.version>1.7.0.Alpha9</arquillian-bom.version>
```
mkarg, line 34: The is now Beta #10 (https://search.maven.org/artifact/org.jboss.arquillian.junit5/arquillian-junit5-container/1.7.0.Alpha10/jar) since June 21.

mkarg:
@OndroMih can you please help us make this work?

OndroMih:
Hi @mkarg,

> Good starting point, but how to see this code in action? Can you put an example ontop that is actually executable using mvn verify? Thanks! :-)

I used @andymc12's project at this specific commit to run the TCK: https://github.com/andymc12/jakartaRest31TCK-openliberty/commit/855c998422c8c46413170f271d49a4958597d3f0 (the master branch now has some commits that replace JUnit5 with TestNG). I ran it against a remote GF 6, with the following command: `-Dhttp.port=8080 -Parq-glassfish-remote verify`

When I debugged the test suite, I found out that the okhttp classes raised ClassNotFound exception in the Arquilllian container. The tests weren't executed, they were ignored, instead of reporting them as failed test. This is probably a bug in the Arquillian connector. To fix this, the missing classes need to be added to the deployment. Or, even better, the methods which use okhttp can be executed on the client, outside of the Arquillian container, by adding `@RunAsClient` next to `@Test` annotation, because they don't need to run in the container.

andymc12:
Thanks so much @OndroMih!!  Likewise, by adding `@RunAsClient`, the tests ran for me too.  I've updated this PR to use that annotation on the test class - as well as replaced OkHttp with the Java 11 HTTP client. With these updates, everything works for me in Open Liberty.  Since the preference of the group was to use JUnit 5 over TestNG, I'd like to close #998 in favor of this PR.  

@spericas @mkarg @chkal - can you review this PR ASAP?  I'd like to get it in (using fast-track) so that we can then start porting over the 3.0 tests and begin adding 3.1-specific tests.  Thanks!!

jaxrs-tck/pom.xml
mkarg: As Andy replaces OKHTTP3 by java.net.HttpClient, I am closing this thread.

mkarg:
@OndroMih Thanks a lot, you saved our day! :-)

mkarg:
Kudos! In general this looks good to my. Sole point is that maybe it would be more nice if you do not clutter `master` with fake tests in the first place, but instead start with real tests right from the start? Also I think it would be cool that the actual pom needed to run the TCK against the real tests would be similarly short as the one from SE: Just pulling in the TCK plus the product as dependencies, and let the TCK do the rest?

chkal (approved):
Thanks, @andymc12! Basically looks good to me. I've some ideas for improvement of the general structure, but let's merge this one soon and improve afterwards.

mkarg:
@andymc12 Can you please rename the test class so that it ends on `IT` as the others (not on `Test` like a unit test)? By doing so, one can simply run the TCK suite using `mvn -am -pl :jersey-tck verify` for example (well, not really actually, as the pom.xml of that project does not yet have a dependency on a arquillian-jersey-adapter, but I think you understand what I like to reach)?

Also I would kindly like to ask to squash all commits into a single one to prevent cluttering the `master`'s log. Your work is great, but in the end it only adds a *single* value: Arquillian.

Last but not least, maybe you would like to rebase ontop of https://github.com/eclipse-ee4j/jaxrs-api/pull/976 and replace the sample code by my actual test? Doing so would turn your sample PR into a really needed TCK test, and I could close #976 in turn, so we have a shared PR #988 with my commit first (without Arquillian) and your commit ontop, simply turning my test into an Arquillian test. That resolves two PRs in one turn, and prevents cluttering the `master` with temporary sample test code. :-)

Issue #989, opened 2021-07-06
Add required apis to serialize the EntityPart


jimma:
At the moment there is no api to get the object which is added to the EntityPart to serialize the entity to output stream. 
It also needs getType() and getGenericType() to select the proper MessageBodyWriter.

andymc12:
Hi @jimma  - the `EntityPart.Builder` class is used for the multipart sender to specify the object to send and the type/generic type of that object.  The `EntityPart` is intended to represent the object as a part of a multipart request to the user - it would be analogous to the `Response` object.  So the user should not need to get the object directly (they would need to go through message body readers - which would require that they specify the type/generic type - this is handled by `getContent` methods with [Class](https://github.com/eclipse-ee4j/jaxrs-api/blob/3afb3fcacf7faf7a93a5ac45ab6a89560227c04c/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java#L151) and [GenericType](https://github.com/eclipse-ee4j/jaxrs-api/blob/3afb3fcacf7faf7a93a5ac45ab6a89560227c04c/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java#L187) as parameters).

For implementors, they would need to have access to the type/generic type and object that the user created using the `Builder` so that they could serialize the part in the HTTP stream - but for that, I would expect that implementations would create their own implementation class of `Builder` and `EntityPart` that would provide them with access to the type and object specified by the user.

Hope this helps.

jimma:
Hi @andymc12  Thanks for your quick response. 

> For implementors, they would need to have access to the type/generic type and object that the user created using the Builder so that they could serialize the part in the HTTP stream - but for that, I would expect that implementations would create their own implementation class of Builder and EntityPart that would provide them with access to the type and object specified by the user.

Yes. Implementor can provide its own access the type and object.  But I see adding these apis will be helpful for  user to get this object  of this Entity before sending out for debug or test purpose, and for implementor these apis are very convenient to select proper message writer to serialize this object. 

```
List<EntityPart> parts = Arrays.asList( 
EntityPart.withName("text").content("plan text".mediaType(MediaType.TEXT_PLAIN_TYPE).build(),
EntityPart.withName("file").content(fileInputStream).mediaType(MediaType.TEXT_PLAIN_TYPE).build());

```

andymc12:
[...]

PR #990, opened 2021-07-06
[#989]:Returns object for getContent() method

jimma:


spericas (changes requested):
See `getContent` methods.

jansupol:
The issue with these `getType` and `getGenericType` methods is that `EntityPart` used on the "incoming" side does not have the type and generic type information available after receiving the wired data. Perhaps a better solution would be to use GenericEntity as content? I assume that these methods are a reaction for `EntityPart.Builder content(T content, GenericType<T> type)`, perhaps the implementation can compose the `GenericEntity` from `content` and `type` if `content` is not a `GenericEntity` already?

spericas:
Closing after not getting the necessary approvals.

jimma:
> The issue with these `getType` and `getGenericType` methods is that `EntityPart` used on the "incoming" side does not have the type and generic type information available after receiving the wired data. Perhaps a better solution would be to use GenericEntity as content? I assume that these methods are a reaction for `EntityPart.Builder content(T content, GenericType<T> type)`, perhaps the implementation can compose the `GenericEntity` from `content` and `type` if `content` is not a `GenericEntity` yet?

@jansupol  Thanks for your input. My proposal is when we add some simple content like : https://github.com/eclipse-ee4j/jaxrs-api/issues/989#issuecomment-876247852, we need  a simple 'getContent' method to read it. When we set a string content, it has to wrap the string content to a InputStream result for 'getConent()' now. Is it better it simply returns a 'Object' ?

jansupol:
@jimma This pull request introduces two independent things:
1) `getType()` & `getEntityType()` - see my comment
2) `Object getContent()`  - You can use [getContent(String.class)](https://github.com/eclipse-ee4j/jaxrs-api/blob/eb2891d33f8dfbde575b7f87e9abe823c7fe4be8/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java#L156), can you not?

jimma:
@jansupol  Thanks for the input again.

> Now the discussion seems to be about a better return type or better convenient method. In your example, you use a client filter that indeed has the object. But the main usage of the method is IMO on the receiving side, where you do not have the deserialized object, you have the InputStream data.  There, the javadoc describes the behavior. Either you get the InputStream, or you need to use the message body provider to deserialize the input stream data to another java type.

 Having a better return type like ```object```, this will work for both sides. If I understand correctly, the EntityPart is for sending and receiving side, it would be good if we define a general return type instead of only for receiving side.

jimma:
@jansupol  I changed this PR to add the `getContentStream()` and return `Object` for `getContent`.  Please review. Thanks.

jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java
```diff
@@ -168,8 +179,8 @@ static Builder withFileName(String partAndFileName) {
      * The implementation is required to close the content stream when this method
      * is invoked, so it may only be invoked once. Subsequent invocations will
      * result in an {@code IllegalStateException}. Likewise this method will throw
-     * an {@code IllegalStateException} if it is called after calling
-     * {@link #getContent} or {@link #getContent(Class)}.
+     * an {@code IllegalStateException} if it is called after if it is called after {@code InputStream}
+     * fully consumed already and has not been buffered prior consuming.
```
andymc12, line 183:
```suggestion
     * an {@code IllegalStateException} if it is called after the {@code InputStream} has been
     * fully consumed already and has not been buffered prior to consuming.
```

jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java
```diff
@@ -127,8 +140,8 @@ static Builder withFileName(String partAndFileName) {
      * The implementation is required to close the content stream when this method
      * is invoked, so it may only be invoked once. Subsequent invocations will
      * result in an {@code IllegalStateException}. Likewise this method will throw
-     * an {@code IllegalStateException} if it is called after calling
-     * {@link #getContent} or {@link #getContent(GenericType)}.
+     * an {@code IllegalStateException} if it is called after {@code InputStream}
+     * fully consumed already and has not been buffered prior consuming.
```
andymc12, line 144:
```suggestion
     * an {@code IllegalStateException} if it is called after the {@code InputStream} has
     * been fully consumed already and has not been buffered prior to consuming.
```

jimma:
@jansupol  @spericas @andymc12  Can you please review this again ?

jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java
```diff
@@ -104,6 +104,18 @@ static Builder withFileName(String partAndFileName) {
      */
     Optional<String> getFileName();
 
+    /**
+     * Get the content object from this part.
+     * <p>
+     * If the content is represented by an un-consumed {@link InputStream InputStream} the method will return the input
+     * stream
+     * </p>
```
andymc12, line 112: The text here seems too vague to me, and I worry that it might confuse users.  IIUC, the intent of the `getContent` method is:
(1) on the _sending_ side, this method returns the actual object (`String`, `File`, `Widget`, `byte[]`, etc.) that was passed in to the EntityPart's builder.
(2) on the _receiving_ side, this method is basically the same as `getContentStream`, but requires the user to cast it.

I think this method really only has any value on the sending side.  Maybe we could change this to just throw an `IllegalStateException` on the receiving side?

Alternatively, we could view the EntityPart as a "holder" for the entity object on both the sending and receiving sides - then it should always return the entity object (`String`, `Widget`, etc.) but only after a (not yet written) `bufferContent()` is invoked followed by `getContent(Class)` - similar to `Response#bufferEntity()` and `Response#readEntity(...)`.

jimma: Thanks for review @andymc12 
> I think this method really only has any value on the sending side. Maybe we could change this to just throw an IllegalStateException on the receiving side?

Do you mean this method is only for sending side and calling this method throws IllegalStateException on the receiving side ? 

> Alternatively, we could view the EntityPart as a "holder" for the entity object on both the sending and receiving sides - then it should always return the entity object (String, Widget, etc.) but only after a (not yet written) bufferContent() is invoked followed by getContent(Class) - similar to Response#bufferEntity() and Response#readEntity(...).

+1 . I think this is a good idea.  Should we still keep 'getContentStream()' ?

spericas:
@jimma @andymc12 @jansupol Given where we are in the release process, I'd suggest we revisit this post 3.1.

andymc12:
@spericas I'm fine with moving this to 4.0.  Thanks.

jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java
```diff
@@ -104,6 +104,18 @@ static Builder withFileName(String partAndFileName) {
...
+     *
+     * @return the an content of this part
```
mkarg, line 114: Typo "the an"

mkarg (changes requested):

mkarg:
@jakartaee/ee4j-jaxrs-committers How to proceed?

spericas:
Closing given that it was never approved

PR #991, opened 2021-07-06
Cleanup of pom names and relative paths to avoid warnings

spericas:
Cleanup of pom names, and changes to relative paths to avoid warnings when building modules. Some minor updates to Javadocs to avoid warnings as well.

Requesting fast track.

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java
@@ -131,6 +131,7 @@ public interface EntityPart {
      * 
      * @param type the {@code Class} that the implementation should convert this
      *             part to
+     * @param <T> the entity type
      * @return an instance of the specified {@code Class} representing the content
      *         of this part
      * @throws IllegalArgumentException if no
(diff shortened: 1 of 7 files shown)


mkarg (dismissed):

andymc12 (approved):

jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java
```diff
@@ -211,7 +213,7 @@ static Builder withFileName(String partAndFileName) {
      * 
      * @since 3.1
      */
-    public interface Builder {
```
mkarg, line 214: Why removing `public`?

andymc12 (dismissed):
LGTM - thanks!

jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java
spericas: It's redundant for inner interfaces, as it is for methods etc. in all interfaces.

chkal (approved):
Thanks! But it looks like there are merge conflicts in one of the pom.xml files.

andymc12 (approved):

chkal (approved):

PR #997, opened 2021-07-08
Removing inactive committers

mkarg:
This section of the spec is intended to provide a reference whom to call with question of the *current* spec and API revision, so it makes no sense to keep people listet not having contributed in any noticeable form *for many months*.

Either we agree on *frequent* updates to this list in the spec, or we replace the names with a link to the Github statistics.


andymc12 (approved):
I checked with @sdaschner on Twitter, and he said that he will likely not be able to contribute to the project for the next few weeks or months - so removing him from the current committers list is reasonable.

chkal (approved):

mkarg:
This PR was open for review for 18 days, which is 4 days more than needed. The result is +3 (+2 and my implied +1 as the PR author), zero -1, zero 0. Hence I hereby merge this change. Thanks for voting, everybody.

PR #998, opened 2021-07-08
Initial check-in of TCK w/Arquillian+TestNG

andymc12:
This PR creates the Arquillian-based TCK. It is based on TestNG because I could not get JUnit 5 to work (it always claimed success on the test runs, but didn't actually run the tests).  FWIW, MicroProfile uses TestNG with Arquillian - very similar to what I have proposed here.

I have some more doc on how a vendor would run these tests here: https://github.com/andymc12/jakartaRest31TCK-openliberty - this will only work locally after building/installing the `jaxrs-tck`. 

This PR includes two very simple test cases (basic GET and POST).  Future PRs should migrate the existing Jakarta REST 3.0 TCK tests as well as create new tests for some of the 3.1-specific features.

Question: Do we want to convert @mkarg's Java SE based tests to use TestNG? Or should it stick with JUnit 5 since JUnit 5 works if we don't need Arquillian? I'm willing to convert the tests in this PR if we have consensus - or we can do it later in another PR.  Another option might be to stick with JUnit 5 for the SE tests and eventually move the Arquillian tests to JUnit 5 if we can resolve the test runtime issues.

### DIFF
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -86,12 +86,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
-            <version>[5.5.2, 5.6-A00)</version>
-        </dependency>
-
-        <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-library</artifactId>
             <version>[2.2, 2.3-A00)</version>
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!--
 
-    Copyright (c) 2020 Markus Karg. All rights reserved.
+    Copyright (c) 2020, 2021 Markus Karg. All rights reserved.
 
     This program and the accompanying materials are made available under the
     terms of the Eclipse Public License v. 2.0, which is available at
@@ -29,8 +29,8 @@
     <url>https://github.com/eclipse-ee4j/jaxrs-api</url>
 
     <properties>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
     </properties>
 
     <organization>
@@ -95,6 +95,27 @@
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-library</artifactId>
             <version>[2.2, 2.3-A00)</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.testng</groupId>
+            <artifactId>testng</artifactId>
+            <version>7.4.0</version>
+            <scope>compile</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.arquillian</groupId>
+            <artifactId>arquillian-bom</artifactId>
+            <version>1.6.0.Final</version>
+            <scope>import</scope>
+            <type>pom</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.arquillian.testng</groupId>
+            <artifactId>arquillian-testng-container</artifactId>
+            <version>1.6.0.Final</version>
         </dependency>
     </dependencies>
 </project>
--- a/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/BasicResourceTest.java
+++ b/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/BasicResourceTest.java
@@ -36,7 +36,7 @@ public class BasicResourceTest extends Arquillian {
     }
 
     @Test
-    public void test_get() throws Exception {
+    public void get() throws Exception {
         String uri = Util.getUrlFor(CONTEXT_ROOT, "app", "basic", "123");
         LOG.finest(() -> "uri = " + uri);
         HttpClient client = HttpClient.newHttpClient();
@@ -49,7 +49,7 @@ public class BasicResourceTest extends Arquillian {
     }
 
     @Test
-    public void test_post() throws Exception {
+    public void post() throws Exception {
         String uri = Util.getUrlFor(CONTEXT_ROOT, "app", "basic");
         LOG.finest(() -> "uri = " + uri);
         HttpClient client = HttpClient.newHttpClient();
--- a/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/BasicResourceTest.java
+++ b/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/BasicResourceTest.java
@@ -1,0 +1,64 @@
+/*******************************************************************
+* Copyright (c) 2021 Eclipse Foundation
+*
+* This specification document is made available under the terms
+* of the Eclipse Foundation Specification License v1.0, which is
+* available at https://www.eclipse.org/legal/efsl.php.
+*******************************************************************/
+package jakarta.ws.rs.tck.basic;
+
+import static org.testng.Assert.assertEquals;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpClient.Version;
+import java.net.http.HttpRequest;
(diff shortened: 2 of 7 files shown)


jaxrs-tck/pom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <!--
 
-    Copyright (c) 2020 Markus Karg. All rights reserved.
+    Copyright (c) 2020, 2021 Markus Karg. All rights reserved.
```
mkarg, line 4: The EF allows us to use the short syntax with a single date only. That would spare us from the clutter of frequently changed header lines. I would really love it if we could all agree that as it is not enforced by the EF we do stop using the two-dates variant.

jaxrs-tck/pom.xml
```diff
@@ -29,8 +29,8 @@
     <url>https://github.com/eclipse-ee4j/jaxrs-api</url>
 
     <properties>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>11</maven.compiler.source>
```
mkarg, line 32: I would prefer moving to Java 11 explicitly in a second PR, and do that switch for all our code then.

mkarg (commented):
I personally would prefer sticking with JUnit 5 but ask for help at the Arquilian team, as clearly one day there will be a working solution with JUnit 5 and it simply is the test tool preferred by most people - so we might want to switch back rather soo... But I do not stand in the way if a majority prefers to switch *really everything* (not just single tests) to TestNG. Using a mix of JUnit and TestNG is something I rather would dislike.

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/BasicResourceTest.java
```diff
@@ -0,0 +1,64 @@
...
+    private final static String CONTEXT_ROOT = "basic";
+
+    @Deployment
+    public static WebArchive createDeployment() {
+        return ShrinkWrap.create(WebArchive.class, "basic.war")
+                         .addPackage(BasicApp.class.getPackage());
+    }
+
+    @Test
+    public void test_get() throws Exception {
```
jansupol, line 39: This does not look like java: https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html.

jaxrs-tck/pom.xml
andymc12: The main requirement here is that I am using the Java 11 HTTP Client for testing, so the change would need to be done either in this PR or in a PR that is merged before mine.

jaxrs-tck/pom.xml
andymc12: I'll defer this to somebody with more IP/Legal experience. Personally, I'd prefer that we didn't need to put or update a copyright date at all, but I think the Eclipse Foundation requires it.

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/BasicResourceTest.java
andymc12: I have no problems changing the method names to `testGet` or even just `get` (since we know we're in a test method).

andymc12 (commented):
> Using a mix of JUnit and TestNG is something I rather would dislike.
100% agree.

jaxrs-tck/pom.xml
mkarg: The EF's rules are clear and publicly available: They demand ONLY the initial year, but do NOT require the second, updated year. This IS the official EF statement. So if nobody wants to have it updated, the EF is fine if we stick with the original single-year variant.

jaxrs-tck/pom.xml
mkarg: The EF's rules are clear and publicly available: They demand ONLY the initial year, but do NOT require the second, updated year. This IS the official EF statement. So if nobody wants to have it updated, the EF is fine if we stick with the original single-year variant.

andymc12:
@mkarg @jansupol I think I've addressed the first round of comments. Want to take a second pass?

@chkal @spericas can you take a look?  I'd also advise checking out my sample project at https://github.com/andymc12/jakartaRest31TCK-openliberty - this demonstrates how a vendor would run the TCK to verify compatibility.

I'd like to get this in ASAP so that we can start moving the 3.0 tests from the EE TCK umbrella project to here for the 3.1 release.  What do you think about fast-tracking this one?  It's not API, Javadoc or spec but the TCK was not part of this project when we discussed the minimum length rules.

mkarg:
I think we can fast track *this* one, but in general the two-weeks rule should be applied to TCK changes in future, too.

mkarg (commented):
I'd like to try this out locally. Is there an implementation-beta that will actually pass the SE tests?

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/Util.java
```diff
@@ -0,0 +1,20 @@
...
+* of the Eclipse Foundation Specification License v1.0, which is
+* available at https://www.eclipse.org/legal/efsl.php.
+*******************************************************************/
+package jakarta.ws.rs.tck.basic;
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+public class Util {
+    public static final int HTTP_PORT = Integer.getInteger("http.port", 8010);
```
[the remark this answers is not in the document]
andymc12: Sorry for the confusion on this - that port should be set by the vendor to tell the test framework what port its server is listening on.  Since the server port is controlled outside of the TCK framework (including Arquillian, etc.) it is the responsibility of the vendor to tell the framework what port it is using.  I was using port 8010 which is why I used that as the default.  I could change that be a more widely-used default port (like 8080), but defaulting to zero won't work.

andymc12:
> I'd like to try this out locally. Is there an implementation-beta that will actually pass the SE tests?

Open Liberty does not implement the SE tests, so you'd probably need to take my [example repo](https://github.com/andymc12/jakartaRest31TCK-openliberty) and re-fit it with GlassFish (I'm assuming GF pulls in the correct version of Jersey that contains the SE implementation).  HTH

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/Util.java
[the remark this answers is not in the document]
mkarg: Isn't there a way the TCK could force a specific port to be used by the vendor's server?

The Java SE Bootstrap is really beneficial here. Too bad that we don't force all servers to support it.

mkarg:
> > I'd like to try this out locally. Is there an implementation-beta that will actually pass the SE tests?
> 
> Open Liberty does not implement the SE tests, so you'd probably need to take my [example repo](https://github.com/andymc12/jakartaRest31TCK-openliberty) and re-fit it with GlassFish (I'm assuming GF pulls in the correct version of Jersey that contains the SE implementation). HTH

Sorry, actually didn't mean the SE tests, but the Arquillian-based tests.

andymc12:
> > >I'd like to try this out locally. Is there an implementation-beta that will actually pass the SE tests?
> >
> >Open Liberty does not implement the SE tests, so you'd probably need to take my example repo and re-fit it with GlassFish (I'm assuming GF pulls in the correct version of Jersey that contains the SE implementation). HTH
>
> Sorry, actually didn't mean the SE tests, but the Arquillian-based tests.

Cool, in that case, I think you should be able to use my example as-is.  First, you'll need to check out my branch and build/install it.  Then you can run the tests from my repo.  The "Basic" tests intentionally don't use any 3.1 APIs so that it will pass on Liberty's 3.0 implementation. If you have any questions or run into any problems, please let me know. Thanks!

andymc12:
Closing this PR, as JUnit 5 is now working - see PR #988.

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/basic/Util.java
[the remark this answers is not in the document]
spericas: Using 8080 as default may be a better option, but otherwise LGTM.

spericas (approved):

andymc12:
sorry @spericas - I meant to close this one, but apparently just commented... can you review #988 instead?  Thanks

PR #1005, opened 2021-08-03
[1004] Wrap service loader calls in privileged actions if the securit…

### LINKED ISSUE #1004: Partial implementation of privileged actions should be completed
In the various `FactoryFinder`'s there is a privileged action used to get the context class loader. However, there is no such privileged action around the service loader which could fail if the security manager is enabled and read access is not explicitly granted to the implementation JAR.

The `ClientBuilder` could also fail with a security issue while attempting to throw a `LinkageError` if the security manager is enabled and the correct permissions are not defined.

jamezp:
…y manager is enabled. Also ensure the LinkageError for the ClientBuilder is thrown if a security manager is enabled.

Resolves #1004

Signed-off-by: James R. Perkins <jperkins@redhat.com>

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
@@ -19,7 +19,9 @@ package jakarta.ws.rs.client;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import java.net.URL;
+import java.security.AccessController;
 import java.security.KeyStore;
+import java.security.PrivilegedAction;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -59,16 +61,14 @@ public abstract class ClientBuilder implements Configurable<ClientBuilder> {
         try {
             Object delegate = FactoryFinder.find(JAXRS_DEFAULT_CLIENT_BUILDER_PROPERTY, ClientBuilder.class);
             if (!(delegate instanceof ClientBuilder)) {
-                Class pClass = ClientBuilder.class;
-                String classnameAsResource = pClass.getName().replace('.', '/') + ".class";
-                ClassLoader loader = pClass.getClassLoader();
-                if (loader == null) {
-                    loader = ClassLoader.getSystemClassLoader();
+                final CreateErrorMessageAction action = new CreateErrorMessageAction(delegate);
+                final String errorMessage;
+                if (System.getSecurityManager() == null) {
+                    errorMessage = action.run();
+                } else {
+                    errorMessage = AccessController.doPrivileged(action);
                 }
-                URL targetTypeURL = loader.getResource(classnameAsResource);
-                throw new LinkageError("ClassCastException: attempting to cast"
-                        + delegate.getClass().getClassLoader().getResource(classnameAsResource)
-                        + " to " + targetTypeURL);
+                throw new LinkageError(errorMessage);
             }
             return (ClientBuilder) delegate;
         } catch (Exception ex) {
@@ -271,4 +271,26 @@ public abstract class ClientBuilder implements Configurable<ClientBuilder> {
      * @return a new client instance.
      */
     public abstract Client build();
+
+    private static class CreateErrorMessageAction implements PrivilegedAction<String> {
+        private final Object delegate;
+
+        private CreateErrorMessageAction(final Object delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public String run() {
+            Class<?> pClass = ClientBuilder.class;
+            String classnameAsResource = pClass.getName().replace('.', '/') + ".class";
+            ClassLoader loader = pClass.getClassLoader();
+            if (loader == null) {
+                loader = ClassLoader.getSystemClassLoader();
+            }
+            URL targetTypeURL = loader.getResource(classnameAsResource);
+            return "ClassCastException: attempting to cast"
+                    + delegate.getClass().getClassLoader().getResource(classnameAsResource)
+                    + " to " + targetTypeURL;
+        }
+    }
 }
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/FactoryFinder.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/FactoryFinder.java
@@ -21,9 +21,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.Iterator;
 import java.util.Properties;
-import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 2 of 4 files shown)


jamezp:
I forgot to mention, if this seems acceptable I can port the changes over to whichever branches desired.

arjantijms:
@jamezp do note that `SecurityManager`, `AccessController` etc  have been deprecated for removal in JDK 17. So we may want to remove such privilege actions, not add them.

jansupol:
There is quite a lot of code duplication in the world of suppliers.

[JEP 411](https://openjdk.java.net/jeps/411) deprecates the SecurityManager in JDK 17, but for JDK 17, it means a warning is issued. Jakarta REST 3.1 is supposed to be working with JDK 11-17 which still has the SecurityManager functional. But I'm surprised noone complained about this untill now (2.0 was released with EE 7) 

@jamezp You seem to use the words may/might, is this a real issue or the could-potentially-happen one?

jamezp:
> @jamezp do note that `SecurityManager`, `AccessController` etc have been deprecated for removal in JDK 17. So we may want to remove such privilege actions, not add them.

Yes, I thought about that before sending this. However, there still is no clear path for a replacement that I'm aware of. I know that some installations require the security manager to be enabled.

The real purpose of this is so RESTEasy no longer has to use a fork of the spec :)

jamezp:
> There is quite a lot of code duplication in the world of suppliers.
> 
> [JEP 411](https://openjdk.java.net/jeps/411) deprecates the SecurityManager in JDK 17, but for JDK 17, it means a warning is issued. Jakarta REST 3.1 is supposed to be working with JDK 11-17 which still has the SecurityManager functional. But I'm surprised noone complained about this untill now (2.0 was released with EE 7)

Yes. AFAIK there is no replacement for the security manager at this point. I have a feeling it won't be removed anytime soon :)

> 
> @jamezp You seem to use the words may/might, is this a real issue or the could-potentially-happen one?

Sorry, yes this is a real issue. I've not tried in other application servers, but I can say in WildFly if you run with a security manager then there are failures. There is a fix in a [fork](https://github.com/jboss/jboss-jakarta-jaxrs-api_spec) which WildFly and JBoss EAP use. However it would be nice to let the fork die and use the spec itself :)

mkarg:
I'd rather vote for removal of security manager instead of correcting it.

spericas:
@jamezp What is the use case for running WildFly under a security manager? The JDK team seems to simply by JEP 411 that this is no longer needed/used in server-side Java.

andymc12:
I know that we (IBM) have customers who require Java 2 security for their app servers - even in Java/Jakarta EE.  Personally, I've never been a fan of Java 2 security, but we do have some users who need it.  I believe most of our Java 2 security customers are governmental users or are contractors for governments that require it. Those requirements might start to be lifted now that Java 17 is deprecating the security manager, but in general (and until we get a little closer to prereq'ing Java 17), I'm in favor of adding doPrivs where necessary.

jamezp:
> @jamezp What is the use case for running WildFly under a security manager? The JDK team seems to simply by JEP 411 that this is no longer needed/used in server-side Java.

@spericas I can't say I know the requirements specifically for WildFly, but I know we (Red Hat) have JBoss EAP customers that are required to run the container with the security manager enabled.

I agree with @andymc12 that I don't like it, but I don't really have a choice but to support it :)

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
```diff
@@ -59,16 +61,31 @@ public static ClientBuilder newBuilder() {
...
+                    if (loader == null) {
+                        loader = ClassLoader.getSystemClassLoader();
+                    }
+                    URL targetTypeURL = loader.getResource(classnameAsResource);
+                    errorMessage = "ClassCastException: attempting to cast"
+                            + delegate.getClass().getClassLoader().getResource(classnameAsResource)
+                            + " to " + targetTypeURL;
+                } else {
+                    errorMessage = AccessController.doPrivileged((PrivilegedAction<String>) () -> {
+                        ClassLoader loader = pClass.getClassLoader();
```
mkarg, line 78: Maybe we should refactor into private method instead of duplicating code.

spericas:
@jamezp @andymc12 Just reviewed JEP 411. Looks to me that it would be OK to fix this for 3.1. We should probably review this for 4.0 as clearly a lot of this functionality is going away. I still think that the "security" provided by the Java Security Manager is a lot less useful in practice than on paper.

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
jamezp: Should be done if I've understood correctly :)

spericas (dismissed):
LGTM. Only comment is that `findFirstService`'s could create an `Action` to avoid duplicating code blocks.

jaxrs-api/src/main/java/jakarta/ws/rs/client/FactoryFinder.java
```diff
@@ -168,4 +163,44 @@ private static Object newInstance(final String className, final ClassLoader clas
...
+        }
+        return AccessController.doPrivileged((PrivilegedAction<ClassLoader>) FactoryFinder.class::getClassLoader);
+    }
+
+    private static <T> T findFirstService(final String factoryId, final ClassLoader cl, final Class<T> service) {
+        final FindServiceAction<T> action = new FindServiceAction<>(factoryId, cl, service);
+        if (System.getSecurityManager() == null) {
+            return action.run();
+        }
+        return AccessController.doPrivileged(action);
```
andymc12, line 179: I understand putting the `getClassLoader` calls into a doPriv block, but I don't understand the `ServiceLoader` calls.  Does that need to be in a doPriv?  The javadoc doesn't mention that it throws a `SecurityException`.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/FactoryFinder.java
```diff
@@ -168,4 +163,44 @@ private static Object newInstance(final String className, final ClassLoader clas
...
+        }
+        return AccessController.doPrivileged((PrivilegedAction<ClassLoader>) FactoryFinder.class::getClassLoader);
+    }
+
+    private static <T> T findFirstService(final String factoryId, final ClassLoader cl, final Class<T> service) {
+        final FindServiceAction<T> action = new FactoryFinder.FindServiceAction<>(factoryId, cl, service);
+        if (System.getSecurityManager() == null) {
+            return action.run();
+        }
+        return AccessController.doPrivileged(action);
```
andymc12, line 179: Same comment as above. Can we avoid the doPriv on the `ServiceLoader.load(...)` calls?

andymc12 (commented):
Overall, I like this changes - the performance changes and adding `getClassLoader` calls into doPriv blocks.  I only have a question about whether `ServiceLoader.load(...)` and subsequent iterating requires a doPriv block.  Thanks!

jaxrs-api/src/main/java/jakarta/ws/rs/client/FactoryFinder.java
jamezp: The `ServiceLoader.load()` is actually safe, but it captures the `AccessContrelContext` is captured and during iteration a security exception could be thrown on `Constructor.newInstance()`.

From the JavaDocs:

> Service loaders always execute in the security context of the caller. Trusted system code should typically invoke the methods in this class, and the methods of the iterators which they return, from within a privileged security context.

spericas:
> > LGTM. Only comment is that `findFirstService`'s could create an `Action` to avoid duplicating code blocks.
> 
> Done.

Note that `PrivilegedAction` is a functional interface, so I think you could simplify this using capturing lambdas for even better readability without the need inner classes.

jamezp:
> Note that `PrivilegedAction` is a functional interface, so I think you could simplify this using capturing lambdas for even better readability without the need inner classes.

Good point. Updated!

spericas (approved):

spericas:
Merging as it has been approved 25 days ago.

PR #1006, opened 2021-08-04
Update UriBuilderIT test

andymc12:
This PR makes two TCK test changes:

(1) Changing test to verify scheme-only URI throws UriBuilderException - it is possible to create an "empty" URI (i.e. `URI.create("")` and RESTEasy allows `UriBuilder.newInstance().build()` to execute without throwing a `UriBuilderException`, but a scheme-only URI is not allowed (i.e. `UriBuilder.newInstance().scheme("http").build()` / `URI.create("http:")`).
 (2) Adding new test for checking `IllegalArgumentException` is thrown for unresolved template variables.

### DIFF
--- a/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/uribuilder/UriBuilderIT.java
+++ b/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/uribuilder/UriBuilderIT.java
@@ -19,6 +19,7 @@ package jakarta.ws.rs.tck.uribuilder;
 import static java.util.concurrent.TimeUnit.HOURS;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import java.net.URI;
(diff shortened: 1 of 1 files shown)


jaxrs-tck/src/main/java/jakarta/ws/rs/tck/uribuilder/UriBuilderIT.java
```diff
@@ -61,18 +61,38 @@ public final void shouldBuildValidInstanceFromScratch()
...
      * 
      * @throws ExecutionException   if the instance didn't boot correctly
      * @throws InterruptedException if the test took much longer than usually
      *                              expected
      */
     @Test
-    public final void shouldNotBuildInvalidUriFromScratch() throws InterruptedException, ExecutionException {
+    public final void shouldThrowUriBuilderExceptionOnSchemeOnlyUri() throws InterruptedException, ExecutionException {
         // given
-        final UriBuilder uriBuilder = UriBuilder.newInstance();
```
mkarg, line 73: Why do you want to allow `UriBuilder.newInstance();` to be called? I'd rather vote for having an *additional* test that disallows scheme-only uri builders instead of allowing completely invalid builders.

mkarg (changes requested):
Unless the Spec / API unambiguously mandates that `UriBuilder.newInstance();` is explicitly allowed, the TCK should not allow any implementation to accept this.

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/uribuilder/UriBuilderIT.java
andymc12: Sorry, I don't understand your comment.  `UriBuilder.newInstance()` was there from before - your change, I believe? The reason I am removing that is that `UriBuilder.newInstance().build()` is valid as it is equivalent to `URI.create("")` which is allowed.  Current implementations are failing on the current line 76, and as this test is not part of a new feature that is requiring new restrictions on the API, we should not be adding it. However, since a scheme-only URI is not allowed by the `javax.net` APIs and by existing JAX-RS implementation, it is a valid test.

mkarg: `newInstance()` was added in https://github.com/eclipse-ee4j/jaxrs-api/pull/641, and unfortunately we did not really discuss at that point, what the actual outcome of `UriBuilder.newInstance()` shall actuall be like - is it really the same as `URI.create("")`, or is it forbidden? So when I wrote the UriBuilderIT, I added that test that explicitly forbids it -- and that test was accepted by the committers, hence it is authoritative until decided by the committers.

In fact I do not have any problem if we change the test to actually proof the exact oppostite, but then please add your test as an *additional* test and *change* my existing test to proof that the outcome really is `URI.create("")` -- but do not leave the case `newInstance();` completely **untested** (hence: undecided by the TCK as an authoritative means).

Having said that, I will remove my veto once either you keep my test (and I open a PR turning it into the opposite) *or* you simply turn my test in the opposite in *this* PR. :-)

andymc12: Changes are in.

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/uribuilder/UriBuilderIT.java
```diff
@@ -61,18 +62,55 @@ public final void shouldBuildValidInstanceFromScratch()
...
      * @throws ExecutionException   if the instance didn't boot correctly
      * @throws InterruptedException if the test took much longer than usually
      *                              expected
      */
     @Test
-    public final void shouldNotBuildInvalidUriFromScratch() throws InterruptedException, ExecutionException {
+    public final void emptyUriBuilderBuildsEmptyUri() throws InterruptedException, ExecutionException {
         // given
-        final UriBuilder uriBuilder = UriBuilder.newInstance();
+        final URI uri = UriBuilder.newInstance().build();
```
mkarg, line 76: Can we please keep the BDD style in this test? "// given" should create the UriBuilder, "// when" should just do "build()". That would be nice. :-)

mkarg (approved):

spericas (approved):
LGTM

Issue #736, opened 2019-02-09
Clarification about Response.readEntity(XXX)


NicoNes:
According to the javadoc of:

- Response.readEntity(Class)
- Response.readEntity(javax.ws.rs.core.GenericType)
- Response.readEntity(Class, Annotation [])
- Response.readEntity(javax.ws.rs.core.GenericType, Annotation [])

1. those methods are supposed to close the original entity input stream (unless the supplied type is input stream) and then cache the result for subsequent retrievals via `Response.getEntity()`.
So it clearly means that those methods **MUST not** close() the response itself, else subsequent retrievals via `Response.getEntity()` will always end up with `IllegalStateException` being thrown.
Instead, only the underlying input stream should be closed.

2. Subsequent call to one of those methods MUST throw an IllegalStateException if the original entity input stream has already been fully consumed without buffering the entity data prior consuming.

In other words:

```
Message s1 = r.readEntity(Message.class);
System.out.println(s1.toString());
System.out.println(r.getEntity());      => should not throw no exception
s1 = r.readEntity(Message.class);                     => should throw an IllegalStateException
```
Is my understanding right for those 2 points ? because it seems that many implementation (https://issues.jboss.org/browse/RESTEASY-2112) including the RI (Jersey) does not behave as exepected for point 1.
If my understanding is wrong what is the intented behavior of those methods ?

chkal:
IMO your understanding is correct and both Jersey and RESTEasy doesn't behave correctly. But it would be interesting to hear other thoughts about this as well.

spericas:
These are good point @NicoNes. We definitely need a clarification here and an understanding as to why implementations do not appear to honor the `getEntity()` semantics here. Any comments @jansupol?

jansupol:
Javadoc to `getEntity` says:
> If the entity is represented by an un-consumed input stream the method will return the input stream.

That's what Jersey does. It does not return the cached entity, as the entity is never cached. 

I am concerned about automatic caching being a performance issue unless the cached entity means a simple reference to a Java entity returned by `readEntity`. (In that case, if the entity is mutable, `readEntity().toString() might differ from getEntity().toString()`). I am not sure what the use case of caching the entity would be. Should the user want to read the entity multiple times, `bufferEntity()` is available to be used. The user is aware of the buffering and the impact on performance and has a choice not to buffer.

The difference between caching and buffering is not clear, which is why the implementations do not honor the javadoc to the letter, IMO.

[...]

Issue #706, opened 2018-11-28
Add new method javax.ws.rs.core.Response.isClosed()


ronsigal:
A number of javax.ws.rs.core.Response methods, e.g., getEntity() and hasEntity(), include a line in the javadoc like

    * @throws IllegalStateException in case the response has been {@link #close() closed}.

However, in the absence of a method like isClosed(), we end up writing code like

>     try {
>        if (response.getEntity() != null) return response;
>     }
>     catch(IllegalStateException ise) {
>        // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
>     }

instead of

>     if (!response.isClosed() && response.getEntity() != null) {
>        return response;
>     }

The implementation of isClosed() should be simple, and it leads to nicer code.

An example of a potential use of Response.isClosed() arises in RESTEasy when:

1. A resource method uses a Client to contact another server and gets an exception back.

  2. A builtin ExceptionMapper tries to unwrap the Exception:

>        protected Response unwrapException(HttpRequest request, Throwable e, RESTEasyTracingLogger logger)
>        {
>           Response jaxrsResponse = null;
>           Throwable unwrappedException = e.getCause();
>
>           /*
>            *                If the response property of the exception does not
>            *                contain an entity and an exception mapping provider
>            *                (see section 4.4) is available for
>            *                WebApplicationException an implementation MUST use the
>            *                provider to create a new Response instance, otherwise
>            *                the response property is used directly.
>            */
>
>           if (unwrappedException instanceof WebApplicationException) {
>              WebApplicationException wae = (WebApplicationException) unwrappedException;
>              Response response = wae.getResponse();
>              if (response != null) {
>                 try { // Here's where we guard against IllegalStateExceptioin
>                    if (response.getEntity() != null) return response;
>                 }
>                 catch(IllegalStateException ise) {
>                    // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
>                 }
>              }
>           }
            ...

Before we wrapped the call to Response.getEntity() in try/catch, we were getting an IllegalStateException, which isn't very helpful.

spericas:
As I stated before, this addition seems reasonable to me. You may want to consider creating a PR to let people comment on it.

arjantijms:
+1 seems quite reasonable indeed.

mkarg:
@ronsigal Will you provide a PR?

mkarg:
@ronsigal Ron? Hello?

[...]

PR #1008, opened 2021-08-10
Add Response.isClosed()

NicoNes:
This PR is the follow-up to the [issue](https://github.com/eclipse-ee4j/jaxrs-api/issues/706) and  [discussion](https://github.com/eclipse-ee4j/jaxrs-api/issues/736) we had with @ronsigal.

To sum up it was about adding a method in the `Response` class to know when a response instance is closed. So that we no longer need to write code like:
```

try {
   if (response.getEntity() != null) return response;
}
catch(IllegalStateException ise) {
   // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
} 
```

but instead: 

```
if (!response.isClosed() && response.getEntity() != null) {
   return response;
}
```
I agree that the default implementation I proposed is a bit weird, not intuitive etc . I did it so that vendors don't need to make changes on their side.
I'm OK to remove it if you think it's not necessary.

Thanks

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
@@ -464,6 +464,22 @@ public abstract class Response implements AutoCloseable {
     public abstract String getHeaderString(String name);
 
     /**
+     * Check if the response is closed. The method returns {@code true} if the response is closed,
+     * returns {@code false} otherwise.
+     *
+     * @return {@code true} if the response has been {@link #close() closed}, {@code false} otherwise.
+     * @since 3.1
+     */
+    public boolean isClosed() {
+        try {
+            hasEntity();
+            return false;
+        } catch (IllegalStateException ignored) {
+            return true;
+        }
+    }
+
+    /**
      * Create a new ResponseBuilder by performing a shallow copy of an existing Response.
      * <p>
      * The returned builder has its own {@link #getHeaders() response headers} but the header values are shared with the


andymc12 (approved):
LGTM - thanks!

chkal (approved):

mkarg (approved):
LGTM. Thank you for this contribution!

mkarg:
I'd like to propose that we add this to 3.1 already as it does not change the API in an incompatible way thanks to the default implementation.

Speaking of the default implementation I'd like to kindly ask all vendors to override it using an implementation that works with higher performance than the default solution.

Last but not least I am kindly asking for TCK coverage.

NicoNes:
@mkarg do you want me to add somehing as follow in the doc ?

> For backwards compatibility, there is a default implementation that just relies on IllegalStateException thrown by `hasEntity` when response instance is closed.
> Implementations should consider replacing this with a more performant implementation

spericas (approved):

mkarg:
> @mkarg do you want me to add somehing as follow in the doc ?
> > For backwards compatibility, there is a default implementation that just relies on IllegalStateException thrown by `hasEntity` when response instance is closed.
> > Implementations should consider replacing this with a more performant implementation

I do not see a need for this, actually.

mkarg:
@NicoNes If you like to contribute it: For this feature we need a note in the "Changes since" section of the spec, an example in the examples folder would be great, and the TCK needs to test that the actual implementation provided by the vendor works just as correct as your default implementation.

spericas:
> @NicoNes If you like to contribute it: For this feature we need a note in the "Changes since" section of the spec

 That's required only for changes to the spec document.

> an example in the examples folder would be great, and the TCK needs to test that the actual implementation provided by the vendor works just as correct as your default implementation.

 Given the simplicity of this (just a convenience method), more of a nice to have IMO.

NicoNes:
Well @spericas and @mkarg given your two different opinions I don't really know what I have to do. Please let me know.

Thanks

mkarg:
> > @NicoNes If you like to contribute it: For this feature we need a note in the "Changes since" section of the spec
> 
> That's required only for changes to the spec document.
> 
> > an example in the examples folder would be great, and the TCK needs to test that the actual implementation provided by the vendor works just as correct as your default implementation.
> 
> Given the simplicity of this (just a convenience method), more of a nice to have IMO.

Okay, let's rephrase: It would be nice to have it, so users know about its existence and usage. :-)

spericas:
@NicoNes The PR has all the necessary approvals to be merged.

PR #1017, opened 2021-09-15
Use GenericEntity for List<EntityPart> entity in EntityPart javadoc.

jansupol:


jansupol:
The `MessageBodyWriter` would not know `#isWritable(Class<?> type, Type generic,...)` the generic type without this and would not know a list of which entities it is about to process.

spericas (approved):
Right, that makes sense.

andymc12 (approved):
LGTM - thanks @jansupol!

spericas:
Merging via fast track as this is a one liner.

jimma:
> The `MessageBodyWriter` would not know `#isWritable(Class<?> type, Type generic,...)` the generic type without this and would not know a list of which entities it is about to process.

@spericas @andymc12 @jansupol Late to look at this PR.  To me, this kind of change should be in api signature instead of only in javadoc.

Issue #979, opened 2021-06-19
Fix JavaDoc Warnings


mkarg:
There are several JavaDoc warnings in `examples` and `jaxrs-api` sources.

**This is a good newcomer's issue.**

andygibson:
Submitted PRs #984 , #985 and #986

chkal:
@andygibson Thanks a lot! Looks great!

mkarg:
There are new JavaDoc warnings. Maybe one of our new contributors like to fix it for 3.1? :-)

```
5 warnings
[WARNING] Javadoc Warnings
[WARNING] /workspace/jaxrs-api/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java:137: warning: @param "T" has already been specified
[WARNING] * @param <T> the entity type
[WARNING] ^
[WARNING] /workspace/jaxrs-api/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java:176: warning: @param "T" has already been specified
[WARNING] * @param <T> the entity type
[WARNING] ^
[WARNING] /workspace/jaxrs-api/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java:322: warning: @param "T" has already been specified
[WARNING] * @param <T>     the entity type
[WARNING] ^
[WARNING] /workspace/jaxrs-api/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java:349: warning: @return has already been specified
[WARNING] * @return the updated builder.
[WARNING] ^
[WARNING] /workspace/jaxrs-api/jaxrs-api/src/main/java/jakarta/ws/rs/core/EntityPart.java:374: warning: @param "T" has already been specified
[WARNING] * @param <T> the entity type
[WARNING] ^
```

mkarg:
@jelemux Like to fix this one? :-)

jelemux:
@mkarg  Yes, I do. I'll see if I can find some time for it over the weekend.

mkarg:
@jelemux Is preparing a PR for this. Thank you, Jeremias! :-)

jelemux:
Submitted PR #1028

---

Also, what is to be done about the issue I mentioned?

> What maybe should be addressed as well are the JavaDoc Errors in `jaxrs-tck`.  
There are lots of errors because the code just got ported and focus was on the migration.  
Maybe we don't need to generate JavaDoc for this at all though and it can be excluded? WDYT?  
[...]

PR #1028, opened 2021-10-03
Fix JavaDoc warnings

jelemux:
This fixes the JavaDoc warnings in `EntityPart` as described by @mkarg in #979.  
It also fixes a JavaDoc error in the `package-info` of the bootstrap example (misplaced semicolon).

What maybe should be addressed as well are the JavaDoc Errors in `jaxrs-tck`.  
There are lots of errors because the code just got ported and focus was on the migration.  
Maybe we don't need to generate JavaDoc for this at all though and it can be excluded? WDYT?  
Should this be discussed in a seperate PR / issue?

### DIFF
--- a/examples/src/main/java/jaxrs/examples/bootstrap/package-info.java
+++ b/examples/src/main/java/jaxrs/examples/bootstrap/package-info.java
@@ -17,6 +17,6 @@
  *
  * @author Markus KARG (markus@headcrashing.eu)
  * @since 3.1
- * @see jakarta.ws.rs.SeBootstrap;
+ * @see jakarta.ws.rs.SeBootstrap
  */
 package jaxrs.examples.bootstrap;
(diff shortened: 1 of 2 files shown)


mkarg (approved):
LGTM. Thanks a lot for this quick PR! :-)

mkarg:
@jelemux asked
> What maybe should be addressed as well are the JavaDoc Errors in `jaxrs-tck`.  
There are lots of errors because the code just got ported and focus was on the migration.  
Maybe we don't need to generate JavaDoc for this at all though and it can be excluded? WDYT?  
Should this be discussed in a seperate PR / issue?

For simplified and accelerated reviews it is always good to have multiple small PRs so we can quickly make progress instead of being hold back by lengthy discussions only needed for some detail of just one specific PR.

Having said that, feel free to open a separate PR for each independent idea you like to propose. :-)

jansupol (approved):

andymc12 (approved):
LGTM - thanks @jelemux!

PR #1034, opened 2021-10-15
Clarify exception processing by CompletionCallback and the default ExceptionMapper

jansupol:

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/container/CompletionCallback.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/container/CompletionCallback.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -21,7 +21,7 @@ package jakarta.ws.rs.container;
  * <p>
  * A completion callback is invoked when the whole request processing is over, i.e. once a response for the request has
  * been processed and sent back to the client or in when an unmapped exception or error is being propagated to the
- * container.
+ * default exception mapper.
  * </p>
  *
  * @author Marek Potociar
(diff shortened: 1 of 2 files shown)


jaxrs-api/src/main/java/jakarta/ws/rs/container/CompletionCallback.java
```diff
@@ -21,7 +21,7 @@
  * <p>
  * A completion callback is invoked when the whole request processing is over, i.e. once a response for the request has
  * been processed and sent back to the client or in when an unmapped exception or error is being propagated to the
```
spericas, line 23: Old text: "in when" -> "when"

jaxrs-api/src/main/java/jakarta/ws/rs/container/CompletionCallback.java
```diff
@@ -21,7 +21,7 @@
...
- * container.
+ * default exception mapper.
  * </p>
```
spericas, line 25: This can be made more accurate. It will be propagated to _the default exception mapper_ only if it hasn't been overridden. So something like "propagated to an exception mapper, possibly the default one if not overridden".

jaxrs-api/src/main/java/jakarta/ws/rs/container/CompletionCallback.java
```diff
@@ -30,19 +30,19 @@
 public interface CompletionCallback {
     /**
      * A completion callback notification method that will be invoked when the request processing is finished, after a
-     * response is processed and is sent back to the client or when an unmapped throwable has been propagated to the hosting
-     * I/O container.
+     * response is processed and is sent back to the client or when an unmapped throwable has been propagated to the default
+     * exception mapper.
      * <p>
```
spericas, line 35: Same comment as above about "default exception mapper". And some more occurrences below as well.

jaxrs-api/src/main/java/jakarta/ws/rs/container/CompletionCallback.java
jansupol: I am fine with updating the javadoc. However, I do not understand the comment. It cannot be propagated to any other exception mapper but the default one. Jakarta REST does not provide a mechanism to override the default exception mapper, per se, it merely allows for registering a custom exception mapper. But if the custom exception mapper applies, there is no unmapped exception at this point, the mapped throwable is not used for the `CompletionCallback` (in 3.0).

Perhaps we can replace "unmapped exception" by "exception unmapped by a registered exception mapper"?

PR #1043, opened 2021-10-28
Reverts changes introduced by PR 972

spericas:
Forcing the default content type of application/octet-stream is problematic in practice. It breaks the common use case of inheriting class-level `@Consumes` annotations on `@GET` methods. Text now also clarifies that specific matching case. See discussion in Issue #970. 

Requesting fast track.


andymc12 (approved):
LGTM - thanks @spericas

jansupol (approved):

PR #1046, opened 2021-11-01
Set explicit versions for compiler and surefire plugins

spericas:
These were dropped together with the profiles of the various JDK's and can result on a Maven warning (see pipeline) or worse failure to compile if picking up an old surefire version (was able to reproduce this locally).

Requesting fast track.


mkarg (approved):
Oops... sorry.

jansupol (approved):

PR #1047, opened 2021-11-02
Fixed pom parent-child relationship 

spericas:
Fixed pom parent-child relationship to make sure deploy works on Jenkins nightly builds. Nightly builds have been failing for a while due to misconfiguration problems. See https://ci.eclipse.org/jaxrs/job/Nightly%20Build/job/master/

Some minor re-formatting too. Fast track.

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -16,7 +16,6 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>jakarta.ws.rs-examples</artifactId>
-    <version>3.1-SNAPSHOT</version>
     <packaging>jar</packaging>
     <name>Jakarta RESTful WS Examples</name>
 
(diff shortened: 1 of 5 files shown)


andymc12 (approved):
Overall, I think this looks good.  One thought/question: would it be possible to make the version a property in the parent pom so that we could change it in one place instead of several?

spericas:
> Overall, I think this looks good. One thought/question: would it be possible to make the version a property in the parent pom so that we could change it in one place instead of several?

I removed the version from the child poms.

Incidentally, a lot of things can move to the parent pom. I had started doing this in my 4.0 branch. Will come back to that after the release.

andymc12 (approved):
LGTM. Thanks!

andymc12 (approved):

jansupol (approved):

Issue #993, opened 2021-07-07
Missing status codes in Response class


reschke:
See <https://github.com/eclipse-ee4j/jaxrs-api/pull/582>, which was closed for reasons that escape me.

There are likely more status codes missing, see the IANA registry at <https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml>. It would also be good to update comments to refer to current RFCs, not obsoleted ones.

cc: @mkarg

spericas:
The reason is there:

 "@mithunsasidharan Are you still interested in working on this? If so, could you rebase against `3.1-SNAPSHOT" and update the pull request accordingly?" 

Needs to be resubmitted for 3.1 using the correct branch.

mithunsasidharan:
@spericas : Is this something thats already been pickedup ? If not I'd like to work on it !

spericas:
@mithunsasidharan Feel free to work on it. See previous work and suggestion by @reschke.

spericas:
@mithunsasidharan Any update on this work?

jelemux:
Sorry for the late response, I've caught the flu and then had a week of vacation after I recovered.

I went through all the missing status codes and singled those out that MAY be of general interest for REST applications.  
I've documented my findings on [this wiki page](https://github.com/jelemux/jaxrs-api/wiki/Missing-status-codes). Feel free to go through them and comment your opinions here.

In short, those are the status codes that may be of _general_ interest to REST applications:
* [203 Non-Authoritative Information](https://tools.ietf.org/html/rfc7231#section-6.3.4) 
* [300 Multiple Choices](https://tools.ietf.org/html/rfc7231#section-6.4.1) 
* [308 Permanent Redirect](https://tools.ietf.org/html/rfc7538)
* [451 Unavailable For Legal Reasons](https://tools.ietf.org/html/rfc7725)

I think it is best to create separate PRs for each of these since there may be arguments against some of these that don't concern the others.  
The only exception being the argument that those are all only _Proposed Standard_ RFCs, which has already been [discussed](https://github.com/eclipse-ee4j/jaxrs-api/pull/582), but with no clear conclusion (to me). Does this argument still stand?

jelemux:
What may not be of _general_ interest to REST applications but still came up is WebDAV.  
The [JAXRS specification states under goals](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#goals):

[...]

PR #1053, opened 2021-11-08
Add status code 308 Permanent Redirect

jelemux:
This adds the status code _[308 Permanent Redirect](https://tools.ietf.org/html/rfc7538)_ as discussed in #993.

One possible application would be if part of an API has been permanently moved to another URI.


andymc12 (approved):
LGTM - thanks @jelemux!

mkarg (approved):

jansupol (approved):

PR #1054, opened 2021-11-08
Add status code 300 Multiple Choices

jelemux:
This adds the status code _[300 Multiple Choices](https://tools.ietf.org/html/rfc7231#section-6.4.1)_ as discussed in #993.

I could see reactive content negotiation being used in a REST service, be it about language, media type or something else. This is why I think we should include this.


andymc12 (approved):
LGTM - thanks @jelemux!

mkarg (approved):

mkarg:
@spericas Will this make it into 3.1 or shall we postpone it for 4.0?

spericas:
> @spericas Will this make it into 3.1 or shall we postpone it for 4.0?

I will merge the 3 PR's related to status codes, but nothing more after that that isn't related to the TCK's. Will assume fast track for these.

PR #1055, opened 2021-11-08
Add status code 451 Unavailable For Legal Reasons (#993)

jelemux:
This adds the status code _[451 Unavailable For Legal Reasons](https://tools.ietf.org/html/rfc7725)_ as discussed in #993.

I think the use-case for this is self-explanatory.


andymc12 (approved):
LGTM - thanks @jelemux!

mkarg (approved):

jansupol (approved):

PR #1056, opened 2021-11-10
Fix NPE for NewCookie#equals when simesite is null

jansupol:


jansupol:
The NPE is caused by using a constructor/builder with no SameSite. It is discovered by one of the TCK tests.

Requesting fast track.

spericas (approved):

andymc12 (approved):
LGTM - thanks!

PR #1062, opened 2021-11-17
separate modules into profiles

senivam:
The purpose of the PR is to separate modules in the parent pom into profiles, which is good for releases to the maven staging/central. 

The issue is that when releasing API (from the jaxrs-api module) the parent pom shall be published into staging/central as well. 

So, when the forward approach is applied (the project is being built from its root folder) all modules are being compiled. But when it comes to releasing where only particular modules are being released related parent pom becomes invalid from the point of modules view. 

In the PR there are 2 profiles with 3 and 1 modules respectively. The module with 3 profiles contains references to API, TCK, and examples. And the module with one module relates to the spec which is never being published to staging/central. 

Thus when the project is being compiled from its root folder using ``mvn clean install`` the project is being assembled with 3 modules> API, TCK, and examples. 

In order to assemble the project with the spec included, the build shall be called in the following way: ``mvn clean install -Djaxrs.all.build``. 

Separation of the spec module into an inactive profile guarantees that the parent pom is valid in the maven central without the spec module. 

Signed-off-by: Maxim Nesen <maxim.nesen@oracle.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,12 +47,30 @@
                 </repository>
             </distributionManagement>
         </profile>
+        <profile>
+            <id>dependentModules</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+                <property>
+                    <name>jaxrs.all.build</name>
+                </property>
+            </activation>
+            <modules>
+                <module>jaxrs-api</module>
+                <module>jaxrs-tck</module>
+                <module>examples</module>
+            </modules>
+        </profile>
+        <profile>
+            <id>dependentSpecification</id>
+            <activation>
+                <property>
+                    <name>jaxrs.all.build</name>
+                </property>
+            </activation>
+            <modules>
+                <module>jaxrs-spec</module>
+            </modules>
+        </profile>
     </profiles>
-
-    <modules>
-        <module>jaxrs-api</module>
-        <module>jaxrs-spec</module>
-        <module>jaxrs-tck</module>
-        <module>examples</module>
-    </modules>
 </project>


spericas (approved):
LGTM. But why use 'jaxrs.all.build' instead of an unqualified name like 'all' or something simpler. Is that convention used elsewhere?

senivam:
@spericas thank you for the thumb up :) I would like to name the property in a way it shows that it belongs to the concrete project (jaxrs), concrete purpose (build, and not release for example) and that it provides all modules (so all). And it will be better searchable just in case of grep or some regExp. That's only an explanation of my intention :)  A simpler name has its advantages as well. So, the name can be changed if my naming is considered too complex.

andymc12 (approved):
LGTM - wrt "jaxrs.all.build" vs "all", I'm fine either way.  Thanks!

mkarg:
I like to remember the developer rules we agreed upon when starting this project: If fast-path is wanted, this has to be claimed in the origin issue description. The reason was that we all, from time to time, will not be able to follow full discussion threads on all issues, and we have a pointer where to focus on. Having said that, I am currently in a situation where I need such pointer. I am not pleased with the recent policy that issues are *not* marked as fast-track and hastily merged after few days. **Please everybody, in particular our project lead, do always actively request fast-track and abstain from fast-track if not requested.** Thanks.

spericas:
I’m making the call to automatically fast track anything that is required to get the release out on time and is not related to API or documentation changes, but just the release process or TCK migration. Without this decision, the release will take months longer to complete.

 If you dislike this, please take the time to review the PR’s faster when coming close to a release date.

— Santiago

On Nov 20, 2021, at 3:10 AM, Markus KARG ***@***.******@***.***>> wrote:



I like to remember the developer rules we agreed upon when starting this project: If fast-path is wanted, this has to be claimed in the origin issue description. The reason was that we all, from time to time, will not be able to follow full discussion threads on all issues, and we have a pointer where to focus on. Having said that, I am currently in a situation where I need such pointer. I am not pleased with the recent policy that issues are not marked as fast-track and hastily merged after few days. Please everybody, in particular our project lead, do always actively request fast-track and abstain from fast-track if not requested. Thanks.

mkarg:
I understand your reasons but that is a violation of the rules we agreed upon. Please stick to the rules unless we agree on new rules. Thanks.

arjantijms:
Maybe we should reconsider and think about what makes Jakarta REST different from all the other projects in that it's the only project that has this fast-track thing.

mkarg:
> Maybe we should reconsider and think about what makes Jakarta REST different from all the other projects in that it's the only project that has this fast-track thing.

You are free to propose a change of our rules and start a voting among all committers, but please discuss it in the mailinglist, as it is not directly related to *this* PR. Thanks.

Issue #938, opened 2021-01-11
TCK Challenge: Using `@Context` parameters in resource constructors should be optional


andymc12:
**Specification:**
Jakarta RESTful Web Services 3.0 (EE 9.0) Full Platform TCK

**Challenged test(s):**
[com.sun.ts.tests.jaxrs.spec.resourceconstructor#visibleTest](https://github.com/eclipse-ee4j/jakartaee-tck/blob/adf6ca2beb09b72a77e685d5ec1ad425e0782bc3/src/com/sun/ts/tests/jaxrs/spec/resourceconstructor/JAXRSClient.java#L48)

**TCK version:**
9.0.0

**Tested implementation:**
Open Liberty using RESTEasy + Weld tested by IBM

**Description:**

The specification allows CDI to manage the dependency lifecycle of resources and provider if present. This test exists to verify that the proper constructor is invoked.  Per the spec, the constructor with the most arguments annotated with `@Context` should be invoked.  However, when CDI is in play, then the rules of [section 11.2.9](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#additional_reqs) take effect, specifically:
> Support for constructor injection of JAX-RS resources is OPTIONAL. Portable applications MUST instead use fields or bean properties in conjunction with a @PostConstruct annotated method. Implementations SHOULD warn users about use of non-portable constructor injection.

In the (still under development) EE 9 version of Open Liberty that uses RESTEasy and Weld, the CDI implementation creates the instance of the resource class using normal CDI rules (invoking the no-arg constructor) which fails the test.  I think that in the long run, the test should be modified to check whether CDI is enabled and then determine the appropriate constructor to be invoked - though these rules may change significantly if future versions of RESTful WS will require CDI.

I'm pulling this "form" from other TCK challenges in other EE9 projects.  If there is a better way, please let me know.  We may also want to create a "challenge" label, but I'm using "tck" for now.

jansupol:
I agree that Section 11.2.9 makes CDI behave in contradiction to Section 3.1.2 which says:
> If more than one public constructor is suitable then an implementation MUST use the one with the most
parameters.

I agree that CDI will be the first-class-citizen in the next Jakarta REST release, Section 11.2.9 should take precedence over 3.1.2, and Section 3.1.2 should be reformulated, perhaps the requirement can be dropped.

[...]

PR #1063, opened 2021-11-23
Ensure resourceconstructor test includes beans.xml

andymc12:
This change adds the beans.xml file to the deployed WAR for the `resourceconstructor` test.  The beans.xml was added as part of challenge issue #938, but from my testing it was not added to the WAR file.  This change fixes that.

Since this only fixes a bug in a TCK test and makes no changes to spec or API, I'd like to request fast-track.

### DIFF
--- a/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/common/JAXRSCommonClient.java
+++ b/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/common/JAXRSCommonClient.java
@@ -533,15 +533,19 @@ public abstract class JAXRSCommonClient {
     return sb.toString();
   }
 
-  public static String editWebXmlString(InputStream inStream) throws IOException{
+  public static String toString(InputStream inStream) throws IOException{
     String line;
-    String webXmlTemplate = "";
+    String str = "";
     try (BufferedReader bufReader = new BufferedReader(new InputStreamReader(inStream, StandardCharsets.UTF_8))) {
       while ((line = bufReader.readLine()) != null) {
-        webXmlTemplate = webXmlTemplate + line + System.lineSeparator();
+        str = str + line + System.lineSeparator();
       }
+      return str;
     }
-    
+  }
+  
+  public static String editWebXmlString(InputStream inStream) throws IOException{
+    String webXmlTemplate = toString(inStream);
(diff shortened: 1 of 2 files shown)


mkarg (commented):
I vote +0 here as I understand your fix is beneficial but I did not yet dive into the TCK source deeply, so I cannot judge the complete correctness.

jansupol (approved):

jaxrs-tck/src/main/java/jakarta/ws/rs/tck/common/JAXRSCommonClient.java
```diff
@@ -533,15 +533,19 @@ protected static String buildHeaderMediaType(String header, MediaType type) {
...
 
-  public static String editWebXmlString(InputStream inStream) throws IOException{
+  public static String toString(InputStream inStream) throws IOException{
     String line;
-    String webXmlTemplate = "";
+    String str = "";
     try (BufferedReader bufReader = new BufferedReader(new InputStreamReader(inStream, StandardCharsets.UTF_8))) {
       while ((line = bufReader.readLine()) != null) {
-        webXmlTemplate = webXmlTemplate + line + System.lineSeparator();
+        str = str + line + System.lineSeparator();
```
spericas, line 541: I know this is just a TCK, but should we replace this by a `StringBuilder`?

andymc12: Sure - I think I just left the existing logic alone there, but we could optimize, modernize, and economize it to be:
```
  public static String toString(InputStream inStream) throws IOException{
    try (BufferedReader bufReader = new BufferedReader(new InputStreamReader(inStream, StandardCharsets.UTF_8))) {
      return bufReader.lines().collect(Collectors.joining(System.lineSeparator()));
    }
  }

  public static String editWebXmlString(InputStream inStream) throws IOException{
    return toString(inStream).replaceAll("servlet_adaptor", servletAdaptor);
  }
```

spericas (approved):

andymc12:
@jansupol could you take another look when you have a minute or two?  The only change since you last approved was to modify the string concatenation logic to use `Collectors.joining(...)` instead.  Thanks!

jansupol (approved):

PR #1064, opened 2021-11-26
Allow vendors to add their deployment descriptors to arquillian archives

jansupol:

### DIFF
--- a/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/common/JAXRSCommonClient.java
+++ b/jaxrs-tck/src/main/java/jakarta/ws/rs/tck/common/JAXRSCommonClient.java
@@ -133,6 +133,16 @@ public abstract class JAXRSCommonClient {
   protected static final String GF_SUFFIX = ".gf";
 
   /**
+   * Folder name of vendor descriptor property files
+   */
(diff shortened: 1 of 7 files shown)


jansupol:
@scottmarlow @andymc12 This is how the vendor-specific descriptors could be brought into the Arquillian archive. WDYT?

arjantijms:
@jansupol Arquillian has a standard SPI for what I think you're doing here, it's `org.jboss.arquillian.container.test.spi.client.deployment.ApplicationArchiveProcessor`

The Arquillian based JWT TCK describes this as well. See: https://github.com/eclipse/microprofile-jwt-auth/blob/master/tck/README.adoc#creating-your-implementation-tck-harness-artifact

jansupol:
Ok, an alternative approach using ApplicationArchiveProcessor is in #1065.

@arjantijms Thanks for pointing it out.

scottmarlow:
For the many Platform TCK tests, it would be nice to support something like this pr that allows vendors to specify their folder that contains all of their vendor descriptor files for various test packages.  For each test package, we would need to indicate the target archive in some way (whether its a jar, war, ear, rar or whatever) for each contained deployment descriptor.  

The ApplicationArchiveProcessor way seems like a group approach as well.

jansupol:
Closing this in favor of #1065.

PR #1066, opened 2021-12-05
Convenience method for checking header value lists

mkarg:
For release 4.0 I'd like to propose a convencience method for response/request contexts which I missed a lot in the past years and which I implemented again and again in multiple ways -- most of them lacking either performance or perfection.

The problem it solves is that it is rather hard to get it perfectly correct checking for items in comma-separated headers - without losing valuable performance.

Example: How to check if there is a `Cache-Control: no-store` value set in case the actual sent headers contains `Cache-Control` multiple times and each of the findings is possibly a comma-separated list? RegEx over getHeaderString() comes into mind, but this is rather slow, as it first concatenates everything just to split it up afterwards. Iterating all findings comes into mind next, which is everything but convenient nor particularly readable.

Solution: An implementation of this new method would be provided by the runtime vendor, hence could provide not only most convenient usage to the caller, but also perform best possible as it is optimized according to the actual implementation's internal map design

WDYT?

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
@@ -238,7 +238,7 @@ public interface ClientRequestContext {
      * @see #getHeaderString(String)
      * @since 4.0
      */
-    default public boolean containsHeaderString(String name, Predicate<String> valuePredicate) {
+    public default boolean containsHeaderString(String name, Predicate<String> valuePredicate) {
         return containsHeaderString(name, ",", valuePredicate);
     }
 
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
@@ -193,6 +193,23 @@ public interface ClientRequestContext {
     public String getHeaderString(String name);
 
     /**
+     * Checks whether a header with a specific name and value (or item of the comma-separated value list) exists.
+     *
+     * Each single header value is converted to String using a {@link jakarta.ws.rs.ext.RuntimeDelegate.HeaderDelegate} if one
+     * is available via {@link jakarta.ws.rs.ext.RuntimeDelegate#createHeaderDelegate(java.lang.Class)} for the header value
+     * class or using its {@code toString} method if a header delegate is not available.
+     *
+     * @param name the message header.
+     * @param value the message header value.
+     * @param ignoreCase whether to ignore upper/lower case.
+     * @return {@code true} if and only if a header with the provided name exists having either the exact value or whose
+     * comma-separated header string contains value as a whole word.
+     * @see #getHeaders()
+     * @see #getHeaderString(String)
+     */
+    public boolean containsHeaderString(String name, String value, boolean ignoreCase);
+
+    /**
      * Get message date.
      *
      * @return the message date, otherwise {@code null} if not present.
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
@@ -194,7 +194,7 @@ public interface ClientRequestContext {
     public String getHeaderString(String name);
 
     /**
-     * Checks whether a header with a specific name and value (or item of the comma-separated value list) exists.
+     * Checks whether a header with a specific name and value (or item of the token-separated value list) exists.
      *
      * Each single non-string header value is converted to String using a {@link jakarta.ws.rs.ext.RuntimeDelegate.HeaderDelegate} if one
      * is available via {@link jakarta.ws.rs.ext.RuntimeDelegate#createHeaderDelegate(java.lang.Class)} for the header value
@@ -210,12 +210,37 @@ public interface ClientRequestContext {
      * @param valueSeparatorRegex Separates the header value into single values. {@code null} does not split.
      * @param valuePredicate value must fulfil this predicate.
      * @return {@code true} if and only if a header with the given name exists, having either a whitespace-trimmed value
-     * matching the predicate, or having at least one whitespace-trimmed single value in a comma-separated list of single values.
+     * matching the predicate, or having at least one whitespace-trimmed single value in a token-separated list of single values.
      * @see #getHeaders()
      * @see #getHeaderString(String)
      * @since 4.0
      */
     public boolean containsHeaderString(String name, String valueSeparatorRegex, Predicate<String> valuePredicate);
+
+    /**
+     * Checks whether a header with a specific name and value (or item of the comma-separated value list) exists.
+     *
+     * Each single non-string header value is converted to String using a {@link jakarta.ws.rs.ext.RuntimeDelegate.HeaderDelegate} if one
+     * is available via {@link jakarta.ws.rs.ext.RuntimeDelegate#createHeaderDelegate(java.lang.Class)} for the header value
+     * class or using its {@code toString} method if a header delegate is not available.
+     *
+     * <p>
+     * For example: {@code containsHeaderString("cache-control", "no-store"::equalsIgnoreCase)} will return {@code true} if
+     * a {@code Cache-Control} header exists that has the value {@code no-store}, the value {@code No-Store} or the value
+     * {@code Max-Age, NO-STORE, no-transform}, but {@code false} when it has the value {@code no-store;no-transform}
+     * (missing comma), or the value {@code no - store} (whitespace within value).
+     *
+     * @param name the message header.
+     * @param valuePredicate value must fulfil this predicate.
+     * @return {@code true} if and only if a header with the given name exists, having either a whitespace-trimmed value
+     * matching the predicate, or having at least one whitespace-trimmed single value in a comma-separated list of single values.
+     * @see #getHeaders()
+     * @see #getHeaderString(String)
+     * @since 4.0
+     */
+    default public boolean containsHeaderString(String name, Predicate<String> valuePredicate) {
+        return containsHeaderString(name, ",", valuePredicate);
+    }
 
     /**
      * Get message date.
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
@@ -202,8 +202,8 @@ public interface ClientRequestContext {
      *
      * @param name the message header.
      * @param valuePredicate value must fulfil this predicate.
-     * @return {@code true} if and only if a header with the provided name exists having either the exact value or whose
-     * comma-separated header string contains value as a whole word.
+     * @return {@code true} if and only if a header with the given name exists, having either a whitespace-trimmed value
+     * matching the predicate, or having at least one whitespace-trimmed single value in a comma-separated list of single values.
      * @see #getHeaders()
      * @see #getHeaderString(String)
      * @since 4.0
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
@@ -200,6 +200,12 @@ public interface ClientRequestContext {
      * is available via {@link jakarta.ws.rs.ext.RuntimeDelegate#createHeaderDelegate(java.lang.Class)} for the header value
      * class or using its {@code toString} method if a header delegate is not available.
      *
+     * <p>
+     * For example: {@code containsHeaderString("cache-control", "no-store"::equalsIgnoreCase)} will return {@code true} if
+     * a {@code Cache-Control} header exists that has the value {@code no-store}, the value {@code No-Store} or the value
+     * {@code Max-Age, NO-STORE, no-transform}, but {@code false} when it has the value {@code no-store;no-transform}
+     * (missing comma), or the value {@code no - store} (whitespace within value).
+     *
      * @param name the message header.
      * @param valuePredicate value must fulfil this predicate.
      * @return {@code true} if and only if a header with the given name exists, having either a whitespace-trimmed value
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
@@ -201,13 +201,13 @@ public interface ClientRequestContext {
      * class or using its {@code toString} method if a header delegate is not available.
      *
      * <p>
-     * For example: {@code containsHeaderString("cache-control", null, "no-store"::equalsIgnoreCase)} will return {@code true} if
+     * For example: {@code containsHeaderString("cache-control", ",", "no-store"::equalsIgnoreCase)} will return {@code true} if
      * a {@code Cache-Control} header exists that has the value {@code no-store}, the value {@code No-Store} or the value
      * {@code Max-Age, NO-STORE, no-transform}, but {@code false} when it has the value {@code no-store;no-transform}
      * (missing comma), or the value {@code no - store} (whitespace within value).
      *
      * @param name the message header.
-     * @param valueSeparatorRegex Separates the header value into single values. Defaults to {@code \s*,\s*} when {@code null}.
+     * @param valueSeparatorRegex Separates the header value into single values. {@code null} does not split.
      * @param valuePredicate value must fulfil this predicate.
      * @return {@code true} if and only if a header with the given name exists, having either a whitespace-trimmed value
      * matching the predicate, or having at least one whitespace-trimmed single value in a comma-separated list of single values.
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
@@ -201,12 +201,13 @@ public interface ClientRequestContext {
      * class or using its {@code toString} method if a header delegate is not available.
      *
      * <p>
-     * For example: {@code containsHeaderString("cache-control", "no-store"::equalsIgnoreCase)} will return {@code true} if
+     * For example: {@code containsHeaderString("cache-control", null, "no-store"::equalsIgnoreCase)} will return {@code true} if
      * a {@code Cache-Control} header exists that has the value {@code no-store}, the value {@code No-Store} or the value
      * {@code Max-Age, NO-STORE, no-transform}, but {@code false} when it has the value {@code no-store;no-transform}
      * (missing comma), or the value {@code no - store} (whitespace within value).
      *
      * @param name the message header.
+     * @param valueSeparatorRegex Separates the header value into single values. Defaults to {@code \s*,\s*} when {@code null}.
      * @param valuePredicate value must fulfil this predicate.
      * @return {@code true} if and only if a header with the given name exists, having either a whitespace-trimmed value
      * matching the predicate, or having at least one whitespace-trimmed single value in a comma-separated list of single values.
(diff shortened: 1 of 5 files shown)


jansupol:
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.3.0:jar (attach-javadocs) on project jakarta.ws.rs-api: MavenReportException: Error while generating Javadoc: 
Error:  Exit code: 1 - /home/runner/work/jaxrs-api/jaxrs-api/jaxrs-api/src/main/java/jakarta/ws/rs/container/ContainerResponseContext.java:129: error: reference not found
Error:       * @see #getHeaderString()

mkarg:
> Error: ... @see #getHeaderString()

Jan, what do you want to tell me? This is just a draft to discuss the *idea*. This is not a mergable PR.

mkarg:
> Error: Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.3.0:jar (attach-javadocs) on project jakarta.ws.rs-api: MavenReportException: Error while generating Javadoc: Error: Exit code: 1 - /home/runner/work/jaxrs-api/jaxrs-api/jaxrs-api/src/main/java/jakarta/ws/rs/container/ContainerResponseContext.java:129: error: reference not found Error: * @see #getHeaderString()

Thanks. Fixed.

mkarg:
If nobody objects to adopting this small feature to JAX-RS 4.0, then I would invest some more time and add a TCK test in this PR and remove the draft state flag. Any objects? Please veto NOW so I spare the time for writing the tests. Thanks.

chkal:
A few thoughts about this proposal.

First of all: Personally, I never had a situation where this method would have been helpful. However, I see use cases for it, so I'm not against adding it to the API.

One thing I'm wondering about is how to deal with case-insensitivity. For the header name, the situation is clear, as HTTP header names are case-insensitive, which should be handled correctly by the implementation. But what about the header value? My first feeling was that it should do a case-sensitive match. However, in case of `Cache-Control` mentioned above, [directives are case-insensitive](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2). So I'm wondering how we should deal with such cases.

arjantijms:
Could Servlet not need the exact same utility?

mkarg:
> One thing I'm wondering about is how to deal with case-insensitivity. For the header name, the situation is clear, as HTTP header names are case-insensitive, which should be handled correctly by the implementation. But what about the header value? My first feeling was that it should do a case-sensitive match. However, in case of `Cache-Control` mentioned above, [directives are case-insensitive](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2). So I'm wondering how we should deal with such cases.

We could provide two methods, one case-sensitive and one case-insensitive, just like the JRE does it https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/String.html#equalsIgnoreCase(java.lang.String).

mkarg:
> Could Servlet not need the exact same utility?

OT: Frankly spoken, I personally do not see any use of Servlets anymore at all these days. JAX-RS is superior from the user's view in most cases other than performance wrt to HTTP use. I have not used neither Servlets, nor Servlet-based JAX-RS since years (this is why I lead the Java SE Bootstrap project). If you ask me, the core idea of Servlet API is deprecated, as it was invented as a pure Java replacement for NSAPI / ISAPI, hence to enable Java code run on popular *web* servers like IIS or Apache (which simply tried to prevent starting native processes using (Fast) CGI). Those times are over since decades. Since most people meanwhile migrated to "real" Java based application servers (hence: products able to natively run JAX-RS), and since mainstream users are migrating to cloud / servlerss architectures, there is not much use left over for Servlets. If you ask me, Servlet API should be dropped in favor of a "Streamlet API" which just provides the interface between a JAX-RS implementation and an NIO engine like Netty or Grizzly. The HTTP part should be completely hidden in the JAX-RS engine; if Grizzly / Netty want to process HTTP, they could natively support JAX-Rs instead of Servlet API instead. Having said that: If somebody likes, he could copy my work provided here for JAX-RS, but I personally will not spend a minute into Servlet API anymore.

mkarg:
> > One thing I'm wondering about is how to deal with case-insensitivity. For the header name, the situation is clear, as HTTP header names are case-insensitive, which should be handled correctly by the implementation. But what about the header value? My first feeling was that it should do a case-sensitive match. However, in case of `Cache-Control` mentioned above, [directives are case-insensitive](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2). So I'm wondering how we should deal with such cases.
> 
> We could provide two methods, one case-sensitive and one case-insensitive, just like the JRE does it https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/String.html#equalsIgnoreCase(java.lang.String).

Thinking deeper about it, I do not see the need for different methods just to prevent a single "if". I modified the proposal accordingly. WDYT?

mkarg:
> > > One thing I'm wondering about is how to deal with case-insensitivity. For the header name, the situation is clear, as HTTP header names are case-insensitive, which should be handled correctly by the implementation. But what about the header value? My first feeling was that it should do a case-sensitive match. However, in case of `Cache-Control` mentioned above, [directives are case-insensitive](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2). So I'm wondering how we should deal with such cases.
> > 
> > We could provide two methods, one case-sensitive and one case-insensitive, just like the JRE does it https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/String.html#equalsIgnoreCase(java.lang.String).
> 
> Thinking deeper about it, I do not see the need for different methods just to prevent a single "if". I modified the proposal accordingly. WDYT?

Or should we go with a *super-flexible* solution instead, like providing a *predicate*? One could then pass even precompiled regex, a lambda expression, or method handles like `String::equalsIgnoreCase`. 😃 To me, this sounds more powerful. @chkal  WDYT?
```java
boolean containsHeaderString(Predicate<String> predicate)
```

jansupol:
Now, this can possibly have a `default` implementation, nicht wahr?

jamezp:
I would think the default should be case-insensitive since in HTTP/1 header names should be case-insensitive. AIUI in HTTP/2 they must be lowercase.

mkarg:
> I would think the default should be case-insensitive since in HTTP/1 header names should be case-insensitive. AIUI in HTTP/2 they must be lowercase.

Please note that we are solely talking about header *values* here, not header *names*, as @chkal outlined in the original question, so whatever HTTP/1 or /2 tell us here is irrelevant.

jamezp:
> > I would think the default should be case-insensitive since in HTTP/1 header names should be case-insensitive. AIUI in HTTP/2 they must be lowercase.
> 
> Please note that we are solely talking about header _values_ here, not header _names_, as @chkal outlined in the original question, so whatever HTTP/1 or /2 tell us here is irrelevant.

My apologies. For some reason I thought we were talking about headers :)

mkarg:
> > Please note that we are solely talking about header _values_ here, not header _names_, as @chkal outlined in the original question, so whatever HTTP/1 or /2 tell us here is irrelevant.
> 
> My apologies. For some reason I thought we were talking about headers :)

Now that we clarified, what is you opinion: Shall we go with predicates, with booleans, or with separate methods?

jamezp:
Let me make sure I understand this correct. The idea is that the predicate would be tested against the value(s) of a header. For example a header like:
```
Cache-Control: max-age=0, must-revalidate
```

The predicate would be tested against "max-age=0" and "must-revalidate"?

Sorry, If I'm not quite understanding.

jansupol:
I meant a default implementation such as 
```
    public boolean containsHeaderString(String name, Predicate<String> valuePredicate) {
        String property = getHeaderString(name);
        return property != null ? valuePredicate.test(property) : false;
    }
```
Should the `Predicate` handle the case sensitivity?

jamezp:
> I meant a default implementation such as
> 
> ```
>     public boolean containsHeaderString(String name, Predicate<String> valuePredicate) {
>         String property = getHeaderString(name);
>         return property != null ? valuePredicate.test(property) : false;
>     }
> ```
> 
> Should the `Predicate` handle the case sensitivity?

+1 This is what I envisioned for the default method.

mkarg:
> Let me make sure I understand this correct. The idea is that the predicate would be tested against the value(s) of a header. For example a header like:
> 
> ```
> Cache-Control: max-age=0, must-revalidate
> ```
> 
> The predicate would be tested against "max-age=0" and "must-revalidate"?
> 
> Sorry, If I'm not quite understanding.

Correct. But as a header can exist multiple times, the actual algorithm would be more complex.

mkarg:
> I meant a default implementation such as
> 
> ```
>     public boolean containsHeaderString(String name, Predicate<String> valuePredicate) {
>         String property = getHeaderString(name);
>         return property != null ? valuePredicate.test(property) : false;
>     }
> ```

That would be wrong. The predicate is to be tested per single value, not per comma-separated list.

> Should the `Predicate` handle the case sensitivity?

Yes, this is the idea of the predicate. So the caller can provide an arbitrarily complex predicate, testing much more than that.

jansupol:
> That would be wrong. The predicate is to be tested per single value, not per comma-separated list.

Hm....then perhaps it would make sense to provide a method that actually returns a list of comma-separated values when these methods are using it?

jamezp:
One option might be to use a `BiPredicate,String, String>` where the first argument is the key/name and the second argument is the optional value. For example in:
```
Cache-Control: max-age=0, must-revalidate
```

It would be something like `predicate.test("max-age", "0")` and `predicate.test("mcust-revalidate" null)`.

jamezp:
Thinking about my above suggestion though, that doesn't make the code real clean on the user side so maybe that's not a great idea.

mkarg:
> > That would be wrong. The predicate is to be tested per single value, not per comma-separated list.
> 
> Hm....then perhaps it would make sense to provide a method that actually returns a list of comma-separated values when these methods are using it?

This is exactly what I do want to prevent. First, it implies that all implementations must concatenate the values first just to let the application tokenize it afterwards, which is squandering performance, particularly if multiple filters need to check header values, so it would run repeatedly. Second, it makes the application (or to be more precise: filter) code much more cluttered with string processing details while the filter author just wants to deal with *his own business* and offload most technological questions to JAX-RS.

The actual intention of this PR is that filters are totally clean to write, e. g. like this, while providing maximum performance by *preventing* string operations *as much as possible*:

```java
if (context.containsHeaderValue("cache", "no-store".equalsIgnoreCase))
    return;
```

In this example, no useless string concatenation occurs at all. The runtime can implement `containsHeaderValue` in the fastest possible way fitting *its particular kind* of collection. The filter code itself is clean to read: "If there is no-store found *somewhere* in *some* cache header, then skip this filter completely". 

As I had to write lots of common-purpose filters in the past that have to run in "unkown" combinations (the filter author is not the application author / deployer), this is exactly what I needed in rather each of them: A *clean* solution that provides *high* performance and does neither *repeat* itself nor *clutter* my code.

mkarg:
> One option might be to use a `BiPredicate,String, String>` where the first argument is the key/name and the second argument is the optional value. For example in:

I do not see the benefit of pushing the *name* into the predicate, actually, and it makes writing *typical* requests more complex. For example, a mono-predicate method is perfectly simple, but try to write the same with your proposed bi-predicate method:

```java
context.containsHeaderValue("cache", "no-store"::equalsIgnoreCase)
```

Also, performance of the mono-predicate method is way better as the runtime can use the name as the key in a map, while it must linearly *iterate* over all entries in the bi-predicate method.

mkarg:
> > I meant a default implementation such as
> > ```
> >     public boolean containsHeaderString(String name, Predicate<String> valuePredicate) {
> >         String property = getHeaderString(name);
> >         return property != null ? valuePredicate.test(property) : false;
> >     }
> > ```
> 
> That would be wrong. The predicate is to be tested per single value, not per comma-separated list.

A *correct* implementation with some first performance optimization could be this, but again, there is much room left open to improve performance, as it still forces string concatenation just to tokenize it afterwards, and it still does not respect the actual kind of collection used by the runtime:

```java
default public boolean containsHeaderValue(String name, Predicate<String> valuePredicate) {
	var values = getHeaderString(name);
	if (values == null)
		return false;
	var tokens = new StringTokenizer(values, ",");
	while (tokens.hasMoreTokens()) {
		var value = tokens.nextToken().strip();
		if (valuePredicate.test(value))
			return true;
	}
	return false;
}
```

Optimizing the code further implies prevention of the initial concatention, hence dealing with raw header objects, hence dealing with `HeaderDelegate::toString` -- which we simply **cannot**, because *a default method* cannot obtain a header delegate! **This fact, and the length and complexity of the resulting algorithm are the drivers behind this PR!**

jamezp:
Given the above example I think it does make sense to keep the method abstract. Implementations are likely going to have to override this anyway to be efficient. It also allows the spec not to have to set a rule on the case of header values,  which could widely differ.

jelemux:
> ```java
> context.containsHeaderValue("cache", "no-store"::equalsIgnoreCase)
> ```
@mkarg I really like this solution and would like to express my support. It's simple, readable and it leaves the users the freedom to do what they want with it. Imho a lot more software should be written like that.

mkarg:
Ok it seems we reached an agreement how the new method should look like, so I think it is time end the draft state now.

mkarg:
Committers, please review and vote. Thanks. :-)

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientResponseContext.java
```diff
@@ -88,6 +89,19 @@ public interface ClientResponseContext {
      */
     public String getHeaderString(String name);
 
+    /**
+     * Checks whether a header with a specific name and value (or item of the comma-separated value list) exists.
+     *
+     * @param name the message header.
+     * @param valuePredicate value must fulfil this predicate.
+     * @return {@code true} if and only if a header with the provided name exists having either the exact value or whose
+     * comma-separated header string contains value as a whole word.
```
spericas, line 98: The phrase "whole word" seems rather imprecise. Could we describe this better in terms of spaces etc? Perhaps add an example in the doc too.

mkarg: I leftover from the original proposal. You are right, this should get adapted to the latest method signature. What about `...comma-separated header string contains one ore more separated values each matching the predicate.`? A bit clumsy but I rather precide.

mkarg: I changed the JavaDocs. Hope it now is unambiguous.

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
```diff
@@ -192,6 +193,29 @@ public default boolean hasProperty(String name) {
      */
     public String getHeaderString(String name);
 
+    /**
+     * Checks whether a header with a specific name and value (or item of the comma-separated value list) exists.
+     *
+     * Each single header value is converted to String using a {@link jakarta.ws.rs.ext.RuntimeDelegate.HeaderDelegate} if one
```
spericas, line 199: For consistency with other text, "Each single header value ..." -> "Each non-string header value ...". We should not use a delegate if already a String.

mkarg: This is the original text taken from `#getHeaderString`, so do you want to change *that* text, too? I deliberately used it unchanged to prevent any inconsistencies, actually.

jansupol:
I am not convinced about this method. It could be good for some headers, such as `Cache-Control` but not for others, that use the comma separator for not separating the header values. [HTTP RFC, Section 4.2](https://www.rfc-editor.org/rfc/rfc2616#section-4.2) says:
> Multiple message-header fields with the same field-name MAY be
   present in a message if and only if the entire field-value for that
   header field is defined as a comma-separated list [i.e., #(values)].

So there is plenty of HTTP headers that use commas for other purposes than separator. All dates containing headers is such an example (with syntax `Date: day-name, day month year hour:minute:second GMT`). `Set-Cookie`, or `Cookie` are also the exmples. `Cookie` [specifies value](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie) as  
> A list of name-value pairs in the form of <cookie-name>=<cookie-value>. Pairs in the list are separated by a semicolon and a space ('; ').

For the generic usage, I'd suggest a separator (nullable for no split) to separate the header values.

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
spericas: Well, I think `#getStringHeaders` in the same class is more precise, isn't it? Seems important to use the _non-string_ qualifier.

spericas:
@jansupol Seems like a reasonable suggestion, and we can still use commas by default.

mkarg:
> For the generic usage, I'd suggest a separator argument (nullable for no split) to separate the header values.

Good idea, I will add this in the next iteration of this PR, stay tuned. :-)

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
mkarg: Sure, but `#getStringHeader(String name)` unfortunately is *missing* the `non-string` token, hence I was asking if you want to thange it *there*, too. Do you?

mkarg:
> > For the generic usage, I'd suggest a separator argument (nullable for no split) to separate the header values.
> 
> Good idea, I will add this in the next iteration of this PR, stay tuned. :-)

On the other hand, we are using a comma (and exactly a comma but nothing else) in `#getHeaderString`. Maybe we should extend that one, too with an optional concatenation character in analogy?

jansupol:
> Maybe we should extend that one, too with an optional concatenation character in analogy?

Sounds good.

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientResponseContext.java
mkarg: @spericas Agreed?

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
mkarg: Assuming you *would* add the token `non-string` to `#getStringHeader(String name)`, too, I added it two both code locations in https://github.com/jakartaee/rest/pull/1066/commits/eb18e39ce3824724f33116f01c4a0cddc64bfbb7.

Ok?

mkarg:
> > For the generic usage, I'd suggest a separator argument (nullable for no split) to separate the header values.
> 
> Good idea, I will add this in the next iteration of this PR, stay tuned. :-)

Implemented by https://github.com/jakartaee/rest/pull/1066/commits/b5f0a9ee051c7fd46cfdf03688095663cd6b324b.

mkarg:
> @jansupol Seems like a reasonable suggestion, and we can still use commas by default.

How do you envision such a default? By an additional variant of the same method having just two parameters?

mkarg:
> > Maybe we should extend that one, too with an optional concatenation character in analogy?
> 
> Sounds good.

What would you like more, breaking backwards compatibility by adding a new argument to the existing method, or adding a second variant of the same method?

spericas:
> > > Maybe we should extend that one, too with an optional concatenation character in analogy?
> > 
> > 
> > Sounds good.
> 
> What would you like more, breaking backwards compatibility by adding a new argument to the existing method, or adding a second variant of the same method?



> > @jansupol Seems like a reasonable suggestion, and we can still use commas by default.
> 
> How do you envision such a default? By an additional variant of the same method having just two parameters?

Yes, if we think using commas is a sufficiently common case --as suggested by your original PR.

mkarg:
> > How do you envision such a default? By an additional variant of the same method having just two parameters?
> 
> Yes, if we think using commas is a sufficiently common case --as suggested by your original PR.

Done. :-)

mkarg:
@jamezp @jansupol @chkal Kindly requesting approval of this PR. :-)

jamezp (approved):
LGTM

spericas (approved):

chkal (approved):

PR #1071, opened 2021-12-22
Adopt OpenJDK now is Eclipse Temurin

mkarg:
Adopt OpenJDK is called Eclipse Temurin since a while. There will be no new JDKs published with the adopt label, but only with the temurin label. So our builds should use Temurin instead of Adopt now. 


jansupol (approved):
Nice!

PR #1075, opened 2021-12-23
Replaced custom code by JRE standards

mkarg:
It makes not much sense to maintain custom code if there is a simple JRE replacement. :-)

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/NewCookie.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/NewCookie.java
@@ -17,6 +17,7 @@
 package jakarta.ws.rs.core;
 
 import java.util.Date;
+import java.util.Objects;
 
 import jakarta.ws.rs.ext.RuntimeDelegate;
 import jakarta.ws.rs.ext.RuntimeDelegate.HeaderDelegate;
@@ -423,14 +424,8 @@ public class NewCookie extends Cookie {
      */
     @Override
     public int hashCode() {
-        int hash = super.hashCode();
-        hash = 59 * hash + (this.comment != null ? this.comment.hashCode() : 0);
-        hash = 59 * hash + this.maxAge;
-        hash = 59 + hash + (this.expiry != null ? this.expiry.hashCode() : 0);
-        hash = 59 * hash + (this.secure ? 1 : 0);
-        hash = 59 * hash + (this.httpOnly ? 1 : 0);
-        hash = 59 * hash + (this.sameSite != null ? this.sameSite.ordinal() : 0);
-        return hash;
+        return Objects.hash(this.getName(), this.getValue(), this.getVersion(), this.getPath(), this.getDomain(),
+                this.comment, this.maxAge, this.expiry, this.secure, this.httpOnly, this.sameSite);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


jaxrs-api/src/main/java/jakarta/ws/rs/core/NewCookie.java
```diff
@@ -423,14 +424,8 @@ public String toString() {
...
     public int hashCode() {
-        int hash = super.hashCode();
-        hash = 59 * hash + (this.comment != null ? this.comment.hashCode() : 0);
-        hash = 59 * hash + this.maxAge;
-        hash = 59 + hash + (this.expiry != null ? this.expiry.hashCode() : 0);
-        hash = 59 * hash + (this.secure ? 1 : 0);
-        hash = 59 * hash + (this.httpOnly ? 1 : 0);
-        hash = 59 * hash + (this.sameSite != null ? this.sameSite.ordinal() : 0);
-        return hash;
+        return Objects.hash(this.getName(), this.getValue(), this.getVersion(), this.getPath(), this.getDomain(),
```
arjantijms, line 427: As there's no other object in scope here, my preference would be to omit the explicit qualification with "this.".

Shorter is better, less code is better code, etc.

jaxrs-api/src/main/java/jakarta/ws/rs/core/Variant.java
```diff
@@ -220,14 +217,13 @@ public boolean equals(final Object obj) {
...
+        if (!Objects.equals(this.language, other.language)) {
             return false;
         }
-        if (this.mediaType != other.mediaType && (this.mediaType == null || !this.mediaType.equals(other.mediaType))) {
+        if (!Objects.equals(this.mediaType, other.mediaType)) {
             return false;
         }
-        // noinspection StringEquality
-        return this.encoding == other.encoding || (this.encoding != null && this.encoding.equals(other.encoding));
+        return (!Objects.equals(this.encoding, other.encoding));
```
jansupol, line 226: Does the new code return the negation of the original code?

mkarg: Good catch! A typo, indeed.

jaxrs-api/src/main/java/jakarta/ws/rs/core/NewCookie.java
mkarg: I respect your personal style preference, but I do not share it, and the original code already used "this". So feel free to open a PR for removal of "this" whereever you like it, but let's concentrate in *this* PR on the its own topic.

jaxrs-api/src/main/java/jakarta/ws/rs/core/Variant.java
mkarg: Thanks for your review! I fixed the typo in https://github.com/eclipse-ee4j/jaxrs-api/pull/1075/commits/3c9220686c543dd9d0b8937819b0f9911d130d6c.

jaxrs-api/src/main/java/jakarta/ws/rs/core/NewCookie.java
spericas: I would drop `this` as well. May have been there before if method was auto-generated, but otherwise, we should only need that in constructors really.

mkarg: Sigh. I hoped we could agree to separate our personal likes and dislikes from the general benefit of a PR ("SoC"). Anyways, removed "this" in https://github.com/eclipse-ee4j/jaxrs-api/pull/1075/commits/666d7513ff6a251f7263bb03ddab0c84791257c5.

chkal (approved):

spericas (approved):

PR #1077, opened 2022-01-08
Fixed: jersey-tck fails on JRE 17 due to Unsafe class

mkarg:
`mvn --file jersey-tck/pom.xml verify` fails on JRE 17.

The reason is that it uses GlassFish 6.1.0, which is not compatible with JRE 17 (uses `Unsafe` class).

The solution is to replace it by GlassFish 6.2.3.

### DIFF
--- a/jersey-tck/pom.xml
+++ b/jersey-tck/pom.xml
@@ -32,7 +32,7 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <jersey.version>3.0.0</jersey.version>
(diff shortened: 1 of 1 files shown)


mkarg:
@jelemux Please check out this proposed solution, it should fix the problem you encountered earlier today. At least it did for me. :-)

jelemux:
@mkarg Sorry, still does not fix my problem. I'm beginning to suspect that it might be something different entirely. I backtracked the commit that introduced it and opened #1078. I'll try to find out more and will post the updates in the issue.

chkal (approved):

jansupol (approved):

PR #1097, opened 2022-02-25
Applications are required to have a public no-arg constructor. The Ec…

### LINKED ISSUE #1099: TCK Challenge: Change JsonbContextProviderIT.EchoApplication to be public
I filed a PR https://github.com/jakartaee/rest/pull/1097 to make the `JsonbContextProviderIT.EchoApplication` public rather than private. This was given a -1 from @mkarg, but I'd like to challenge this.

RESTEasy has required constructors for applications be public for at least 11 years, according the git history. I don't see any reason why this should change, especially for a TCK test. If RESTEasy starts allowing non-public constructors we now have to support this going forward and I'm opposed to this.

jamezp:
…hoApplication must be public.

If this needs an issue file please let me know.

resolves #1099

### DIFF
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/spec/client/exceptions/ClientExceptionsIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/spec/client/exceptions/ClientExceptionsIT.java
@@ -197,7 +197,7 @@ public class ClientExceptionsIT {
     }
 
     @ApplicationPath("")
-    private static final class StatusApplication extends Application {
+    public static final class StatusApplication extends Application {
 
         @Override
         public Set<Class<?>> getClasses() {
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/spec/contextprovider/JsonbContextProviderIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/spec/contextprovider/JsonbContextProviderIT.java
@@ -162,7 +162,7 @@ public final class JsonbContextProviderIT {
     }
 
     @ApplicationPath("")
-    private static final class EchoApplication extends Application {
+    public static final class EchoApplication extends Application {
         @Override
         public final Set<Class<?>> getClasses() {
             return Collections.singleton(EchoResource.class);


jamezp:
@mkarg I didn't see anything in the specification which directly referenced an `Application`. However, resources and providers are required to have a public constructor. It seems applications should as well.

mkarg:
As you did not find an unumbiguous spec wording, I do not see why the TCK needs to be changed. Unless it is explicitly forbidden by the spec, it is clearly valid to do what the TCK currently does. If you want to change the spec, please open an issue for a spec change. Thanks.

jamezp:
The TCK did not include a test like this before and has not tested a non-public application before. While it may be ambiguous, I don't see how it could be considered good practice. If resources and providers explicitly state they require public constructors why would it be assumed an application can be private?

>3.1.2. Constructors
>
>Root resource classes are instantiated by the JAX-RS runtime and MUST have a public constructor for which the JAX-RS >runtime can provide all parameter values. Note that a zero argument constructor is permissible under this rule.

jansupol:
[Section 11.2.3](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#cdi):
> In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource classes, providers and Application subclasses.

As argued before with the test to instantiate the class using the most arguments constructor - When CDI is used, we need to be able to instantiate the classes with public no-arg constructors. RestEasy was able to turn off CDI with a proper beans.xml that is added to the tests.

jamezp:
> [Section 11.2.3](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#cdi):
> 
> > In a product that supports CDI, implementations MUST support the use of CDI-style Beans as root resource classes, providers and Application subclasses.
> 
> As argued before with the test to instantiate the class using the most arguments constructor - When CDI is used, we need to be able to instantiate the classes with public no-arg constructors. RestEasy was able to turn off CDI with a proper beans.xml that is added to the tests. But we are approaching CDI-only injection times...

I will say RESTEasy has a ways to go to truly support CDI like it should be supported IMO. That said this test does not include a `beans.xml` file and does not qualify for an empty `beans.xml` file: https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#what_classes_are_beans.

I really don't understand why making this `public` is a big deal. It changes nothing about what the test is actually testing.

jimma:
Except this `EchoApplication` and `StatusApplication` in `./src/main/java/ee/jakarta/tck/ws/rs/spec/client/exceptions/ClientExceptionsIT.java`, other tck application classes are all public.
Is there particular reason we set ` EchoApplication` and `StatusApplication' as `private` ?

mkarg:
The TCK test are spec compliant. The spec is the sole source of truth, not the TCK. So if we want to enforce public, we break backwards compatibility of existing applications running on Jersey. Hence, that change is a spec change. Hence we need a spec change proposal. Unless that proposal is agreed, the TCK is correct as-is, including `private`.

jansupol:
> So if we want to enforce public, we break backwards compatibility of existing applications running on Jersey

No, that is not what this is about. This is about not enforcing private so that each implementation can instantiate the class.

spericas (approved):

arjantijms (approved):
Agree with this. A private Application is not testing any specific spec text or javadoc, should we should not just introduce it at this point.

spericas:
@mkarg Can you please review your vote?

jansupol:
> Except this EchoApplication and StatusApplication in ./src/main/java/ee/jakarta/tck/ws/rs/spec/client/exceptions/ClientExceptionsIT.java, other tck application classes are all public.

@jamezp @jimma Are you going to add the second Application to this PR?

jamezp:
> Except this EchoApplication and StatusApplication in ./src/main/java/ee/jakarta/tck/ws/rs/spec/client/exceptions/ClientExceptionsIT.java, other tck application classes are all public.
>
> Are you going to add the second Application to this PR?

@jansupol This one is kind of weird. It's not in the [downloaded artifact](https://download.eclipse.org/ee4j/jakartaee-tck/jakartaee10/staged/eftl/jakarta-restful-ws-tck-3.1.0.zip). It's also got a `@since 4.0` so I'm not sure if it's intended to target 3.1 or not. However, I'm happy to make the change there too.

mkarg:
I am sorry but as long as the spec does not clearly say that Application MUST be public, any application programmer is free to use private Application, so the TCK ensures this is possible. I have not problem chaning the spec to allow this, but first spec change, *then* TCK change.

mkarg:
To quickly resolve the current situation I propose that @jamezp adds a commit into this PR that changes the wording of the spec accordingly. I am +1 for the spec change + TCK change, but -1 for just a TCK change.

spericas:
> I am sorry but as long as the spec does not clearly say that Application MUST be public, any application programmer is free to use private Application, so the TCK ensures this is possible. I have not problem chaning the spec to allow this, but first spec change, _then_ TCK change.

Does this imply that we need TCK tests for all things that the spec is not clear about? This makes no sense, and it would be a never ending task. As explained by others, the TCKs test spec assertions, and this is not one (al least not yet). Only this TCK needs updating for 3.1.

jamezp:
I've updated the PR to also fix the `ClientExceptionsIT`.

jansupol (approved):
I do not see any problems with this PR.

jimma:
+1 for this change.

mkarg (commented):
Turning my vote form -1 to 0 as I do not want to stand in the way for releasing 3.1. For 4.0 I would propose we actively add Application to the list of classes that MUST have a public no-args constructor, as apparingly this is an essential information that application programmers MUST be sure about for application portability - it works in Jersey but not in RESTeasy, and ANYTHING that a Java programmer MUST NOT do to achieve portability MUST be unambiguously told in the spec. That might be a long list, but application programmer MUST know that for maximized portability.

mkarg (approved):

arjantijms:
Thanks everyone!

jamezp:
Thank you all! This is much appreciated.

PR #1098, opened 2022-02-26
Allow the host to be set and do not require the protocol to return HT…

jamezp:
…TP in all uppercase.

Again if this needs an issue file please let me know.

For the `HOST` change I don't think it's generally safe to assume `localhost` will work on all environments, e.g. CI.

### DIFF
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/sebootstrap/SeBootstrapIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/sebootstrap/SeBootstrapIT.java
@@ -32,6 +32,7 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 
 import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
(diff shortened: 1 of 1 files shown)


mkarg (changes requested):
-1 unless there is a proof that `localhost` is not supported on all Java platforms

-1 as `HTTP` in all uppercase *is deliberately mandatory*  to simplify application code using this API

jamezp:
The host I've got no strong argument for. I can see where using localhost is fine, but I've definitely seen issues in some environments before. Maybe not a big deal here though.

The HTTP I respecivtely disagree with.

mkarg:
Regarding "HTTP" the explicit uppercase is enforced by the commonly accepted API already. This is *not* the schema string, this is the protocol name. If e. g. a product supports HTTP/3 then the schema still is lowercase "http" but to enforce it, the protocol name needs to be "HTTP/3" vs "HTTP". Maybe it is more clear now?

Please note that the time to change the API is over. It is discussed and accepted. The TCK only enforces the API as-is.

jamezp:
I can't say I fully agree, but it's not really that big of a deal so I'll just close this.

PR #1102, opened 2022-03-04
add dependencies to run tck independently

alwin-joseph:
Adding dependency jars in jersey-tck/pom.xml to run the module independently. 
Requesting fast-track. 
@spericas @jansupol

### DIFF
--- a/jersey-tck/pom.xml
+++ b/jersey-tck/pom.xml
@@ -38,9 +38,9 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <jersey.version>3.1.0-M3</jersey.version>
-        <glassfish.container.version>6.2.3</glassfish.container.version>
+        <glassfish.container.version>6.2.5</glassfish.container.version>
         <glassfish.home>${project.build.directory}/glassfish6</glassfish.home>
-        <jakarta.platform.version>9.1.0</jakarta.platform.version>
+        <jakarta.platform.version>10.0.0-RC1</jakarta.platform.version>
         <junit.jupiter.version>5.7.2</junit.jupiter.version>
         <jakarta.rest.version>3.1.0</jakarta.rest.version>
         <tck.artifactId>jakarta-restful-ws-tck</tck.artifactId>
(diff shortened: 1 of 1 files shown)


mkarg (commented):
I abstain from voting as I cannot judge the exact impact of the modifications. I likely would cast a +1 if there would be some brief explanation why these changes are needed.

alwin-joseph:
> I abstain from voting as I cannot judge the exact impact of the modifications. I likely would cast a +1 if there would be some brief explanation why these changes are needed.

Hi @mkarg , These 2 jars (jakarta.ws.rs-api and hamcrest) are certainly required for the tests to pass . We would need to use jersey-tck/ as an independent module to run the test. When api jars are available in staging/maven we will fetch from the remote location rather than by `mvn install` from this repo.  

To clarify further, we will perform below steps to run the tck against jersey once all jars are released and available in maven repository: 
```
-wget https://download.eclipse.org/ee4j/jakartaee-tck/jakartaee10/staged/eftl/jakarta-restful-ws-tck-3.1.0.zip
-unzip jakarta-restful-ws-tck-3.1.0.zip
-mvn install:install-file -DcreateChecksum=true -Dpackaging=jar -Dfile=jakarta-restful-ws-tck-3.1.0.jar -DgroupId=jakarta.ws.rs -DartifactId=jakarta-restful-ws-tck -Dversion=3.1.0
-git clone https://github.com/jakartaee/rest
-mvn -f rest/jersey-tck/pom.xml verify (-Pstaging)
```

mkarg:
I understand this, but I doubt that running jersey-tck as a *standalone* module is a target of JAX-RS. I invented jersey-tck just as an example how to use jaxrs-tck and used Jersey just by incident; also the target is to be able to do `mvn -am -pl jersey-tck verify` to *test-drive* the JAX-RS tck by the developers of the JAX-RS tck. This is *not* a replacement for Jersey's *own* driver. My actual target was (and this is written in the readme) that all vendors write (and publish) *their own* standalone module in *their own* repository. Hence I do not see why the *JAX-RS* repository should support your use case? It would be better if Jersey provides its own POM in its own repo that *builds upon* the jersey-tck module, instead of *directly using it* standalone.

spericas (approved):

chkal (approved):
I agree with @mkarg that the tck module in the spec repo isn't meant to be consumed by implementation projects and is more like a "demo" for how to run the TCK, and therefore it would be preferable if Jersey creates such a module in the Jersey repo. However, I also think that the proposed changes make sense, so +1 for merging.

chkal:
Just for reference: eclipse-ee4j/jersey#5000

jansupol (approved):

jansupol:
The TCK always used to contain a setting that was preset to run the TCK against RI. We no longer have an RI, but keeping a pom to be able to run the standalone TCK against any CI (it does not be Jersey) would make sense to me. Any vendor could come and see how it is supposed to work.

Jakarta REST standalone TCK is a new concept (the tests are separated from the jakartaeetck repo), that is already used for JSON-P and JSON-B, too. To me it would make sense to contain similar information to JSON-P,B to run the TCKs to keep some uniformity level that helps vendors to run the TCK,

mkarg:
Guys, remember that it actually was *me* who originally created the `jersey-tck` folder, so there is no need to speculate about the reason why *I* did that. ;-) The reason solely was still is like I described, and since there is not RI anymore, we should replace jersey-tck by demo-tck which is more like a common template (maybe a Maven BOM?) instead of a full standalone Jersey-specific application.

PR #1103, opened 2022-03-06
Removed unnecessary workarond for ephemerial port

mkarg:
This PR removes an unnecessary workaround: There is not need to check the port number, as nobody ever has seen any JRE on any OS that returned a non-ephemerial port. Oracle's OpenJDK team even disputes that this could happen ever, and there is no proof that there would be a bug. Hence we should not fix something that is not broken.

### DIFF
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/common/util/JaxrsUtil.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/common/util/JaxrsUtil.java
@@ -35,15 +35,10 @@ import java.util.TimeZone;
 
 public abstract class JaxrsUtil {
 
-  public static final int unprivilegedPort() throws IOException {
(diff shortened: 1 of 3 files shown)


alwin-joseph:
> This seems to be unnecessary pushing the TCK on an edge. It has been reported to be an issue and unless the original reporter (@alwin-joseph) confirms there is no issue, I'd stay safe here.

There was no evidence in the issue https://github.com/jakartaee/rest/issues/1088 that `ServerSocket(0)` was returning port < 1024. The only two failing tests in the issue were fixed by the Jersey upgrade to M3 , those tests were not using the `someFreeIpPort` method modified here.

alwin-joseph (approved):

jansupol (approved):
I am fine with the PR then.

spericas (approved):

chkal (approved):

pdudits (approved):

arjantijms (approved):

PR #1114, opened 2022-05-14
Support for Maven Wrapper

mkarg:
Since long time this project uses Maven as its build tool. Since short time there is an [**official** Maven Wrapper](https://maven.apache.org/wrapper/) release provided by the Apache Maven team, i. e. no third-party. To make it as easy as possible for new contributors to get onboard, I think it is a good idea if we support the official Maven Wrapper. The benefit is that nobody needs to manually download and install Maven on his local disk. Instead, simply invoke `mvnw verify` to get this done automatically.


arjantijms (approved):
si si, why not huh?

chkal (approved):

PR #1117, opened 2022-06-29
exclude additional Jakarta XML Binding tests that were previously excluded in the Jakarta EE Platform TCK

WhiteCat22:
While running the TCK without Jakarta XML Binding present, we've been running into failures with these tests that were previously excluded in the Jakarta EE Platform TCK.

### DIFF
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/core/linkjaxbadapter/JAXRSClientIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/core/linkjaxbadapter/JAXRSClientIT.java
@@ -37,6 +37,8 @@ import jakarta.xml.bind.JAXBException;
 import jakarta.xml.bind.Marshaller;
 import jakarta.xml.bind.Unmarshaller;
 
+import org.junit.jupiter.api.Tag;
+
 /*
  * @class.setup_props: webServerHost;
  *                     webServerPort;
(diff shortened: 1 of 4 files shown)


Emily-Jiang:
@spericas @jansupol @jim-krueger please review asap as it blocks Jakarta EE 10 Core Profile ballot. Thank you!

jansupol:
The test `jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/spec/provider/standardhaspriority/JAXRSClientIT.java` seems to suspiciously exclude tests that shall not be handled by a JAX-B framework, the media type is TEXT/PLAIN.

The Spec, Section 4.2.4 defines default providers for 
> java.lang.Boolean, java.lang.Character, java.lang.Number Only for text/plain. Corresponding
primitive types supported via boxing/unboxing conversion

They should not be handled by JAX-B.

jansupol:
The other three files seem fine by me.

jansupol:
Same with 
> MultivaluedMap<String,String> Form content (application/x-www-form-urlencoded).

The implementation should contain a provider, no matter the JAX-B.

jansupol:
> it's the reference in ee.jakarta.tck.ws.rs.spec.provider.standardhaspriority.Resource that causes the servlet to fail to initialize.

I see. That sounds correct. The tests should be split into two apps for the future version, then. For the time being, let's mark all the tests.

jansupol:
I have created issue #1118 for this.

jansupol (approved):

arjantijms (approved):

spericas (approved):
Given that this is a blocker and it only affects the TCK, we should fasttrack it to resolve this ASAP.

arjantijms:
>  we should fasttrack it to resolve this ASAP.

+1

9 days and 3 reviews seem enough, right? I'm never sure why Jakarta REST always needs so many reviews and so much times. For basically all other specs there's either no reviews at all, or just one reviews. It doesn't look like those specs are so much worse, are they?

spericas:
Merged

mkarg:
> 9 days and 3 reviews seem enough, right? I'm never sure why Jakarta REST always needs so many reviews and so much time. For basically all other specs there's either no reviews at all, or just one review. It doesn't look like those specs are so much worse, are they?

All specs are individually governed and your frequent rants about it do not change anything.

PR #1133, opened 2022-09-19
Deprecating "APPLICATION/SVG+XML" for Removal

mkarg:
`APPLICATION/SVG+XML` is not recognized by modern browsers, as the official MIME type of SVG graphics is `IMAGE/SVG+XML`.

As least RESTful applications produce SVG at all, it makes not much sense to keep this and add the new variant. Instead, a custom String or a custom library should replace the outdated elements.

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/MediaType.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/MediaType.java
@@ -88,13 +88,13 @@ public class MediaType {
     public static final String APPLICATION_SVG_XML = "application/svg+xml";
     /**
      * A {@link MediaType} constant representing {@value #APPLICATION_SVG_XML} media type.
-     */
-    public static final MediaType APPLICATION_SVG_XML_TYPE = new MediaType("application", "svg+xml");
-    /**
-     * A {@code String} constant representing {@value #APPLICATION_JSON} media type.
      * @deprecated since 4.0, use a custom {@code MediaType} instead. Will be removed in a future release of this API.
      */
     @Deprecated(forRemoval = true)
+    public static final MediaType APPLICATION_SVG_XML_TYPE = new MediaType("application", "svg+xml");
+    /**
+     * A {@code String} constant representing {@value #APPLICATION_JSON} media type.
+     */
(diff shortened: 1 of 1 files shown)


jaxrs-api/src/main/java/jakarta/ws/rs/core/MediaType.java
```diff
@@ -82,15 +82,19 @@ public class MediaType {
...
     public static final String APPLICATION_SVG_XML = "application/svg+xml";
     /**
      * A {@link MediaType} constant representing {@value #APPLICATION_SVG_XML} media type.
      */
     public static final MediaType APPLICATION_SVG_XML_TYPE = new MediaType("application", "svg+xml");
     /**
      * A {@code String} constant representing {@value #APPLICATION_JSON} media type.
+     * @deprecated since 4.0, use a custom {@code MediaType} instead. Will be removed in a future release of this API.
      */
+    @Deprecated(forRemoval = true)
```
spericas, line 97: Why is this part of this PR? Title states scope is SVG.

jamezp: My assumption is this was intended for `APPLICATION_SVG_XML_TYPE` which is the variable above this one.

mkarg: Oops, sorry. James ist right, somehow the change slipped into the wrong lines.

jamezp (approved):

spericas (approved):

jansupol:
Shall we not add the `image/svg+xml` instead?

mkarg:
> Shall we not add the `image/svg+xml` instead?

This was my first idea, but then I actually didn't see what it has to do with what JAX-RS is made for. I couldn't imagine such a lot of JAX-RS applications processing SVG entities that it would make worth such a step, IMHO. What I would prefer would be that "someone" starts a separate open source project implementing a JAX-RS `Feature` to provides SVG support using MBW and MBR etc. Such a library should provide a constant for this media type.

mkarg:
@chkal Kindly requesting approval. :-)

chkal (approved):

PR #1164, opened 2023-07-24
TCK Close the original stream of setEntityStream

### LINKED ISSUE #1163: TCK Challenge: ee.jakarta.tck.ws.rs.ee.rs.container.responsecontext.JAXRSClientIT#setEntityStreamTest
The test uses the original stream to write into it with a new stream set on jakarta.ws.rs.container.ContainerResponseContext#setEntityStream. The specification details that the implementation must close the output stream, but in the way the test is written, the original stream should also be closed:

https://github.com/jakartaee/rest/blob/release-3.1.x/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/container/responsecontext/ResponseFilter.java#L300-L328

radcortez:
- Fixes #1163

### DIFF
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/container/responsecontext/ResponseFilter.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/container/responsecontext/ResponseFilter.java
@@ -317,6 +317,11 @@ public class ResponseFilter extends TemplateFilter {
         write(b);
       }
 
+      @Override
+      public void close() throws IOException {
+        stream.close();
+      }
+
       public final byte[] intToByteArray(int value) {
         return new byte[] { (byte) (value & 0xff) };
       }


jamezp:
I can't link the issues, but this resolves #1163

jansupol:
@alwin-joseph @jamezp is this the right branch? Or should it be PR to master?

jamezp:
IMO it just depends. There definitely needs to be a master one, but if there is any chance a new 3.1.x release of the TCK would be done, I do think it makes sense to add it to this branch as well.

alwin-joseph:
Referring to https://jakarta.ee/committees/specification/tckprocess/ what I understand is that 
- the challenge resolution is usually test exclusion. 
- But there is another option to modify the test during challenge for a service release which is explained starting : 
"_As another alternative to excluding a challenged test, it may be possible to adjust the test validation logic to “expand” the validation check_..". Does this fix qualify 
The idea is that the challenge resolution should not cause already certified implementations to fail the modified test. They mention ".._modified TCK should be run against at least one such implementation_..".

To conclude I think we have below options
1. exclude the test for 3.1.4 and fix for 4.0.0 or
2. fix the test for 3.1.4 , make sure already passed implementations do not fail with this change. Are there any compatible implementations other than Jersey for 3.1?

Please correct me if I made any incorrect assumptions.

jamezp:
> fix the test for 3.1.4 , make sure already passed implementations do not fail with this change. Are there any compatible implementations other than Jersey for 3.1?

Yes, RESTEasy should be compatibile. It was accepted here https://github.com/jakartaee/rest/issues/1111 and has been shipping with WildFly.

radcortez:
I'm fine if we can just exclude the test to certify and use 3.1.3. This is also RESTEAsy, but with Vert.x.

Now sure why this was not an issue before. I guess some runtimes may be closing the stream, but this is not reliable.

spericas:
> In that case, we can redirect this PR to the master branch. Once the challenge #1163 is accepted we will exclude the test in a new service release (3.1.4).
> 
> Thoughts ?

I'm good with this plan

alwin-joseph:
@radcortez Can you redirect this PR to branch release-4.0 or master. Will create another change to exclude the test for 3.1.4 TCK.

spericas:
@alwin-joseph Until we decide how to proceed with the branches, we should target PR's such as this one to _both_ branches.

alwin-joseph:
Gentle reminder, the issue is resolved by excluding the test in 3.1.4 release in https://github.com/jakartaee/rest/pull/1181.
This PR should be redirected to release-4.0 and master as per previous discussion.

jansupol (approved):

radcortez:
> This PR should be redirected to release-4.0 and master as per previous discussion.

I've changed the base to `release-4.0`. Thanks!

spericas (approved):

PR #1166, opened 2023-07-25
Clarification about Response.readEntity(XXX) 

NicoNes:
This PR replaces this one https://github.com/jakartaee/rest/pull/1001.  
Exact same content but targeting branch `release-4.0`

-- Nicolas

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
@@ -74,16 +74,20 @@ public abstract class Response implements AutoCloseable {
     public abstract StatusType getStatusInfo();
 
     /**
-     * Get the message entity Java instance. Returns {@code null} if the message does not contain an entity body.
-     * <p>
-     * If the entity is represented by an un-consumed {@link InputStream input stream} the method will return the input
-     * stream.
-     * </p>
+     * Get the message entity Java instance. Returns {@code null} if the message does not contain an entity body, otherwise
+     * behaves as follow:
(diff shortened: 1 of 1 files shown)


jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
```diff
@@ -232,19 +232,18 @@ protected Response() {
...
 
     /**
-     * Buffer the message entity data.
+     * Buffer the message entity input stream.
      * <p>
-     * In case the message entity is backed by an unconsumed entity input stream, all the bytes of the original entity input
-     * stream are read and stored in a local buffer. The original entity input stream is consumed and automatically closed
-     * as part of the operation and the method returns {@code true}.
+     * In case the message entity is backed by an input stream not consumed at all, all the bytes of that input
+     * stream are read (i.e until <code>inputStream.read() == -1</code>) and stored in a local buffer. That input
```
spericas, line 238: Overall, this looks like a step in the right direction. However, I think the phrase "not consumed at all" is a bit vague. If we mean, "not a single byte from the input stream has been read", we should be more explicit IMO. This phrase is used multiple times.

spericas: @NicoNes Could you review and respond to the comment?

NicoNes: @spericas I agree we can be more exeplicit.  
What about saying " In case the message entity is backed by an input stream of which no byte has been read yet" ?

PR #1168, opened 2023-08-12
Compiling for Java 21

### LINKED ISSUE #1167: JDK 21
The Jakarta EE Platform 11 has a lower bound of JDK 21, so we need to raise bars a bit.

mkarg:
Fixes #1167

This PR enables compilation of JDK 21 in the release-4.0 branch as proposed in the discussion of https://github.com/jakartaee/rest/issues/1167.

### DIFF
--- a/.github/workflows/maven-package.yml
+++ b/.github/workflows/maven-package.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '11', '17' ]
+        java: [ '21-ea' ]
 
     steps:
     - uses: actions/checkout@v2
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,7 @@ pipeline {
 		pollSCM('H H * * *')
 	}
 	tools {
-		jdk 'openjdk-jdk11-latest'
+		jdk 'openjdk-jdk21-latest'
 		maven 'apache-maven-latest'
 	}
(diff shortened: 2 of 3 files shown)


arjantijms (approved):

arjantijms:
Instead of source/target, the release attribute can be used as well.

chkal (approved):

andymc12 (approved):

.github/workflows/maven-package.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '11', '17' ]
+        java: [ '21' ]
```
jamezp, line 16: This would currently need to be `21-ea` I think.

jamezp:
Do we want to do this before Java 21 is released?

mkarg:
> Do we want to do this before Java 21 is released?

This happens *solely* in the release-4.0 branch, where compiling against 21 is mandatory anyways according to the Jakarta EE 11 release plan. JDK 21 published RC 1 already; according to the JDK rules, the sole difference between now and later is more bug fixes. So why should we wait? What would break that could not get fixed within a single business day (and possibly doesn't care as this is not the master branch)?

.github/workflows/maven-package.yml
mkarg: Good catch. Thanks! Fixed. :-)

PR #1172, opened 2023-08-21
1162 sse broadcast test

jim-krueger:
Fixes #1162

PR #1182, opened 2023-09-12
fast-track: 1162 sse broadcast test

jim-krueger:
When moving fix from test system for #1172 I missed an import and for some reason the maven build done as part of the checks did not flag this.  This PR adds the missing import.


spericas (approved):
Merge it!

PR #1189, opened 2023-10-19
[1183] - rename Entity annotation to Body

### LINKED ISSUE #1183: Rename the new @Entity annotation
Hi all,

I am a developer for Open Liberty and we are starting work on our new feature for RESTfulWS-4.0.

During our EE11 design presentation, we received strong feedback about the name of the new `@Entity` annotation. 

The main concern is that both EJB and JPA already have an `@Entity` annotation and we want to avoid another annotation with the same name. Multiple annotations with the same name causes confusion and can also cause problems when IDEs auto import the incorrect class.

Would it make sense to rename `@Entity` to `@RequestEntity`? `@Payload`? `@PayloadEntity`? Something similar?

WhiteCat22:
Fixes #1183

Rename the `@Entity` annotation to `@Body` and update references.

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/asynchronous_processing/_server_api.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/asynchronous_processing/_server_api.adoc
@@ -1,6 +1,6 @@
 ////
 *******************************************************************
-* Copyright (c) 2019 Eclipse Foundation
+* Copyright (c) 2019, 2023 Eclipse Foundation
 *
 * This specification document is made available under the terms
 * of the Eclipse Foundation Specification License v1.0, which is
@@ -30,7 +30,7 @@ Let us illustrate these concepts via an example:
 public class MyResource {
 
     @GET
-    public void longRunningOp(@Suspended final AsyncResponse ar) {
+    public void longRunningOp(final AsyncResponse ar) {
         executor.submit(
             new Runnable() {
                 public void run() {
@@ -44,14 +44,14 @@ public class MyResource {
 ----
 
 A resource method that elects to produce a response asynchronously must
-inject as a method parameter an instance of the class `AsyncResponse`
-using the special annotation `@Suspended`. In the example above, the
-method `longRunningOp` is called upon receiving a `GET` request. Rather
-than producing a response immediately, this method forks a (non-request)
-thread to execute a long running operation and returns immediately. Once
-the execution of the long running operation is complete, the connection
-is resumed and the response returned by calling `resume` on the injected
-instance of `AsyncResponse`.
+have an instance of the class `AsyncResponse` as a method parameter.
+
+In the example above, the method `longRunningOp` is called upon receiving
+a `GET` request. Rather than producing a response immediately, this method
+forks a (non-request) thread to execute a long running operation and returns
+immediately. Once the execution of the long running operation is complete,
+the connection is resumed and the response returned by calling `resume` on
+the injected instance of `AsyncResponse`.
 
 For more information on executors, concurrency and thread management in
 a Jakarta EE environment, the reader is referred to Jakarta Concurrency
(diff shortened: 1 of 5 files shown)


jaxrs-spec/src/main/asciidoc/chapters/asynchronous_processing/_server_api.adoc
```diff
@@ -29,12 +29,13 @@ Let us illustrate these concepts via an example:
 @Path("/async/longRunning")
 public class MyResource {
 
-    @GET
-    public void longRunningOp(@Suspended final AsyncResponse ar) {
+    @POST
+    public void longRunningOp(final AsyncResponse ar, @Body String message) {
```
spericas, line 33: Dropping `@Suspended` is OK, but why the addition of `@Body` in this example?

WhiteCat22: I changed it to a POST example to show the new `@Body` requirement that I'm documenting below. Is this incorrect? I basically took it from the slideshow.

spericas: I don't think it is wrong, but seems unnecessary. Note that the same example is repeated later in that chapter, so perhaps it is better to leave it as it was.

WhiteCat22: I think it would be a good idea to have at least one example showing the new `@Body` annotation being used.

That being said, I did miss that they example is repeated later. That example should at least have the `@Suspended` annotation removed regardless.

WhiteCat22: I did remove `@Suspended` from the 2nd example, but I would like some more opinions on changing the example to include `@Body`. I'd be happy to make the 2nd example match the first if that is the consensus.

jim-krueger: >I think it would be a good idea to have at least one example showing the new @Body annotation being used.
I'd tend to agree since this is where `@Body` is introduced.  Nice to see it in practice.

jim-krueger (approved):

jaxrs-spec/src/main/asciidoc/chapters/asynchronous_processing/_server_api.adoc
spericas: There must be a better example in some other section where we can show the use of `@Body`. This one is really about explaining async.

jaxrs-spec/src/main/asciidoc/chapters/resources/_resource_method.adoc
```diff
@@ -55,7 +55,8 @@ parameters. See <<consuming_multipart_formdata>> for more details.
 
 The value of a parameter not annotated with `@FormParam` or any of the
 annotations listed in in <<resource_field>>, called the entity
-parameter, is mapped from the request entity body. Conversion between an
+parameter, is mapped from the request entity body. This parameter MUST
+be annotated with the `@Body` annotation. Conversion between an
```
WhiteCat22, line 59: @spericas I moved the `@Body` requirement here.

jaxrs-spec/src/main/asciidoc/chapters/asynchronous_processing/_server_api.adoc
```diff
@@ -44,14 +44,14 @@ public class MyResource {
...
-is resumed and the response returned by calling `resume` on the injected
-instance of `AsyncResponse`.
+have an instance of the class `AsyncResponse` as a method parameter.
+
+In the example above, the method `longRunningOp` is called upon receiving
+a `GET` request. Rather than producing a response immediately, this method
+forks a (non-request) thread to execute a long running operation and returns
+immediately. Once the execution of the long running operation is complete,
+the connection is resumed and the response returned by calling `resume` on
+the injected instance of `AsyncResponse`.
```
WhiteCat22, line 54: I am leaving these changed (minus the `@Body` section) since I think I've made it easier to read. I can remove this if you want.

spericas (approved):

PR #1192, opened 2023-11-15
M1 spec changes1

jim-krueger:
First pass at changes needed to Jakarta Rest-4.0 specification.

### DIFF
--- a/jaxrs-spec/src/main/asciidoc/chapters/context/_contexttypes.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/context/_contexttypes.adoc
@@ -1,6 +1,6 @@
 ////
 *******************************************************************
-* Copyright (c) 2019 Eclipse Foundation
+* Copyright (c) 2019, 2023 Eclipse Foundation
 *
 * This specification document is made available under the terms
 * of the Eclipse Foundation Specification License v1.0, which is
@@ -21,26 +21,26 @@ other types are server-side only.
 ==== Application
 
 The instance of the application-supplied `Application` subclass can be
-injected into a class field or method parameter using the
-`@Context` annotation. Access to the `Application` subclass instance
-allows configuration information to be centralized in that class. Note
-that this cannot be injected into the `Application` subclass itself
-since this would create a circular dependency.
+injected a method parameter, since it is a CDI bean, or injected into
+a class field using the`@Inject` annotation. Access to the
+`Application` subclass instance allows configuration information to be
+centralized in that class. Note that this cannot be injected into the
+`Application` subclass itself since this would create a circular dependency.
 
 [[uris-and-uri-templates]]
 ==== URIs and URI Templates
 
-An instance of `UriInfo` can be injected into a class field or method
-parameter using the `@Context` annotation. `UriInfo` provides both
-static and dynamic, per-request information, about the components of a
-request URI. E.g. the following would return the names of any query
-parameters in a request:
+An instance of `UriInfo` can be injected into a method parameter, since it
+is a CDI bean, or injected into a class field using the `@Inject` annotation.
+`UriInfo` provides both static and dynamic, per-request information, about
+the components of a request URI. E.g. the following would return the names of
+any query parameters in a request:
 
 [source,java]
 ----
 @GET
 @Produces("text/plain")
-public String listQueryParamNames(@Context UriInfo info) {
+public String listQueryParamNames(UriInfo info) {
     StringBuilder buf = new StringBuilder();
     for (String param: info.getQueryParameters().keySet()) {
         buf.append(param);
@@ -56,17 +56,17 @@ information following the pre-processing described in <<reqpreproc>>.
 [[httpheaders]]
 ==== Headers
 
-An instance of `HttpHeaders` can be injected into a class field or
-method parameter using the `@Context` annotation. `HttpHeaders` provides
-access to request header information either in map form or via strongly
-typed convenience methods. E.g. the following would return the names of
-all the headers in a request:
+An instance of `HttpHeaders` can be injected into a method parameter,
+since it is a CDI bean, or injected into a class field using the
+`@Inject` annotation. `HttpHeaders` provides access to request header
+information either in map form or via strongly typed convenience methods.
+E.g. the following would return the names of all the headers in a request:
 
 [source,java]
 ----
 @GET
 @Produces("text/plain")
-public String listHeaderNames(@Context HttpHeaders headers) {
+public String listHeaderNames(HttpHeaders headers) {
     StringBuilder buf = new StringBuilder();
     for (String header: headers.getRequestHeaders().keySet()) {
         buf.append(header);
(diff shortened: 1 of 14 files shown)


jaxrs-spec/src/main/asciidoc/chapters/context/_contexttypes.adoc
```diff
@@ -21,26 +21,26 @@ other types are server-side only.
 ==== Application
 
 The instance of the application-supplied `Application` subclass can be
-injected into a class field or method parameter using the
-`@Context` annotation. Access to the `Application` subclass instance
-allows configuration information to be centralized in that class. Note
-that this cannot be injected into the `Application` subclass itself
-since this would create a circular dependency.
+injected a method parameter, since it is a CDI bean, or injected into
```
spericas, line 24: "injected as a method parameter"?

spericas: I think we should say "resource method parameter" not just "method parameter" given that an arbitrary method that isn't a resource method, won't support that. I guess this comment applies to other cases below.

jim-krueger: I agree and have applied the recommended changes.

spericas: One additional comment, the switch to CDI also allows for injection into constructors, which didn't work all the time before, and that's why it was not mentioned explicitly in this section.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
```diff
@@ -14,11 +14,10 @@
...
 <<provider_class_constructor>>) is called, then any requested dependencies
-are injected (see Chapter <<context>>), then the appropriate provider
-methods may be called multiple times (simultaneously), and finally the
-object is made available for garbage collection.
-<<providercontext>> describes how a provider obtains access to other
-providers via dependency injection.
+are injected, then the appropriate provider methods may be called multiple
+times (simultaneously), and finally the object is made available for
+garbage collection. <<providercontext>> describes how a provider obtains
+access to other providers via dependency injection.
```
spericas, line 20: Perhaps all this paragraph could be replaced by simply stating that a Jakarta REST provider is in application scope (as defined by CDI) and instantiated and initialized accordingly. WDYT?

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
```diff
@@ -63,11 +62,10 @@ have a public constructor for which the JAX-RS runtime can provide all
...
 
-A public constructor MAY include parameters annotated with
-`@Context`—Chapter <<context>> defines the parameter types permitted for
-this annotation. Since providers may be created outside the scope of a
-particular request, only deployment-specific properties may be available
-from injected interfaces at construction time; request-specific
+A public constructor MAY include injected parameters, provided the
+parameter type is a CDI bean. Since providers may be created outside the
+scope of a particular request, only deployment-specific properties may
+be available from injected interfaces at construction time; request-specific
```
spericas, line 68: This seems fine, but it's all governed by CDI now, so perhaps we don't need to be so explicit anymore.

jaxrs-spec/src/main/asciidoc/chapters/resources/_resource-classes.adoc
```diff
@@ -40,10 +40,10 @@ parameter values. Note that a zero argument constructor is permissible
...
 A public constructor MAY include parameters annotated with one of the
-following: `@Context`, `@HeaderParam`, `@CookieParam`, `@MatrixParam`,
-`@QueryParam` or `@PathParam`. However, depending on the resource class
-lifecycle and concurrency, per-request information may not make sense in
-a constructor. If more than one public constructor is suitable then an
+following: `@HeaderParam`, `@CookieParam`, `@MatrixParam`, `@QueryParam` or
+`@PathParam`. However, depending on the resource class lifecycle and
+concurrency, per-request information may not make sense in a
+constructor. If more than one public constructor is suitable then an
 implementation MUST use the one with the most parameters. Choosing
```
spericas, line 47: I don't know if we can inject these in constructors anymore. We can in resource methods since we should be in control of the invocation. Perhaps we could do this in constructors too, but would need some CDI expert to help us here.

jim-krueger: I've asked Emily Jiang and a few others to give an opinion.

jansupol: It seems to be similar to `@Valid` for instance. It could be easier for the customers to transition to 4.0 allowing it, but I am not sure what CDI shenanigans it would mean to support these annotations right now.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
jansupol: By default in application scope, for the case someone would want to use some other scope, or singleton for user-provided instances (`Application.getSingletons()`)

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
jansupol: I am pretty sure I used this comment about the unavailable injection of objects outside of the request scope when explaining this to a customer. But it would be likely mentioned in the CDI Spec as well.

jansupol (approved):
I am fine with the changes for M1. While there might be doubts, they can be addressed later. This PR improves the document, no clear no-no to me.

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
jim-krueger: @jansupol Side note:  Application.getSingletons() was deprecated in 3.1.  Should we consider removing it in 4.0?

jim-krueger:
[spericas](https://github.com/spericas), [arjantijms](https://github.com/arjantijms) if you concur with [jansupol](https://github.com/jansupol) could you please approve this PR to allow and M1 version to be created?   Thanks

jaxrs-spec/src/main/asciidoc/chapters/resources/_resource-classes.adoc
jamezp: I'll need to go back and read the CDI issues, but my assumption or thought, was that these are CDI qualifiers?

arjantijms (approved):

jaxrs-spec/src/main/asciidoc/chapters/providers/_lifecycle_and_environment.adoc
spericas: @jim-krueger I think we should remove it. There's no need to confuse developers with the notion of a "singleton" when we now can take advantage of CDI's application scope.

jim-krueger:
> Don't we have until the end of January for M1? IMO we should work on this and the Javadocs more before talking about M1.

Perhaps I am under an misunderstanding, but I thought from the Jakarta EE Platform call that the deadline for M1 was December 5th (https://docs.google.com/document/d/1DF0T2hbJqDNsY0v_lYFKCz-ZgHPbnMUIu6mhmt5LCJU/edit#heading=h.8is3kgp5ds15) and I think all other Jakarta EE specs are done.

spericas:
> > Don't we have until the end of January for M1? IMO we should work on this and the Javadocs more before talking about M1.
> 
> Perhaps I am under a misunderstanding, but I thought from the Jakarta EE Platform call that the deadline for M1 was December 5th (https://docs.google.com/document/d/1DF0T2hbJqDNsY0v_lYFKCz-ZgHPbnMUIu6mhmt5LCJU/edit#heading=h.8is3kgp5ds15) and I think all other Jakarta EE specs are done.
> 
> If we truly do have until the end of January I agree, plus there are a few other changes that would be nice to include by then as well. I was assuming those would need to go into M2.

I don't get these dates, but it seems that you are right. I'll merge this one then to get M1 "ready".

spericas (approved):
Approving for M1

jim-krueger:
Thanks @spericas.  I try to attend the weekly Platform calls and will update our [4.0 Epic](https://github.com/jakartaee/rest/issues/1158) with pertinent information going forward.

Also, FWIW: I opened an issue in gitlab.eclipse.org and they added missing ssh public key to eclipse-rest-bot github account, which allowed me to successfully run a dry-run for the M1 apis.   I think there needs to be a few more changes to create a true M1 (not a SNAPSHOT).

jamezp:
I'm not sure this is the correct place for this comment, but I wasn't sure where else to put it. In the pom the [`jaxrs-tck`](https://github.com/jakartaee/rest/blob/3a41a4dd349002b5c4703a4a5b891418266df3d8/pom.xml#L378) module is commented out. Presumably this is because of the changes needed for new TCK tests and/or migration from `@Context` injection. Is there a plan to address that as well?

jamezp:
> @jamezp M1 is only spec and api. tck will come later.

Thank you @jim-krueger. I had a feeling, but wasn't really too sure what the plan was :)

jim-krueger:
@jamezp No problem.  I believe the schedule for the TCK and a RI is early in 2024 (which might be the cause of the M1 confusion that @spericas mentioned).

PR #1203, opened 2023-12-13
Add the enforcer plugin and require a minimum of Java 21 to compile t…

### LINKED ISSUE #1202: Compile the API to Java SE 17
I had started a discussion on the [mailing list](https://www.eclipse.org/lists/rest-dev/msg00114.html) about this.

In https://github.com/jakartaee/rest/pull/1168 the project was updated to require Java SE 21. We don't however use any Java SE 21 specific features. In my opinion it would be better for adoption to have a minimum of Java SE 17. 

There are consumers outside of Jakarta EE that consume the Jakarta REST spec. Some of these consumers are in environments where they are likely not able to use the latest version of Java for various reasons.

I propose we compile down to Java 17, but keep the minimum JDK requirement for compiling at Java 21. No other individual specification, other than Jakarta Concurrency, are requiring Java 21. There is nothing in the specification that requires a project to use any specific version of Java at all.

jamezp:
The minimum level should remain at Java SE 17. The TCK itself should also require Java SE 17 as a minimum.

resolves #1202

### DIFF
--- a/.github/workflows/maven-package.yml
+++ b/.github/workflows/maven-package.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '21-ea' ]
+        java: [ '17', '21' ]
 
     steps:
     - uses: actions/checkout@v2
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -35,7 +35,8 @@
     </parent>
 
     <properties>
-        <maven.compiler.release>21</maven.compiler.release>
+        <!-- Explicitly set the release level to Java SE 17 -->
+        <maven.compiler.release>17</maven.compiler.release>
         <junit.jupiter.version>[5.7.2, 5.8-A00)</junit.jupiter.version>
         <tck.artifactId>jakarta-restful-ws-tck</tck.artifactId>
     </properties>
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -35,8 +35,7 @@
     </parent>
 
     <properties>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.release>21</maven.compiler.release>
         <junit.jupiter.version>[5.7.2, 5.8-A00)</junit.jupiter.version>
         <tck.artifactId>jakarta-restful-ws-tck</tck.artifactId>
     </properties>
@@ -125,7 +124,6 @@
                 <plugins>
                     <plugin>
                         <artifactId>maven-compiler-plugin</artifactId>
-                        <version>2.3.2</version>
                     </plugin>
                     <plugin>
                         <artifactId>maven-surefire-plugin</artifactId>
@@ -133,7 +131,6 @@
                     </plugin>
                     <plugin>
                         <artifactId>maven-jar-plugin</artifactId>
-                        <version>3.2.0</version>
                     </plugin>
                     <plugin>
                         <artifactId>maven-source-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,10 @@
     </scm>
 
     <properties>
-        <java.version>21</java.version>
+        <!-- Used in the enforcer plugin to require the minimum Java version -->
+        <jdk.min.version>21</jdk.min.version>
+        <!-- used in the compiler plugin for the release level -->
+        <maven.compiler.release>17</maven.compiler.release>
 
         <apidocs.title>Jakarta RESTful Web Services ${spec.version} API Specification ${spec.version.revision}</apidocs.title>
         <legal.doc.folder>${project.basedir}</legal.doc.folder>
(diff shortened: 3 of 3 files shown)


jaxrs-tck/pom.xml
```diff
@@ -35,8 +35,8 @@
     </parent>
 
     <properties>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.source>21</maven.compiler.source>
```
NicoNes, line 38: Is the `maven.compiler.source` still needed if you set the `maven.compiler.release` ?

-- Nicolas

pom.xml
```diff
@@ -175,8 +179,8 @@
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>${maven.compiler.plugin.version}</version>
                     <configuration>
-                        <source>${java.version}</source>
-                        <target>${java.version}</target>
+                        <source>${maven.compiler.source}</source>
```
NicoNes, line 182: I think that this configuration section is not necessary because the maven compiler plugin autoconfigure itself reading properties `maven.compiler.source` and  `maven.compiler.release`.

-- Nicolas

pom.xml
```diff
@@ -401,5 +427,50 @@
                 <module>jersey-tck</module>
             </modules>
         </profile>
+
+        <!-- This profile is activated when Java 21 or later is used to build and the project has a src/main/java21
+             directory. -->
+        <profile>
```
mkarg, line 433: I am -1 for making this dependend of the used JDK. I agreed solely to *adding* 17 into the MR, not to making 21 *optional*-

mkarg (changes requested):

pom.xml
jamezp: No worries, I can remove that. This was just setup in case there was any need for an MR JAR. It's easily removed and serves no purpose yet anyway.

jaxrs-tck/pom.xml
jamezp: You're correct. Some old habits die hard, I'll remove that :)

pom.xml
```diff
@@ -175,8 +178,7 @@
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>${maven.compiler.plugin.version}</version>
                     <configuration>
-                        <source>${java.version}</source>
-                        <target>${java.version}</target>
+                        <release>${maven.compiler.release}</release>
```
NicoNes, line 181: @jamezp I think you forgot to remove this "release" element since the maven compiler plugin autoconfigure itself with properties `maven.compiler.release`

-- Nicolas

NicoNes (changes requested):

pom.xml
NicoNes: Ok. So to be consistent should'nt we be that explicit here too : https://github.com/jakartaee/rest/pull/1203/files#diff-0b9d1df7da9cc5c30f2c2599dc9a182117210c78c941d576dfaca5106aa50be1R126 ?

-- Nicolas

jamezp: I only didn't there because it will be inherited from here. Again though, I have no strong opinion and will be happy to remove it from here. :)

mkarg: A best practice of Maven is CoC, so I would ask you to not repeat implicit defaults without a good reason. Thanks.

jamezp:
Please note I've changed this PR to have the TCK compile to Java SE 17 as well. It was noted in the [Jakarta EE spec meeting](https://docs.google.com/document/d/1DF0T2hbJqDNsY0v_lYFKCz-ZgHPbnMUIu6mhmt5LCJU/edit#heading=h.ckdwbxovtrpb) there was never an intent for specification TCK's to require Java 21.

jamezp:
> So we can just set `release` to 17 in the compiler plugin and drop the enforcer plugin?

Yes, we could do that as well. The only reason for the enforcer plugin is to know early if you're on the right JDK. It's in no way critical.

jansupol (approved):
Early JDK version check makes sense

asoldano (approved):

.github/workflows/maven-package.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ '21-ea' ]
+        java: [ '17', '21' ]
```
mkarg, line 16: This is not what we discussed. It simply runs two parallel builds, one using JDK 17 and one using JDK 21. For a MR, we just need a single build on JDK 21 but with two compilation targets, 17 and 21.

jamezp: We currently have no reason for a MR JAR. We can consider adding it back, but without the need for a `META-INF/versions/21` it makes no sense.

mkarg: The MR was the last that you and me agreed upon, but indeed, now that Arjan vetoed against the MR, my comment is obsolete.

jansupol:
[Here](https://github.com/azure-javaee/jakartaee-platform/blob/7f264e0f6260aca64f81ba994d072362f3ef33be/jakartaee11/JakartaEE11ReleasePlan.md#api-source-code---release-level) it says: 
>For the Jakarta EE Platform (Platform, Web and Core), the Java compiler --release option is 17. For the component specs, the Jakarta EE Platform requires the Java compiler --release option is at most 17, but component specifications can decide on a lower level.

jamezp:
Given the proposal of 3.2, I'm going to close this. We can re-address it if plans change.

PR #1210, opened 2024-01-24
Deprecate both @Context and @Suspended

jim-krueger:
Deprecate `@Context` and `@Suspended` in release-3.2 in preparation for removal in release-4.0

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/container/Suspended.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/container/Suspended.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -76,7 +76,9 @@ import java.util.concurrent.TimeUnit;
  *
  * @author Marek Potociar
  * @since 2.0
+ * @deprecated This class will be removed in a future version.  Better align with Jakarta CDI.
  */
+@Deprecated(forRemoval = true)
 @Target({ ElementType.PARAMETER })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/Context.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/Context.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -38,7 +38,9 @@ import java.lang.annotation.Target;
  * @see SecurityContext
  * @see jakarta.ws.rs.ext.Providers
  * @since 1.0
+ * @deprecated This class will be removed in a future version.  Better align with Jakarta CDI.
  */
+@Deprecated(forRemoval = true)
 @Target({ ElementType.PARAMETER, ElementType.METHOD, ElementType.FIELD })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
--- a/.github/workflows/maven-package.yml
+++ b/.github/workflows/maven-package.yml
@@ -3,9 +3,9 @@ name: Jakarta REST CI
 
 on:
   push:
-    branches: [ '3.2.0-SNAPSHOT' ]
+    branches: [ 'release-3.2' ]
   pull_request:
-    branches: [ '3.2.0-SNAPSHOT' ]
+    branches: [ 'release-3.2' ]
 
 jobs:
   build:


jaxrs-api/src/main/java/jakarta/ws/rs/container/Suspended.java
```diff
@@ -76,7 +76,9 @@
  *
  * @author Marek Potociar
  * @since 2.0
+ * @deprecated This class will be removed in a future version.  Better align with Jakarta CDI.
  */
+@Deprecated
```
jamezp, line 81: Should this be `@Deprecated(forRemoval = true)`?

jaxrs-api/src/main/java/jakarta/ws/rs/core/Context.java
```diff
@@ -38,7 +38,9 @@
  * @see SecurityContext
  * @see jakarta.ws.rs.ext.Providers
  * @since 1.0
+ * @deprecated This class will be removed in a future version.  Better align with Jakarta CDI.
  */
+@Deprecated
```
jamezp, line 43: Same here too, should this be `@Deprecated(forRemoval = true)`?

jaxrs-api/src/main/java/jakarta/ws/rs/container/Suspended.java
jim-krueger: very true... thanks

jamezp (approved):

jamezp (approved):

mkarg:
This PR is a violation of the committer rules of this project if we take them word-by-word: https://github.com/jakartaee/rest/wiki/Committer-Conventions. There were no three +1 votes, there was no two-weeks voting time. **We need to discuss if we set up lower barriers for 3.2** (which I am +1 actually; one way could be to extend the rules by the notion of "working branches", hence 
 `release-3.2` would be merged into a "releasing-branch" (like `3.2`) later using a separate PR). **Until then I would kindly ask you to respect the rules.** Thanks.

jim-krueger:
My apologies Markus.  I was thinking of release-3.2 as a "working branch" since we do not have a release plan etc. for it yet and Santiago had requested that we determine if certain issues could be worked through concerning deprecation prior to taking that step.  I should have stated that clearly.  That being said it sounds like you are not adverse to proceeding this way until such a time as 3.2 is "official", at which time the content of PRs already applied could be confirmed via a vote, with the content removed if found to not be acceptable.

mkarg:
I am not standing in the way of dealing with `release-3.2` as a "working branch" in that sense, and I think nobody else would consider to reject that idea. Nevertheless it would be best to have an official rule. Hence I would encourage you to start a voting on the mailing list to get this new exemption added into our official ruleset.

paulrutter:
Just came across this PR while moving to 3.1.
What is the designated replacement for `@Suspended`?

I know that `@Context` is replaced with `@Inject`, to align with CDI. Thanks. 

@jim-krueger

jamezp:
@paulrutter Effectively it will be the same, `@Inject`, but AIUI implied on the method. In other words you simply just remove the `@Suspended` annotation from your parameter and the implementation will resolve the value.

As an example:
```java
@POST
@Path("async")
@Consumes(MediaType.TEXT_PLAIN)
public void putMessageAsync(
    @QueryParam("override") boolean override,
    String message,
    @Suspended AsyncResponse ar) {
    managedExecutor.submit() -> {
        ar.resume("done");
    });
}
```
Would become:
```java
@POST
@Path("async")
@Consumes(MediaType.TEXT_PLAIN)
public void putMessageAsync(
    @QueryParam("override") boolean override,
    @Body String message,
    AsyncResponse ar) {
    managedExecutor.submit() -> {
        ar.resume("done");
    });
}
```

paulrutter:
Thanks for clarifying, that makes perfect sense.

paulrutter:
@jim-krueger one more question about `@Context` being deprecated, how would the following example look like after it has been removed?

```
    @Path("/resetpassword")
    @POST
    @Consumes(APPLICATION_FORM_URLENCODED)
    @Produces(APPLICATION_JSON)
    public Response requestResetPassword(@FormParam(USERNAME) final String username,
            @Context final HttpServletRequest request) {
        // do something with the HttpServletRequest
    }
```

Is it still possible to get a handle on the `HttpServletRequest` after `@Context` has been sunset?

jim-krueger:
@paulrutter This is still a work in progress, but I believe the goal is for it to look like this:

```
@Path("/resetpassword")
    @POST
    @Consumes(APPLICATION_FORM_URLENCODED)
    @Produces(APPLICATION_JSON)
    public Response requestResetPassword(@FormParam(USERNAME) final String username,
            final HttpServletRequest request) {
        // do something with the HttpServletRequest
    }
```
You'll want to keep an eye on [Issue 1214](https://github.com/jakartaee/rest/issues/1214) as that is where this problem is being tracked.

paulrutter:
Thanks, will do.

hantsy:
I think a simple guide or notice in Javadoc and a section in Speciction should be available to describe how to migrate to use the replacement of `@Context` and `@Suspend`.

mkarg:
> I think a simple guide or notice in Javadoc and a section in Speciction should be available to describe how to migrate to use the replacement of `@Context` and `@Suspend`.

Maybe you like to file a PR against branch `Release-5.0`? 🙂

hantsy:
>  @POST
> @Path("async")
> @Consumes(MediaType.TEXT_PLAIN)
> public void putMessageAsync(
>     @QueryParam("override") boolean override,
>     @Body String message,
>     AsyncResponse ar) {
>     managedExecutor.submit() -> {
>         ar.resume("done");
>     });
> }

Why not remove the AsyncResponse ?

Make the following working.

```java
@POST
@Path("async")
@Consumes(MediaType.TEXT_PLAIN)
public CompletionStage/CompletableFuture/Future putMessageAsync(
    @QueryParam("override") boolean override,
    @Body String message) {
    managedExecutor.submit() -> {
    });
}
```

or return `Response<CompletionStage>` or `CompletionStage<Reponse>`.

PR #1211, opened 2024-01-25
Cleanup of pom files to use dependency management and dependency plug…

jamezp:
…ins to avoid repetition.

This is effectively a port of https://github.com/jakartaee/rest/commit/e5f23e8cb328aee010f9ea1dae85ec8186d7cf93.

One thing I did not do was remove the dependency on `jakarta.xml.bind.api`. I wasn't too clear on why that was needed.

This also sets the minimum JDK required and compiles to Java SE 17.


jim-krueger (approved):
Approving but adding Santiago to review since he added the original commit referenced and James has added additional changes.

jamezp:
Yes, I could use some clarification on why the `jakarta.xml.binding.api` was removed too. I also noticed I forgot to update the `module-info` for the API to include the requirement for the two CDI modules. We can do that now, or do it when we actually update the annotations to use CDI annotations.

jamezp:
It seems like currently we need to keep that static dependency on JAXB. Just for reference this is discussed on https://github.com/jakartaee/rest/issues/906 and https://github.com/jakartaee/rest/issues/907. The latter makes it effectively optional. However, we still have a compile time dependency on it.

I did remove the references in the examples, which seems okay. I will wait to add the CDI dependencies on the module until we add the CDI annotations to the REST annotations.

jaxrs-tck/pom.xml
```diff
@@ -236,5 +103,71 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
mkarg, line 124: It is a best practice to not add that particular group ID as it is the default group ID of Maven.

pom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2021 Markus KARG. All rights reserved.
+    Copyright (c) 2024 Oracle and/or its affiliates. All rights reserved.
```
mkarg, line 4: I veto the replacement of my copyright, just as Oracle would veto theirs in such a case.

mkarg (changes requested):
Overall LGTM, but I veto the replacement of my Copyright.

jaxrs-tck/pom.xml
jamezp: I guess that's debatable. The default value in the XSD is for `groupId` is `org.apache.maven.plugins`. I honestly don't mind either way, so if we prefer to have it I can make sure to add it everywhere. If we prefer to not have it defined, I can make sure they are not used anywhere.

pom.xml
jamezp: I just pulled that from https://github.com/jakartaee/rest/commit/e5f23e8cb328aee010f9ea1dae85ec8186d7cf93#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L4 which is in release-4.0 as well. That said, I'm happy to change it back. I don't really know what it should be so keeping it as is, e.g. "Copyright (c) 2021 Markus KARG. All rights reserved." seems fine to me.

mkarg: Thank you for keeping it. Once fixed I am +1 with the rest of this PR.

jaxrs-tck/pom.xml
mkarg: My personal vote would be to go with the Maven best practice of CoC and *neven* copy a default value explicitly.

mkarg:
IIRC the target for 3.x originally was Java 11. This PR will set it to Java 17 (which I do like very much). But is that allowed in SemVer? I mean, downstreams that are not able to run in Java 17 yet would not be able to upgrade the 3.x line then.

jaxrs-tck/pom.xml
jamezp: I went ahead and removed all the `<groupId>org.apache.maven.plugins</groupId>`.

jamezp:
> IIRC the target for 3.x originally was Java 11. This PR will set it to Java 17 (which I do like very much). But is that allowed in SemVer? I mean, downstreams that are not able to run in Java 17 yet would not be able to upgrade the 3.x line then.

That is a good question and I don't know the answer. I do know other Jakarta EE 10 specifications are moving to require Java SE 17. For now I've left this at 17, but if it's determined we cannot do this upgrade I will update the PR, or send a separate PR, to downgrade back to Java SE 11 as a mnimum.

jamezp:
> > That is a good question and I don't know the answer. I do know other Jakarta EE 10 specifications are moving to require Java SE 17. For now I've left this at 17, but if it's determined we cannot do this upgrade I will update the PR, or send a separate PR, to downgrade back to Java SE 11 as a minimum.
> 
> I believe you meant ...other Jakarta EE 11 specifications.... and I posed this question in the [JakartaEE Development slack channel ](https://eclipsefoundationhq.slack.com/archives/CK4UW9DPS/p1706716440515709) and the replies seem to indicate that this change will not be a problem in a point release.

Yes, I meant Jakarta EE 11, sorry about that typo :)

jim-krueger:
@mkarg @spericas Can we get a couple more review approvals on this to get it in?    Thanks

jamezp:
Note that I changed the requirement of Java SE 17 back to Java SE 11 as we may not be able to change this in a minor version. If we find out we can, we can change it in a separate PR.

mkarg:
> Note that I changed the requirement of Java SE 17 back to Java SE 11 as we may not be able to change this in a minor version. If we find out we can, we can change it in a separate PR.

What is the benefit of 3.2 if it does only support Java 11? IIRC the sole idea of 3.2 was to support JDK 17.

jamezp:
> > Note that I changed the requirement of Java SE 17 back to Java SE 11 as we may not be able to change this in a minor version. If we find out we can, we can change it in a separate PR.
> 
> What is the benefit of 3.2 if it does only support Java 11? IIRC the sole idea of 3.2 was to support JDK 17.

I personally don't see a benefit, but I thought you had concerns about it being allowed https://github.com/jakartaee/rest/pull/1211#issuecomment-1913088963. Maybe I misunderstood, but the idea was to not block this PR as we can upgrade the minimum JDK version later. We can't really progress with other changes until this is in without creating a bunch of conflicts.

spericas (approved):

jim-krueger (approved):

jamezp:
Moved back to Java SE 17 :)

jim-krueger:
@mkarg Can you please approve or express your concerns with this PR as it is holding up progress on 3.2.   Thanks

jaxrs-tck/pom.xml
```diff
@@ -195,38 +66,34 @@
...
             <groupId>jakarta.activation</groupId>
             <artifactId>jakarta.activation-api</artifactId>
-            <version>2.1.0-RC1</version>
         </dependency>
 
         <dependency>
             <groupId>jakarta.servlet</groupId>
             <artifactId>jakarta.servlet-api</artifactId>
-            <version>5.0.0</version>
+            <version>6.0.0</version>
```
jansupol, line 96: This should probably be 6.1.0-M1

jamezp: I was going to follow up with PR's to upgrade components to the milestones. If we want to address it in this PR we can though.

pom.xml
```diff
@@ -34,7 +34,472 @@
...
+        <legal.doc.folder>${project.basedir}</legal.doc.folder>
+
+        <maven.bundle.plugin.version>3.5.0</maven.bundle.plugin.version>
+        <maven.compiler.plugin.version>3.8.1</maven.compiler.plugin.version>
+        <maven.javadoc.plugin.version>3.1.0</maven.javadoc.plugin.version>
+
+        <api.package>jakarta.ws.rs</api.package>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <skip.release.tests>false</skip.release.tests>
+        <spec.version>4.0</spec.version>
```
jansupol, line 97: spec.version=3.2

jaxrs-tck/pom.xml
jansupol: leaving up to you :)

jim-krueger (approved):

jaxrs-tck/pom.xml
jamezp: IMO we wait for right now. This kind of blocks any other work as it's such a big change.

jansupol (approved):

jaxrs-tck/pom.xml
jim-krueger: This conversation appears to be resolved so marking it as such.

jim-krueger:
Merging as the number of approvers has been met and all review comments have been resolved.

PR #1228, opened 2024-03-01
fast track: Port missed fixes that were in release-4.0

jamezp:
I found these commits I missed that were in release-4.0 and not in main or release-3.2. 

---

###  GItHub Links
The first commit is to fix a bad port which reverted some changes from #1177 

### Asciidoctor Dependency Updates
These are from #1170 

### Maven Plugin Dependency Updates
This includes a JUnit and mockito upgrades as well. This is a port of #1169

### DIFF
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -119,14 +119,6 @@
             </plugin>
             <plugin>
                 <artifactId>maven-source-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>attach-sources</id>
-                        <goals>
-                            <goal>jar</goal>
-                        </goals>
-                    </execution>
-                </executions>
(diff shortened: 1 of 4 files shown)


jaxrs-tck/pom.xml
```diff
@@ -119,14 +119,6 @@
             </plugin>
             <plugin>
                 <artifactId>maven-source-plugin</artifactId>
-                <executions>
```
jamezp, line 122: This was required due to https://issues.apache.org/jira/browse/MSOURCES-121. The `jar-no-fork` goal is defined as an execution in the root POM.

jim-krueger (approved):
Good that you caught these.

spericas (approved):

arjantijms (approved):

Issue #1226, opened 2024-02-27
Remove JAXB usage in Link class


jim-krueger:
Issue #853 deprecated the Link inner classes that resulted in a JAXB dependency within Jakarta RESTful Web Services in version 3.1.   In Jakarta RESTful Web Services 4.0 these inner classes should be removed to eliminate this dependency.

spericas:
@jim-krueger I can take this one

jim-krueger:
Closing as it is addressed.

PR #1232, opened 2024-03-11
Removes JaxbLink and related classes and tests

spericas:
Removes JaxbLink and related classes and tests. Includes some minor cleanup of versions and version management in pom files too. Issue #1226.


jaxrs-api/src/main/java/jakarta/ws/rs/core/Link.java
```diff
@@ -16,18 +16,11 @@
```
jim-krueger, line 16: Change to 2024

jaxrs-api/src/main/java/module-info.java
```diff
@@ -19,8 +19,6 @@
  */
 module jakarta.ws.rs {
```
jim-krueger, line 20: Change to 2024

jim-krueger (commented):
Just a couple copyright changes, sorry.

jim-krueger (approved):
Looks good!

jamezp (approved):
This LGTM. The only other thing that needs to be done is the signature test file needs to be updated. That said, it looks like the signature test file is not correct as I'm seeing it fail locally for other reasons. We could just update it in a follow up PR.

jaxrs-api/src/main/java/jakarta/ws/rs/core/Link.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2020 Oracle and/or its affiliates. All rights reserved.
```
jim-krueger, line 2: Sorry,  according to https://www.eclipse.org/projects/handbook/#ip-copyright-headers this should really be just
`Copyright (c) 2011`
The same with the rest.
Sorry for the confusion!

jim-krueger (commented):
Sorry for the copyright confusion.   Shame on me for assuming that copyright practices would be universal. https://www.eclipse.org/projects/handbook/#legaldoc-faq

jaxrs-api/src/main/java/jakarta/ws/rs/core/Link.java
spericas: Now is it wrong to have two years (create and modify) or is it simply no longer required? There's many files with two-year copyrights we probably don't want to update at this time.

jim-krueger: Apparently it is wrong to have two years  [](https://www.eclipse.org/projects/handbook/#legaldoc-faq) .  And yes, I see our code is all of the board in this respect.   I certainly wouldn't be in favor of fixing every file and I probably wouldn't have even brought it up if I hadn't requested that you make an incorrect change earlier (sorry).   @mkarg brought this up in a review of one of my PRs that was also incorrect.   And I will try to follow the rules going forward and I'll probably fix any part I touch that is incorrect.

spericas: Ok, I just updated those two files again

jamezp (approved):

jim-krueger (approved):

jaxrs-api/src/main/java/jakarta/ws/rs/core/Link.java
mkarg: > Now is it wrong to have two years (create and modify) or is it simply no longer required? There's many files with two-year copyrights we probably don't want to update at this time.

According to the Eclipse Handbook it is not necessarily wrong or forbidden to have two years, but having the second year is not adviced anymore. Hence to not make it wrong, it is easiest to remove the second date whenever we touch a file, but there is no need to remove the second date unless the file is touched anyways.

jansupol (approved):

PR #1237, opened 2024-03-14
Add stale while revalidate

khaledosama999:
### Problem
Cache control doesn't support `stale-while-revalidate` directive


jansupol:
This comes from [RFC 5861](https://datatracker.ietf.org/doc/html/rfc5861), saying:
> This document is not an Internet Standards Track specification; it is
   published for informational purposes.

[RFC 9111 HTTP Caching](https://www.rfc-editor.org/rfc/rfc9111#name-cache-control), which is an Internet Standards Track document, does not know `stale-while-revalidate`.

Issue #1227, opened 2024-02-27
Add TCK for Convenience method for checking header value lists


jim-krueger:
TCK tests are needed for the function added under [PR 1066](https://github.com/jakartaee/rest/pull/1066)

mkarg:
@jim-krueger As that PR was merged, can we close this issue?

PR #1238, opened 2024-03-18
Add tests for containsHeaderString

jim-krueger:
Addresses #1227 

Awaiting formal testing.


jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
```diff
@@ -860,6 +861,53 @@ protected void checkFilterContext(ClientRequestContext context)
...
+        .header("header1", "value")
+        .header("header1", "value1 , value2")
+        .header("header1", "Value3,white space ")
+        .header("header2", "Value4::Value5")
+        .header("header3", "value6;value7")
+        .buildGet();
+    Response response = invoke(invocation);
+
+    String entity = response.readEntity(String.class);
+    assertTrue(entity.contains("Sucess"));
```
jansupol, line 907: `Success` needs double `c`

jansupol (changes requested):

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
```diff
@@ -882,7 +882,7 @@ protected void checkFilterContext(ClientRequestContext context) throws Fault {
           //Shouldn't find first value when separator character is incorrect
           assertFalse(context.containsHeaderString("header1", ";", "value1"::equalsIgnoreCase));
           //Test regular expression
-          assertFalse(context.containsHeaderString("header1", "; | ,", "value2"::equalsIgnoreCase));
+          assertFalse(context.containsHeaderString("header1", ";|,", "value2"::equalsIgnoreCase));
```
jansupol, line 885: `"value,value1 , value2,Value3,white space".split(";|,") = ["value", "value1 ", " value2", "Value3", "white space "]`
Why is this supposed to assert false?

jim-krueger: It shouldn't.  That was not my intent.  thanks

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientresponsecontext/JAXRSClientIT.java
```diff
@@ -426,6 +427,55 @@ protected void checkFilterContext(ClientRequestContext requestContext,
...
+   * 
+   * @assertion_ids:  JAXRS:JAVADOC:1355; JAXRS:JAVADOC:1356; 
+   * 
+   * @test_Strategy: Check if the specified header contains a specified value.
+   * 
+   * ClientRequestFilter.abortWith
+   */
+  @Test
+  public void containsHeaderStringTest() throws Fault {
+      final String header1 = "Header1";
```
jansupol, line 441: These are no longer used.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/HttpHeadersTest.java
```diff
@@ -67,6 +70,39 @@ public String headersGet() {
...
+        //Verify Predicate and separator character usage
+        assertTrue(hs.containsHeaderString("Accept", ",", "Text/html;Level=1"::equalsIgnoreCase));
+        sb.append("Header: accept contains text/html;level=1; ");
+        
+        //Verify incorrect separator character fails
+        assertFalse(hs.containsHeaderString("Accept", ";", "text/html;level=1"::equals));
+        sb.append("Incorrect separator character fails as expected; ");
+        
+        //Verify white space in parameter not trimmed
+        assertFalse(hs.containsHeaderString("header3", "**", "value3"::equalsIgnoreCase));
```
jansupol, line 92: I get the following:
```
java.util.regex.PatternSyntaxException: Dangling meta character '*' near index 0
**
^
	at java.base/java.util.regex.Pattern.error(Pattern.java:2095)
	at java.base/java.util.regex.Pattern.sequence(Pattern.java:2270)
	at java.base/java.util.regex.Pattern.expr(Pattern.java:2136)
	at java.base/java.util.regex.Pattern.compile(Pattern.java:1836)
	at java.base/java.util.regex.Pattern.<init>(Pattern.java:1467)
	at java.base/java.util.regex.Pattern.compile(Pattern.java:1092)
	at java.base/java.lang.String.split(String.java:3208)
	at java.base/java.lang.String.split(String.java:3290)
```
I'd better use that exception than consume it and return false.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/HttpHeadersTest.java
```diff
@@ -67,6 +70,39 @@ public String headersGet() {
...
+        sb.append("White space not trimmed from parameter as expected; ");
+        
+        //Verify white space in front and back of value trimmed
+        assertTrue(hs.containsHeaderString("HEADER3", "**", "value2"::equalsIgnoreCase));
```
jansupol, line 96: Same here, perhaps star is not the best reg exp separator.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/JAXRSClientIT.java
```diff
@@ -210,6 +210,26 @@ public void requestHeadersTest() throws Fault {
...
+  @Test
+  public void containsHeaderStringTest() throws Fault {
+    setProperty(Property.REQUEST_HEADERS,
+        "Accept:text/*, text/html, text/html;level=1, */*");
+    setProperty(Property.REQUEST_HEADERS,
+        "Content-Type:application/xml;charset=utf8");
+    setProperty(Property.REQUEST_HEADERS,
+        "Header3:value1 ** value2 **value 3");
+    setProperty(Property.REQUEST, buildRequest(Request.GET, "contains-headers"));
+    invoke();
```
jansupol, line 230: the test does not check the result, no-matter the exception being thrown on the server, the test just passes.

jansupol (changes requested):

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/HttpHeadersTest.java
jamezp: Yeah, that looks like a glob style pattern. A match all would be `.*`.

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientRequestContext.java
```diff
@@ -207,7 +207,8 @@ public default boolean hasProperty(String name) {
      * (missing comma), or the value {@code no - store} (whitespace within value).
      *
      * @param name the message header.
-     * @param valueSeparatorRegex Separates the header value into single values. {@code null} does not split.
+     * @param valueSeparatorRegex String or regular expression that separates the header value into single values. 
+     * {@code null} does not split.
```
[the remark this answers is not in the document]
jim-krueger: Changed wording.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
```diff
@@ -860,6 +861,53 @@ protected void checkFilterContext(ClientRequestContext context)
...
+   * @test_Strategy: Check if the specified header contains a specified value.
+   * 
+   * ClientRequestFilter.abortWith
+   */
+  @Test
+  public void containsHeaderStringTest() throws Fault {
+    ContextProvider provider = new ContextProvider() {
+      @Override
+      protected void checkFilterContext(ClientRequestContext context) throws Fault {
+          assertTrue(context.containsHeaderString("header1", "value"::equalsIgnoreCase));
```
mkarg, line 878: As "value" is already in the same case, this test for ignoring the case in not effective.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
```diff
@@ -860,6 +861,53 @@ protected void checkFilterContext(ClientRequestContext context)
...
+          assertTrue(context.containsHeaderString("HEADER1", ",", "value2"::equals));
+          //Incorrect separator character
```
mkarg, line 880: IMHO a TCK test should only test for the existence of a given rule, not for the absence of particular bugs.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
```diff
@@ -860,6 +861,53 @@ protected void checkFilterContext(ClientRequestContext context)
...
+          //Shouldn't find first value when separator character is incorrect
+          assertFalse(context.containsHeaderString("header1", ";", "value1"::equalsIgnoreCase));
+          //Test regular expression
+          assertTrue(context.containsHeaderString("header1", ";|,", "value2"::equalsIgnoreCase));
+          //White space in value not trimmed
+          assertFalse(context.containsHeaderString("header1", "whitespace"::equalsIgnoreCase));
+          //Multiple character separator
+          assertTrue(context.containsHeaderString("header2", "::", "Value5"::equalsIgnoreCase));
+          //Test default separator is comma
+          assertFalse(context.containsHeaderString("header3","value6"::equalsIgnoreCase));
```
mkarg, line 891: I cannot see how this line actually tests what its comment says.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
```diff
@@ -860,6 +861,53 @@ protected void checkFilterContext(ClientRequestContext context)
...
+        .header("header1", "value")
+        .header("header1", "value1 , value2")
+        .header("header1", "Value3,white space ")
+        .header("header2", "Value4::Value5")
+        .header("header3", "value6;value7")
+        .buildGet();
+    Response response = invoke(invocation);
+
+    String entity = response.readEntity(String.class);
+    assertTrue(entity.contains("Success"));
```
mkarg, line 907: This assertion is needless as it does not test something that is related to containsHeaderString (if containsHeaderStringwould be implemented incorrectly, another assertion would have noticed this at this point already).

mkarg (changes requested):

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
jim-krueger: Agreed.  Changed to ::equals

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
jim-krueger: This test is ensuring that if a separator parameter is not set, that it will default to comma and containsHeaderString in this case will return "false" because ";" is the separator.   However, it is somewhat redundant so I have no problem removing it.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
jim-krueger: Agreed.  It is unnecessary

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/HttpHeadersTest.java
jim-krueger: Agreed

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/HttpHeadersTest.java
jim-krueger: Agreed.  Changed to "::" separator

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
jim-krueger: I think this test is valid since it confirms that the value of the separator character parameter is actually used and taken into consideration by the impl.

mkarg: "Incorrect separator character" does not imply that. Checking for correct pickup of separator is implied in line 879 already.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
mkarg: I am not convinced that this test actually proofs usage of tha specific default separator. IMHO it only proofs that this lines fails, independent of the cause.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
jim-krueger: Like 879 only proves that the line separator parameter can be included in the invocation, but not that it is actually used by the impl.  The impl could be ignoring that parameter and accepting any separator character as valid.  This is why the "incorrect separator character" test is good.  It tests that the impl is using the the separator parameter and that it is checking it specifically and not just allowing any separator.

mkarg: I wonder if instead there could be a test that actively checks usage of the separator, and positively comments about that.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientresponsecontext/JAXRSClientIT.java
jim-krueger: agreed

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/JAXRSClientIT.java
jim-krueger: Thanks.   I believe this is addressed.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
jim-krueger: I see your point, I'm just not sure there is a way to ensure that the separator parameter is actually being used without having at lease one test that expects failure in that respect.

jamezp (approved):
This looks good to me. There are a couple spots where there are some whitespace only changes, but I don't want to block this PR based on that.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/container/responsecontext/ResponseFilter.java
```diff
@@ -152,6 +152,37 @@ else if (type != null)
     setEntity(name);
   }
 
+  public void containsHeaderString() {
+      StringBuffer sb = new StringBuffer();
+      sb.append("containsHeaderString= ");
+
+      try {
+          assertTrue(requestContext.containsHeaderString("accept", "text/html"::equals));
```
jansupol, line 160: This should have been tested on responseContext, should it not?

jansupol (approved):

jamezp (approved):

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
spericas: To me the test makes sense. If you split with the wrong separator (so you don't really split) then the assertion is false.

spericas (approved):

spericas:
@mkarg Your change request is now blocking this PR

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
mkarg: That is correct, but just because using the wrong separator fails the test does not imply that using the correct separator passes the test.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
```diff
@@ -860,6 +861,46 @@ protected void checkFilterContext(ClientRequestContext context)
...
+          assertFalse(context.containsHeaderString("header1", ";", "value2"::equals));
+          //Shouldn't find first value when separator character is incorrect
+          assertFalse(context.containsHeaderString("header1", ";", "Value1"::equalsIgnoreCase));
+          //Test regular expression
+          assertTrue(context.containsHeaderString("header1", ";|,", "VALUE2"::equalsIgnoreCase));
+          //White space in value not trimmed
+          assertFalse(context.containsHeaderString("header1", "whitespace"::equals));
+          //Multiple character separator
+          assertTrue(context.containsHeaderString("header2", ";;", "Value5"::equalsIgnoreCase));
+          Response r = Response.ok().build();
```
mkarg, line 890: Will the test fail when we remove context.abortWith? If not, we can remove it.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/container/requestcontext/RequestFilter.java
```diff
@@ -149,7 +180,7 @@ public void getHeaders() {
     }
     abortWithEntity(sb.toString());
   }
-
+
```
mkarg, line 183: Excess blank.

mkarg (approved):
I am not used to writing TCK tests, so I am not actively voting +1, but feel free to proceed anyways.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/clientrequestcontext/JAXRSClientIT.java
jim-krueger: > That is correct, but just because using the wrong separator fails the test does not imply that using the correct separator passes the test.
Yes, but using the correct separator is tested in other variations.

> Will the test fail when we remove context.abortWith? If not, we can remove it.
Yes, the test will fail if context.abortWith is removed.  If abortWith is removed several tests in this class will also fail and need to be modified.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/container/requestcontext/RequestFilter.java
jim-krueger: Normally I would remove this change, but all other methods appear to be separated by a blank line so I think it is appropriate to just leave this.

mkarg: This is no a blank line, it is a single blank before EOL, and I am used to remove those because some open source projects fail their QA in that case. :-)

jim-krueger: Thanks for explaining.   I'll fix this under #1248 since this PR has been merged.

PR #1240, opened 2024-03-19
[1239] Use the correct Variant constructor for the header content lan…

### LINKED ISSUE #1239: TCK Challenge: ee.jakarta.tck.ws.rs.ee.rs.core.headers.JAXRSClientIT.contentLanguageTest uses the wrong Variant arguments
In the `ee.jakarta.tck.ws.rs.ee.rs.core.headers.JAXRSClientIT.contentLanguageTest` when the `Variant` is constructed the language "en-US" is passed. However, in the JavaDoc for the argument it states:

> `@param` language the language of the variant (two-letter ISO-639 code); may be `{@code null}`.

The test should be using `new Variant(MediaType.WILDCARD_TYPE, "en", "US", null);` instead of `new Variant(MediaType.WILDCARD_TYPE, "en-US", null);`.

This was found in RESTEasy with a change to use language tags for locales, for reference https://github.com/resteasy/resteasy/pull/4054.

jamezp:
…guage TCK test.

resolves #1239 

This potentially may need to be back ported to 3.1 as well.

### DIFF
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/JAXRSClientIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/core/headers/JAXRSClientIT.java
@@ -140,7 +140,7 @@ public class JAXRSClientIT extends JaxrsCommonClient {
   public void contentLanguageTest() throws Fault {
     setProperty(Property.REQUEST_HEADERS,
         buildAccept(MediaType.TEXT_PLAIN_TYPE));
-    Variant variant = new Variant(MediaType.WILDCARD_TYPE, "en-US", null);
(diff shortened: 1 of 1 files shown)


spericas (approved):
Indeed, this looks like a bug.

spericas:
@alwin-joseph Do you agree?

jim-krueger (approved):

jansupol (approved):

alwin-joseph:
> @alwin-joseph Do you agree?

Agree.

I think the test was passing with previous changes as java.util.Locale accepted "en-US" also as language at https://github.com/jakartaee/rest/blob/c97bac8a5bc91628611bd46e71890ab16986e394/jaxrs-api/src/main/java/jakarta/ws/rs/core/Variant.java#L55.

alwin-joseph (approved):

PR #1224, opened 2024-02-21
Update the TCK generation , remove old shell script (tckbundle.sh)

alwin-joseph:
**Describe the changes:**

- Moved jaxrs-tck-docs as jaxrs-tck/docs
- To generate userguide : `mvn -f jaxrs-tck/docs/userguide` 
- To build TCK jar and zip bundle (with EFTL) : 
`cd jaxrs-tck; mvn clean install`
- To build TCK jar and zip bundle (with EPL) : 
`cd jaxrs-tck; mvn clean install -PEPL`

PR #1241, opened 2024-03-20
fast-track: Fix the including of resources in the TCK.

jamezp:
This was caused by #1224. Just a minor change from `com/` to `ee/` since the package name starts with `ee`.

### DIFF
--- a/jaxrs-tck/pom.xml
+++ b/jaxrs-tck/pom.xml
@@ -113,7 +113,7 @@
                 <directory>src/main/resources</directory>
                 <includes>
                     <include>LICENSE_${license}.md</include>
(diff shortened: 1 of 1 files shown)


jim-krueger (approved):

jim-krueger:
@alwin-joseph If you approve of this change I think it should be merged.

alwin-joseph (approved):
+1 , Looks like it was missed in my previous change. Thanks for picking this.

jansupol (approved):

jim-krueger:
Given that this fixes multiple TCK failures I think it should be merged without delay.

PR #1243, opened 2024-03-23
TCK : Get hostname and port number injected using ArquillianResource url 

alwin-joseph:
- Avoids hardcoding the webServerHost and webServerPort as system properties.
- Tested with Glassfish/Jersey only now.

### DIFF
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/ext/interceptor/reader/interceptorcontext/JAXRSClientIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/ext/interceptor/reader/interceptorcontext/JAXRSClientIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -50,10 +50,6 @@ public class JAXRSClientIT extends ReaderClient<ContextOperation> {
 
   private static final long serialVersionUID = -8828149277776372718L;
 
-  public JAXRSClientIT() {
-    setup();
-  }
-
   @BeforeEach
   void logStartTest(TestInfo testInfo) {
     TestUtil.logMsg("STARTING TEST : "+testInfo.getDisplayName());
@@ -62,6 +58,15 @@ public class JAXRSClientIT extends ReaderClient<ContextOperation> {
   @AfterEach
   void logFinishTest(TestInfo testInfo) {
     TestUtil.logMsg("FINISHED TEST : "+testInfo.getDisplayName());
+  }
+
+  @BeforeEach
+  public void setup() {
+    // Only the values are verified, using verifySettings() in JaxrWebTestCase.java
+    // but the URL with these values is not required to work. 
+    // TODO: remove validation of these values for the tests in this file.
+    _hostname = "localhost";
+    _port = 8080;
   }
   
   /*
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/resource/java2entity/JAXRSClientIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/resource/java2entity/JAXRSClientIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2020, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -46,8 +46,12 @@ public class JAXRSClientIT extends JAXRSCommonClient {
   private static final long serialVersionUID = 1L;
 
   public JAXRSClientIT() {
-    setup();
     setContextRoot("/jaxrs_resource_java2entity_web/resource");
+  }
+
+  @BeforeEach
+  public void setup() {
+    super.setup();
   }
 
   @Deployment(testable = false)
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/cookieparam/locator/JAXRSLocatorClientIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/cookieparam/locator/JAXRSLocatorClientIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -17,6 +17,7 @@
 package ee.jakarta.tck.ws.rs.ee.rs.cookieparam.locator;
 
 import java.io.InputStream;
+import java.net.URL;
 import java.io.IOException;
 
 import ee.jakarta.tck.ws.rs.ee.rs.JaxrsParamClient;
@@ -34,11 +35,14 @@ import ee.jakarta.tck.ws.rs.ee.rs.cookieparam.JAXRSClientIT;
 import ee.jakarta.tck.ws.rs.lib.util.TestUtil;
 
 import org.jboss.arquillian.junit5.ArquillianExtension;
+import org.jboss.arquillian.test.api.ArquillianResource;
 import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.OperateOnDeployment;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.StringAsset;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 import org.junit.jupiter.api.Test;
@@ -59,21 +63,33 @@ public class JAXRSLocatorClientIT
   private static final long serialVersionUID = 1L;
 
   public JAXRSLocatorClientIT() {
-    setup();
     setContextRoot("/jaxrs_ee_rs_cookieparam_locator_web/resource/locator");
   }
 
+  @ArquillianResource
+  @OperateOnDeployment("jaxrs_ee_rs_cookieparam_locator_web")
+  private URL url;
+
   @BeforeEach
-  void logStartTest(TestInfo testInfo) {
-    TestUtil.logMsg("STARTING TEST : "+testInfo.getDisplayName());
+  public void setup() {
+
+    TestUtil.logTrace("setup method JAXRSLocatorClientIT");
+
+    assertFalse((url==null), "[JAXRSLocatorClientIT] 'url' was not injected.");
+    
+    String hostname = url.getHost();
+    String portnum = Integer.toString(url.getPort());
+    
+    assertFalse(isNullOrEmpty(hostname), "[JAXRSLocatorClientIT] 'webServerHost' was not set.");
+    _hostname = hostname.trim();
+    assertFalse(isNullOrEmpty(portnum), "[JAXRSLocatorClientIT] 'webServerPort' was not set.");
+    _port = Integer.parseInt(portnum.trim());
+    TestUtil.logMsg("[JAXRSLocatorClientIT] Test setup OK");
+
   }
 
-  @AfterEach
-  void logFinishTest(TestInfo testInfo) {
-    TestUtil.logMsg("FINISHED TEST : "+testInfo.getDisplayName());
-  }
 
-  @Deployment(testable = false, name = "cookieparamlocator")
+  @Deployment(testable = false, name="jaxrs_ee_rs_cookieparam_locator_web")
   public static WebArchive createDeployment() throws IOException{
 
     InputStream inStream = JAXRSLocatorClientIT.class.getClassLoader().getResourceAsStream("ee/jakarta/tck/ws/rs/ee/rs/cookieparam/locator/web.xml.template");
--- a/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/formparam/sub/JAXRSSubClientIT.java
+++ b/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/formparam/sub/JAXRSSubClientIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -27,15 +27,19 @@ import ee.jakarta.tck.ws.rs.ee.rs.WebApplicationExceptionMapper;
 import ee.jakarta.tck.ws.rs.ee.rs.formparam.FormParamTest;
 import ee.jakarta.tck.ws.rs.ee.rs.formparam.JAXRSClientIT;
 import java.io.InputStream;
+import java.net.URL;
 import java.io.IOException;
 import ee.jakarta.tck.ws.rs.lib.util.TestUtil;
 
 import org.jboss.arquillian.junit5.ArquillianExtension;
+import org.jboss.arquillian.test.api.ArquillianResource;
 import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.OperateOnDeployment;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.StringAsset;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 import org.junit.jupiter.api.Test;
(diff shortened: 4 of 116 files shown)


spericas:
@alwin-joseph Seems like the preferred way now is to drop the second year in the copyright header, instead of updating it as you have done.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/ext/interceptor/reader/interceptorcontext/JAXRSClientIT.java
```diff
@@ -63,6 +59,15 @@ void logStartTest(TestInfo testInfo) {
   void logFinishTest(TestInfo testInfo) {
     TestUtil.logMsg("FINISHED TEST : "+testInfo.getDisplayName());
   }
+
+  @BeforeEach
```
jbescos, line 63: There is another @BeforeEach in line 53. Are they both going to be invoked by the test runner?.

jbescos:
I tried it with Helidon and @alwin-joseph branch:

```
[ERROR] Errors: 
[ERROR]   MultipartSupportIT.basicTest » Runtime Could not lookup value for field private java.net.URI ee.jakarta.tck.ws.rs.jaxrs31.ee.multipart.MultipartSupportIT.uri
[ERROR]   MultipartSupportIT.multiFormParamTest » Runtime Could not lookup value for field private java.net.URI ee.jakarta.tck.ws.rs.jaxrs31.ee.multipart.MultipartSupportIT.uri
[ERROR]   JAXRSSigTestIT.signatureTest:415 Fault signatureTest failed with an unexpected exception
[INFO] 
[ERROR] Tests run: 2649, Failures: 0, Errors: 3, Skipped: 128
```

It seems MultipartSupportIT was added in other commit, so I guess from Helidon point of view this PR looks fine:
```
commit 4a3ca8c9856a2d0d16d39ff722e30e5a7807291b
Author: Jim Krueger <31767744+jim-krueger@users.noreply.github.com> 2024-03-20 15:14:07
Committer: GitHub <noreply@github.com> 2024-03-20 15:14:07
Branches: alwin/javadoc_assertions, alwin/TCK_ArquillianURLInjection

fast-track: Add multipart EntityPart tests (#1230)

* Add multipart EntityPart tests

* Fix tabs to spaces

* Fix issue when changing/adding headers to an EntityPart

* Fix code review comments.
```

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/ext/interceptor/reader/interceptorcontext/JAXRSClientIT.java
jamezp: They should definitely both run, but there is no reason not to combine them IMO.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/resource/java2entity/JAXRSClientIT.java
```diff
@@ -46,10 +46,14 @@ public class JAXRSClientIT extends JAXRSCommonClient {
...
   public JAXRSClientIT() {
-    setup();
     setContextRoot("/jaxrs_resource_java2entity_web/resource");
   }
 
+  @BeforeEach
+  public void setup() {
+    super.setup();
+  }
+
```
jamezp, line 56: This shouldn't be required.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/formparam/sub/JAXRSSubClientIT.java
```diff
@@ -52,22 +56,33 @@
...
+    _port = Integer.parseInt(portnum.trim());
+    TestUtil.logMsg("[JAXRSSubClientIT] Test setup OK");
 
-  @AfterEach
-  void logFinishTest(TestInfo testInfo) {
-    TestUtil.logMsg("FINISHED TEST : "+testInfo.getDisplayName());
   }
 
+  private static final long serialVersionUID = 1L;
+
```
jamezp, line 85: This shouldn't be required.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/cookieparam/locator/JAXRSLocatorClientIT.java
```diff
@@ -59,21 +63,33 @@ public class JAXRSLocatorClientIT
...
-    TestUtil.logMsg("STARTING TEST : "+testInfo.getDisplayName());
-  }
+  public void setup() {
+
+    TestUtil.logTrace("setup method JAXRSLocatorClientIT");
+
+    assertFalse((url==null), "[JAXRSLocatorClientIT] 'url' was not injected.");
+    
+    String hostname = url.getHost();
+    String portnum = Integer.toString(url.getPort());
```
jamezp, line 81: This pattern is in several places and I wonder if it should be simplified. Instead of converting it to a string, then back to an integer, we could just use the integer and check if the port is > 0. The `isNullOrEmpty(portnum)` will never be false as the `Integer.toString()` will never return an empty string or `null`.

jamezp (dismissed):
I think this makes a lot of sense. I made a couple comments that IMO should not block merging given our short time frame.

The two main things I've seen is most of the simple invocations to the `super.setup()` are not required as the `@BeforeEach` will be inherited.

Also the pattern of port -> string back to an integer doesn't give us much for the `isNullOrEmpty()` as it will never be `null` or empty.

I also ran this change against RESTEasy and it passes all but the known issues with the signature tests.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/cookieparam/locator/JAXRSLocatorClientIT.java
alwin-joseph: Agree, latest commit resolves it.

jamezp (dismissed):

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/rs/formparam/sub/JAXRSSubClientIT.java
alwin-joseph: Agree. It was not introduced in this PR, diff has the same line removed at the beginning. Hopefully these along with other occurrences can be addressed in later changes.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/resource/java2entity/JAXRSClientIT.java
alwin-joseph: The parent setup() method in jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/common/JAXRSCommonClient.java need to be invoked only for test classes that deploys archives and not others.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/spec/client/webtarget/JAXRSClientIT.java
```diff
@@ -44,10 +44,8 @@
 public class JAXRSClientIT extends JAXRSCommonClient {
 
   public JAXRSClientIT() {
-    setup();
```
jansupol, line 47: The setup() is not called in this class?

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/ee/resource/java2entity/JAXRSClientIT.java
jamezp: Sorry, for some reason I thought it was also annotated with `@BeforeEach`. I must have looked at the wrong file.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/spec/client/instance/JAXRSClientIT.java
```diff
@@ -71,7 +70,6 @@ public JAXRSClientIT() {
 
   }
 
-
```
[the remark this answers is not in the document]
jamezp: It's not an Arquillian test so it would produce an assertion failure as the URL would be `null`. I don't see the ports/host being used here so it's likely not a big issue here.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/spec/client/webtarget/JAXRSClientIT.java
jamezp: Same with this one, it doesn't appear to be an Arquillian test.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/ext/interceptor/reader/interceptorcontext/JAXRSClientIT.java
jansupol: I am more confused about removing the call of a logic for getting hostname and port and hard-coding the hostname and port here.

jamezp:
I'm not Alwin, but I can answer some of this :)

> @alwin-joseph
> 
> * How does that work? Where does the Arquillian get the hostname and port from? Can the default "localhost:8080" be changed?

It's pulled from the `org.jboss.arquillian.container.spi.client.protocol.metadata.HTTPContext`. Generally this is initialized in the Arquillian protocol being used during the server startup. To my knowledge it's the responsibility of the Arquillian adapter to create this. In WildFly Arquillian we get it from our server configuration after the server has been started.

> * Is it necessary to parse the hostname and port for each test? Is it not possible to do it once for the test class in the constructor as it was done?

It more than likely depends, but we could try it with a static `@BeforeAll`. The depending part depends on how the container is managed by Arquillian.

alwin-joseph:
> @alwin-joseph
> 
>     * How does that work? Where does the Arquillian get the hostname and port from? Can the default "localhost:8080" be changed?
> 
IIUC the Arquillian adapter for each container handles it. The current jersey-tck/ module uses `org.jboss.arquillian.container:arquillian-glassfish-managed-6` to run with Glassfish.

We could also try `org.omnifaces.arquillian:arquillian-glassfish-server-managed:1.2` (also used for JSP TCK) where we need to set glassfish.home as system property to locate the glassfish server. The properties for glassfish is set in this connector at https://github.com/OmniFish-EE/arquillian-container-glassfish/blob/36b96e6a9501c1029e76f3eabb746a64695f4920/glassfish-managed/src/main/java/org/omnifaces/arquillian/container/glassfish/managed/GlassFishManagedContainerConfiguration.java#L78C1-L90C96 as suggested [here](https://github.com/jakartaee/platform-tck/pull/1182/#discussion_r1382629703). From what I see the default httpport is 8080 here, glassfish.httpPort is the system property that can be overridden to set the port.
 

> > * Is it necessary to parse the hostname and port for each test? Is it not possible to do it once for the test class in the constructor as it was done?
> 
> It more than likely depends, but we could try it with a static `@BeforeAll`. The depending part depends on how the container is managed by Arquillian.

Agree static `@BeforeAll` can be tried but it looked to me as a lot of places needing changes.

alwin-joseph:
> @alwin-joseph
> 
>     * How does that work? Where does the Arquillian get the hostname and port from? Can the default "localhost:8080" be changed?
> 
>     * Is it necessary to parse the hostname and port for each test? Is it not possible to do it once for the test class in the constructor as it was done?

The url is injected with `@OperateOnDeployment` so invoking setup in constructor didn't work for me as the static createDeployment method was not complete then.

spericas (approved):

jansupol:
I am not very happy with this PR, but maybe I do not see something.

Basically this PR:

- drops calling `setup()` from the constructor and instead it calls it for each  test method 
- I do not know HOW EXACTLY the port is changed for instance for running on an app server that runs on a different server than 8080
- it hard codes the ports in some tests
- it comes right at a hectic time of finishing the release, the PR is huge and needs extra time for testing, and I do not see an urge to have it in 4.0

jamezp:
FWIW JUnit constructs the test for each test method anyway so `setup()` in a constructor is effectively the same as a `@BeforeEach`. 

The port would be changed in the container and the test would get the correct URL, with the correct deployment context path.

The hard-coded host/ports should probably be configurable via a system property.

That said, I agree this could wait.

jansupol:
> FWIW JUnit constructs the test for each test method anyway so setup() in a constructor is effectively the same as a @BeforeEach

That's my bad, then.

> The port would be changed in the container and the test would get the correct URL, with the correct deployment context path.

Which container? The Arquillian container, the WebServer, or some other? I really would think it is changed in arquillian.xml.

jamezp:
> > FWIW JUnit constructs the test for each test method anyway so setup() in a constructor is effectively the same as a `@beforeeach`
> 
> That's my bad, then.

Nah, it's definitely not well known and honestly took me by surprised when I was debugging once :)

> 
> > The port would be changed in the container and the test would get the correct URL, with the correct deployment context path.
> 
> Which container? The Arquillian container, the WebServer, or some other? I really would think it is changed in arquillian.xml.

By container, I mean the application server. Arquillian has a `HTTPContext` that is created from which Arquillian Adaptor is used. I WildFly what we do is look at the Undertow subsystem to get the HTTP port number. I assume something similar would be done in GlassFish as well.

I will say this only works for container tests. API tests that are not run with Arquillian wouldn't work.

jansupol:
> By container, I mean the application server. Arquillian has a HTTPContext that is created from which Arquillian Adaptor is used. I WildFly what we do is look at the Undertow subsystem to get the HTTP port number. I assume something similar would be done in GlassFish as well.

This is hard to imagine tbh. whatever lib lies beneath needs to ask the server, but the server needs to listen on some port. If the port is changed, there would need to be some way to tell the arquillian whom to ask. Or instead, tell the Arquillian the port directly. There is the following settings in the arquillian.xml:

```
<container qualifier="...." default="true">
  <configuration>   
   <property name="managementPort">9991</property>
   <property name="username">admin</property>
   <property name="password">admin</property>
  </configuration>
  <protocol type="Servlet 3.0">
   <property name="port">8081</property>
  </protocol>
</container> 
```

Which I assume is what is needed, but I would need to test it.

> I will say this only works for container tests. API tests that are not run with Arquillian wouldn't work.

The API tests do not need the server, so I assume either port is fine.

jamezp:
> This is hard to imagine tbh. whatever lib lies beneath needs to ask the server, but the server needs to listen on some port. If the port is changed, there would need to be some way to tell the arquillian whom to ask. Or instead, tell the Arquillian the port directly. There is the following settings in the arquillian.xml:
> 
> ```
> <container qualifier="...." default="true">
>   <configuration>   
>    <property name="managementPort">9991</property>
>    <property name="username">admin</property>
>    <property name="password">admin</property>
>   </configuration>
>   <protocol type="Servlet 3.0">
>    <property name="port">8081</property>
>   </protocol>
> </container> 
> ```
> 
> Which I assume is what is needed, but I would need to test it.

tl;dr

Yes, the URL is created based on the `arquillian.xml` configuration.

Longer answer:

Yes, that is how it should work. Arquillian works on the observer pattern and it sends events. It sends an event to start the server and the `org.jboss.arquillian.container.spi.client.container.DeployableContainer` implementation handles starting the server based on the `org.jboss.arquillian.container.spi.client.container.ContainerConfiguration` setting, which is created from the `arquillian.xml` file.

For injecting the URL, that happens after the deployment has been done. The `DeployableContainer.deploy()` method, returns a `ProtocolMetaData` which is where the `HTTPContext` would be attached to the `ProtocolMetaData`. That `HTTPConext` is used to get the `URL`.

The port itself can only be changed before the server is started, so we don't really need to be too concerned about it changing.

jansupol:
I still do not understand the purpose. Instead of setting the port in the `pom.xml` we will set the port in the `arquillian.xml`.
For that, the tests do 

1. instead of calling `super.setup()` they do

```
  @BeforeEach
  public void setup() {
    super.setup();
  }
```

2. Many tests include additional logic in a setup method distributed among them:
```
 public void setup() {

    TestUtil.logTrace("setup method JAXRSSubClientIT");

    assertFalse((url==null), "[JAXRSSubClientIT] 'url' was not injected.");

    String hostname = url.getHost();
    int portnum = url.getPort();

    assertFalse(isNullOrEmpty(hostname), "[JAXRSSubClientIT] 'webServerHost' was not set.");
    _hostname = hostname.trim();
    assertFalse(isPortInvalid(portnum), "[JAXRSSubClientIT] 'webServerPort' was not set.");
    _port = portnum;
    TestUtil.logMsg("[JAXRSSubClientIT] Test setup OK");
```
3. The tests no longer contain information messages
```
TestUtil.logMsg("STARTING TEST : "+testInfo.getDisplayName());
TestUtil.logMsg("FINISHED TEST : "+testInfo.getDisplayName());
```
which might be intentional, or might be moved to the super-class, but the PR description does not say.

jamezp:
> I still do not understand the purpose. Instead of setting the port in the `pom.xml` we will set the port in the `arquillian.xml`. For that, the tests do
> 
> 1. instead of calling `super.setup()` they do
> 
> ```
>   @BeforeEach
>   public void setup() {
>     super.setup();
>   }
> ```
> 
> 2. Many tests include additional logic in a setup method distributed among them:
> 
> ```
>  public void setup() {
> 
>     TestUtil.logTrace("setup method JAXRSSubClientIT");
> 
>     assertFalse((url==null), "[JAXRSSubClientIT] 'url' was not injected.");
> 
>     String hostname = url.getHost();
>     int portnum = url.getPort();
> 
>     assertFalse(isNullOrEmpty(hostname), "[JAXRSSubClientIT] 'webServerHost' was not set.");
>     _hostname = hostname.trim();
>     assertFalse(isPortInvalid(portnum), "[JAXRSSubClientIT] 'webServerPort' was not set.");
>     _port = portnum;
>     TestUtil.logMsg("[JAXRSSubClientIT] Test setup OK");
> ```
> 
> 3. The tests no longer contain information messages
> 
> ```
> TestUtil.logMsg("STARTING TEST : "+testInfo.getDisplayName());
> TestUtil.logMsg("FINISHED TEST : "+testInfo.getDisplayName());
> ```
> 
> which might be intentional, or might be moved to the super-class, but the PR description does not say.

These are some very good points and is likely a reason we should wait until after 4.0 for this given our very short time frame. I do think the issues will be addressable, but I don't feel we should rush them.

alwin-joseph:
> 
>     2. Many tests include additional logic in a setup method distributed among them:
> ```
>  public void setup() {
> 
>     TestUtil.logTrace("setup method JAXRSSubClientIT");
> 
>     assertFalse((url==null), "[JAXRSSubClientIT] 'url' was not injected.");
> 
>     String hostname = url.getHost();
>     int portnum = url.getPort();
> 
>     assertFalse(isNullOrEmpty(hostname), "[JAXRSSubClientIT] 'webServerHost' was not set.");
>     _hostname = hostname.trim();
>     assertFalse(isPortInvalid(portnum), "[JAXRSSubClientIT] 'webServerPort' was not set.");
>     _port = portnum;
>     TestUtil.logMsg("[JAXRSSubClientIT] Test setup OK");
> ```
> 

This does the same as the common setup method in ee.jakarta.tck.ws.rs.common.JAXRSCommonClient.

In some scenarios the test class files(*IT.java) get inherited from other test files (eg: ee.jakarta.tck.ws.rs.ee.rs.cookieparam.sub.JAXRSSubClientIT -> ee.jakarta.tck.ws.rs.ee.rs.cookieparam.JAXRSClientIT )
and they both have deployments; some deployments were named and others were "_DEFAULT_" which caused issues with the `@OperateOnDeployment("<DeploymentName>")`.

I dont know if removing `@OperateOnDeployment` for injecting the URL will cause issues with any implementations, but that works for me. In this case the URL will be annotated with only `@ArquillianResource`.


>     3. The tests no longer contain information messages
> ```
> TestUtil.logMsg("STARTING TEST : "+testInfo.getDisplayName());
> TestUtil.logMsg("FINISHED TEST : "+testInfo.getDisplayName());
> ```
> 
> which might be intentional, or might be moved to the super-class, but the PR description does not say.

Yes, this is something that could be moved to super-class ee.jakarta.tck.ws.rs.common.JAXRSCommonClient, which was not intended to be in this PR, I can add it here.

> These are some very good points and is likely a reason we should wait until after 4.0 for this given our very short time frame. I do think the issues will be addressable, but I don't feel we should rush them.

I agree we can address these issues after 4.0, my only concern is that to make large changes to the TCK we will have to wait till next major release. But considering there are real concerns with regards to this PR we can hold this off for later time.

jansupol:
So it sounds like instead of copying the setup logic over to sub-classes, it would be better to change the DEFAULT deployment name to some unique one.

From what I understand from the conversation related to the current release, we want to start working on 5.0 somewhat soonish so that we are ready for EE12, as the transition to CDI is rather a big deal. We will have time to revisit this and address the concerns.

spericas:
@alwin-joseph @jansupol @jamezp @jbescos @jim-krueger Is anyone against pushing this PR to 5.0?

jamezp:
Pushing off to 5.0 seems fine with me.

PR #1270, opened 2024-06-11
Avoid using old synchronized java.util.Stack

turbanoff:
Use modern ArrayDeque instead

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/GenericType.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/GenericType.java
@@ -21,8 +21,8 @@ import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
+import java.util.ArrayDeque;
 import java.util.Arrays;
-import java.util.Stack;
 
 /**
  * Represents a generic message entity type {@code T}.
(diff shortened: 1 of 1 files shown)


jaxrs-api/src/main/java/jakarta/ws/rs/core/GenericType.java
```diff
@@ -198,7 +198,7 @@ private static Class getArrayClass(final Class c) {
      */
     static Type getTypeArgument(final Class<?> clazz, final Class<?> baseClass) {
         // collect superclasses
-        Stack<Type> superclasses = new Stack<Type>();
+        ArrayDeque<Type> superclasses = new ArrayDeque<>();
```
mkarg, line 201: Why forcing array-based deque implementation on the *left* side, instead of simply using an interface as before, or using `var`?

mkarg (approved):
LGTM.

jaxrs-api/src/main/java/jakarta/ws/rs/core/GenericType.java
turbanoff: >instead of simply using an interface as before

Before it was using a Class - `java.util.Stack`. It's not an interface.

mkarg: You're right, I missed that.

jansupol (approved):
Very much +1

PR #1288, opened 2024-10-02
closes #1266, implement Form equals method and add unit tests

### LINKED ISSUE #1266: Form class missing equals impl
Hi, Coud you please impl an equals method to Form class 

https://github.com/jakartaee/rest/blob/main/jaxrs-api/src/main/java/jakarta/ws/rs/core/Form.java

david-walters-14:


jaxrs-api/src/main/java/jakarta/ws/rs/core/Form.java
```diff
@@ -93,4 +93,19 @@ public Form param(final String name, final String value) {
...
     }
+
+    /**
+     * Compare for equality
+     * 
+     * @param object the object to compare to.
+     * @return {@code true}, if the object is a {@code Form} with the same form parameters, {@code false} otherwise.
+     */
+    @Override
+    public boolean equals(Object object) {
```
jansupol, line 104: I do believe that when `equals` is being implemented, the `hashCode` should be implemented as well.

jaxrs-api/src/test/java/jakarta/ws/rs/core/FormTest.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (c) 2010, 2021 Oracle and/or its affiliates. All rights reserved.
```
jansupol, line 2: A new test would better have the year 2024.

jansupol (changes requested):

jaxrs-api/src/main/java/jakarta/ws/rs/core/Form.java
```diff
@@ -93,4 +93,19 @@ public Form param(final String name, final String value) {
...
+        if (object == null) return false;
+        if (object == this) return true;
+        if (getClass() != object.getClass()) return false;
+        Form form = (Form) object;
+        return parameters.equals(form.asMap());
```
mkarg, line 109: Why not simply `return parameters.equals(form.parameters)`?

jaxrs-api/src/test/java/jakarta/ws/rs/core/FormTest.java
mkarg, line 2: Copyright must be date of first publication, which is 2024, not 2024. Also, unless you are Oracle, it must mention the original author, which are you.

jaxrs-api/src/main/java/jakarta/ws/rs/core/Form.java
mkarg: Indeed, the JavaDocs say:
>Note that it is generally necessary to override the hashCode method whenever this method is overridden, so as to maintain the general contract for the hashCode method, which states that equal objects must have equal hash codes.

jaxrs-api/src/test/java/jakarta/ws/rs/core/FormTest.java
jansupol: Strictly speaking, the author can consider himself an affiliate, so the copyright is ok. But sure, the original author's name works, too.

jaxrs-api/src/main/java/jakarta/ws/rs/core/Form.java
```diff
@@ -93,4 +93,29 @@ public Form param(final String name, final String value) {
     public MultivaluedMap<String, String> asMap() {
         return parameters;
     }
+
```
mkarg, line 96: As this file has a two-year copyright note, the EF's rules require that you update it. To make future updates easier, it is best if you simply change to to a one-year copyright note.

mkarg (changes requested):

mkarg (approved):

jaxrs-api/src/main/java/jakarta/ws/rs/core/Form.java
```diff
@@ -93,4 +93,29 @@ public Form param(final String name, final String value) {
...
+    }
+
+    /**
+     * Returns the hashCode of the underlying form parameters.
+     * 
+     * @return a hash code value for the underlying form parameters.
+     */
+    @Override
+    public int hashCode() {
+        return parameters.hashCode();
```
jamezp, line 119: IMO we should use something like `Objects.hash(parameters)`.

mkarg: What would be the benefit over the current draft?

jamezp: It's not critical by any means, it's just not the standard pattern to not calculate a hash like this. I suppose that's up to opinion though :) This should not block this FWIW.

jansupol (approved):

david-walters-14:
Assuming there are no further changes required, would someone mind merging this PR?

mkarg:
> Assuming there are no further changes required, would someone mind merging this PR?

According to our general [Committer Conventions](https://github.com/jakartaee/rest/wiki/Committer-Conventions#minimum-length-of-review-period-before-merge--close-of-prs-and-closing-of-issues) the default voting period is two full weeks. Hence this PR cannot get merged before October 17. I will do the merge once the period is over and there was no -1 vote.

PR #1303, opened 2025-01-06
New method ResourceInfo.getAnnotation()

### LINKED ISSUE #1292: Enhancement Proposal: ResourceInfo.getAnnotation(Class annotationClass)
I hereby propose the addition of the following method:

`public <A extends Annotation> A ResourceInfo.getResourceAnnotation(Class<A> annotationClass)`

The idea is that this method returns the annotation instance of the given `annotationClass` **with full respect to the rules found in spec chapter 3.6**. This effectively means, this method returns *either* the super class annotation, *or* the interface annotation, *or* the resource class annotation, *or* the resource method annotation, depending on whether *any* JAX-RS annotation is found in the particular *deeper* levels.

The benefit of this method is that there is no need for filters / interceptors to re-implement this algorithm again and again for custom annotations, in case the annotation author wants to apply the same rules as found in chapter 3.6 (hence: the new annotation shall "look and feel" like an official JAX-RS annotation). Implementations could make the already existing algorithm publicly available here, so it is guaranteed that the *exact same* treatment happens for custom annotations and for official annotations.

mkarg:
This PR is a **draft** for further discussion of `ResourceInfo.getAnnotation()`.

**It is not intended for code review, voting, or merge.** Please use it solely for discussion, unless we agreed upon the final API and have added TCK tests.

Closes #1292

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/container/ResourceInfo.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/container/ResourceInfo.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
@@ -16,6 +16,7 @@
 
 package jakarta.ws.rs.container;
 
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 
 /**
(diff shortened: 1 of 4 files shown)


mkarg:
Kindly requesting to continue our discussion *here*. :-)

In particular the question is: What to do with *custom* annotations like `@Tracing` (where possible the annotation author wants to treat it *as* a JAX-RS annotation) or `@RolesAllowed` (where there is a published spec mandating inheritance *always*, independent of JAX-RS annotation). What first came to my mind (and is found in this first draft) is to have an option for *custom* annotations: `JAX_RS` (treat it as it would be a JAX-RS annotation), `INHERIT` (performan inheritance *always*) or `NONE` (never inherit it).

WDYT?

chkal:
Shouldn't a JAX-RS API for retrieving the effective annotations for a given resource method always use the JAX-RS annotation inheritance rules? Just thinking out loudly. I'm not 100% myself sure yet.

mkarg:
> Shouldn't a JAX-RS API for retrieving the effective annotations for a given resource method always use the JAX-RS annotation inheritance rules? Just thinking out loudly. I'm not 100% myself sure yet.

You missed the point that the customization of the algorithm is for *custom* annotations **only**. For JAX-RS annotations certainly *always* the JAX-RS algorithm is used. This customization is needed as e. g. `@RolesAllowed` is commonly used with JAX-RS, but is *not* a JAX-RS annotation itself, and mandates a different "always-inherit" policy.

jamezp:
> > Shouldn't a JAX-RS API for retrieving the effective annotations for a given resource method always use the JAX-RS annotation inheritance rules? Just thinking out loudly. I'm not 100% myself sure yet.
> 
> You missed the point that the customization of the algorithm is for _custom_ annotations **only**. For JAX-RS annotations certainly _always_ the JAX-RS algorithm is used. This customization is needed as e. g. `@RolesAllowed` is commonly used with JAX-RS, but is _not_ a JAX-RS annotation itself, and mandates a different "always-inherit" policy.

I'm not too sure how I feel about that. The enum feels a bit clumsy, but I might not fully understand the use-case. While I can appreciate that `@RolesAllowed` is no inherited via the annotation, there's nothing really stopping a user from doing `ResourceInfo.getResourceAnnotation(RolesAllowed.class, CustomAnnotationsInheritancePolicy.JAX_RS)`. It feels out of scope for the Jakarta REST API.

Note, I'm not convinced one way or the other, this is just my initial take on it.

I also wonder if the return type should be something like `List<A>` for repeatable annotations if we're going to allow any annotation to be resolved. 

I'm also not to clear on the meaning on `CustomAnnotationsInheritancePolicy.INHERIT`. It states:
> Inherit custom annotations but do not treat them as JAX-RS annotation.

What does that mean? Is there an expectation that annotations on interfaces get included? Which annotation wins in a case where both an interface and implementation are annotated?

mkarg:
More comments please! :-)

PR #1313, opened 2025-03-31
added status code 422 - Unprocessable Entity

guitcastro:
This PR just fix a typo in doc `PARTIAL_CONTENT` docs and add the missing `422` status (Unprocessable Entity).

### DIFF
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
@@ -1183,7 +1183,7 @@ public abstract class Response implements AutoCloseable {
          */
         RESET_CONTENT(205, "Reset Content"),
         /**
-         * 206 Reset Content, see <a href="http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.7">HTTP/1.1
+         * 206 Partial Content, see <a href="http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.7">HTTP/1.1
          * documentation</a>.
          *
          * @since 2.0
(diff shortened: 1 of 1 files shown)


jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
```diff
@@ -1350,6 +1350,13 @@ public enum Status implements StatusType {
          * @since 2.1
          */
         PRECONDITION_REQUIRED(428, "Precondition Required"),
+        /**
+         * 422 Unprocessable Entity, see <a href="https://datatracker.ietf.org/doc/html/rfc4918#section-11.2">RFC 4918:
```
mkarg, line 1354: Please don't refer to WebDAV, but instead go with https://httpwg.org/specs/rfc9110.html#status.422.

guitcastro: @mkarg Thanks for the suggestion, I've updated the link

spericas (approved):

jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
```diff
@@ -1350,6 +1350,13 @@ public enum Status implements StatusType {
          * @since 2.1
          */
         PRECONDITION_REQUIRED(428, "Precondition Required"),
+        /**
+         * 422 Unprocessable Entity, see <a href="https://httpwg.org/specs/rfc9110.html#status.422">RFC 9110: 
+         * 422 Unprocessable Content</a>.
+         *
+         * @since 4.0
```
mkarg, line 1357: 4.0 is already released. We should have the next version here.

mkarg (changes requested):

jaxrs-api/src/main/java/jakarta/ws/rs/core/Response.java
guitcastro: @mkarg What version should I use?

mkarg: @spericas Forwarding question to project lead to trigger a final decision if we want a 4.1 or directly work towards 5.0.

mkarg: @jakartaee/ee4j-rest-committers I think the correct answer is 5.0, right?

jamezp: +1 5.0 ans 4.0 has shipped now.

mkarg: @guitcastro Please use "5.0"

PR #1331, opened 2025-10-15
[1328] Initial preparation for 5.0.0-M1 release

jamezp:
This is the initial preparation for an M1 release, see #1328.

The component upgrades are not required, but seemed useful. 

I enabled all the previously disabled tests and tested against RESTEasy and WildFly. All tests passed with the exception of 1 test, `ee.jakarta.tck.ws.rs.api.rs.core.uribuilder.JAXRSClientIT.replaceQueryTest4`. I've not looked into why, but we can look later..

The one point on #1328 I need a second set of eyes on is "Version update to the new version number including updating Export and Import versions in API jar’s MANIFEST.MF and Specification document". I think that is updated correctly, but if not please let me know and I'll fix it.

This did yet tackle the "Updates for the SEO change in the specification document as outlined in". I need to look at that.


pom.xml
```diff
@@ -15,26 +15,20 @@
...
     <parent>
         <groupId>org.eclipse.ee4j</groupId>
         <artifactId>project</artifactId>
-        <version>1.0.9</version>
+        <version>2.0.0-M1</version>
         <relativePath/>
     </parent>
 
-    <distributionManagement>
-        <snapshotRepository>
```
mkarg, line 32: We do not support snapshots anymore? 🤔

jersey-tck/pom.xml
```diff
@@ -35,8 +35,7 @@
     <description>This test verifies the compliance of Eclipse Jersey with Jakarta REST</description>
 
     <properties>
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.releas>21</maven.compiler.releas>
```
mkarg: Typo: `releas`

pom.xml
jamezp: We do, it's in the parent-pom now, https://github.com/eclipse-ee4j/ee4j/blob/93c96342bad91438b61a94fbc48979f620a13625/parent/pom.xml#L298.

jersey-tck/pom.xml
jamezp: Good catch, thank you!

spericas (approved):

asoldano (approved):

jim-krueger (approved):

jamezp:
It's been 2 weeks and 3 approvals so I'm going to merge this.

Issue #1215, opened 2024-02-07
Result of Application methods for CDI beans


jansupol:
Application getClasses() and getSingletons() return the classes and instances used by the application. Should they not be registered, they would not be used by the application.

However, in the CDI, should the providers be CDI beans, no matter whether they are registered or not, they are known to CDI.
Hence, getting all providers from CDI will include them.

The questions:

1. Should the CDI bean providers be included just when registered in the Application? Or is it ok to have the provider be a bean and it no longer needs to be registered (using the META-INF/services/jakarta.ws.rs.core.Feature for instance)? 
2. If so, does it matter that the behaviour could be potentially be different in the CDI-managed environment and non-cdi env using the non-cdi injection?

jamezp:
This is a good question and is relevant to both 3.2 and 4.0.

The `Application.getSingletons()` is deprecated which helps us out a bit. We can't really use CDI for anything returned from this method as it returns already instantiated objects.

For `Application.getClasses()`, I think those should be usable in that you could use the `BeanManager` to lookup those types.

With regards to the service loading, that's a great question. Without creating some kind of custom `ServiceLoader` which returns only class types, there really isn't a good way to make those providers a CDI bean. Therefore, IMO, we should state anything loaded from a service loader will not work with CDI. Not sure on the exact wording we should use though.

That said, from an implementation point of view, I could see creating a service loader which registers those as CDI beans in an CDI extension.

hantsy:
I think `Application` class can be deprecated, use CDI to scan/find beans directly.

jamezp:
It's currently the indicator to enable Jakarta REST for deployments. There are cases where CDI may not be enabled, but Jakarta REST is still wanted. It's a standalone specification that can work in servlet containers for example where CDI may not be available.

hantsy:
If switched to CDI `@Inject` and components(CDI scopes) instead of the `@Resource/@Context` for dependency injection in the next version, the none-CDI environment is meaningless.

Maybe programmatic APIs https://github.com/jakartaee/rest/issues/1301 with a Servlet Context initializer is more suitable in the Servlet env.

Issue #1214, opened 2024-01-26
Support injection via CDI to indicate a dependency on a Servlet-defined resource


jim-krueger:
The following Servlet-defined types are currently supported via `@Context` injection and will need to be supported via CDI going forward:

- ServletConfig
- ServletContext
- HttpServletRequest
- HttpServletResponse

See: https://jakarta.ee/specifications/restful-ws/3.1/jakarta-restful-ws-spec-3.1#servlet_container

jamezp:
IMO we shouldn't add anything specific here for the REST spec. In fact, we should probably deprecate these injection types. If servlet is available, the CDI spec requires servlet containers make these types injectable https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#additional_builtin_beans. The one exception here is the `ServletConfig`.

jim-krueger:
> The one exception here is the ServletConfig

Looks like jakarta.servlet.http.HttpServletResponse is another exception.
@Azquelt Does CDI have any equivalents for these?

jamezp:
Good catch. I would guess most implementations do allow it to be injected, but you're correct that it's not listed.

jansupol:
@arjantijms Can you, please, shed light on what CDI beans are provided by Servlet 6.1 (Remember we discussed some changes) and what provides the Glassfish Servlet impl?

Azquelt:
> Looks like jakarta.servlet.http.HttpServletResponse is another exception. @Azquelt Does CDI have any equivalents for these?

No, the only beans I'm aware of provided by CDI are in the [spec section James linked before](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#additional_builtin_beans). However, note that these requirements will [move to the web profile spec][move-reqs] for EE 11, so the REST spec may want to restate them anyway.

If the REST implementation already has a way to get hold of these objects, I would have thought that making them available for injection should be fairly straight forward with a producer method.

[move-reqs]: https://github.com/jakartaee/cdi/issues/687#issuecomment-1921973263

jamezp:
If we do include those as required in the spec, we should be specific they are only available in a servlet environment. The Jakarta EE Core Profile does not include servlet.

jansupol:
As well as in the CDI-managed environment

arjantijms:
[...]

Issue #1209, opened 2024-01-24
Deprecate @Context injection and @Suspended for Jakarta Rest 3.2


jim-krueger:
`@Context` injection and @Suspended will be deprecated in Jakarta Rest 3.2

jansupol:
When I voted my +1 for the Jakarta REST 3.2, I did so with an expectation of 3.2 being less work than 4.0, i.e. the 3.1 (proprietary) injection being mandatory, while CDI injection being optional. I hope we agree on this, otherwise it makes more sense to have the release version 4.0.

For this, I would propose not to `@Deprecate` the `@Context`, but rather make it deprecated in the Spec document and Javadoc only.

jamezp:
The `@Context` annotation is already deprecated in the spec document and the Javadoc. IMO users will not see that and it's going to catch users by surprise that it's gone.

From what I understand there are two issues:

1. You can't remove things in a minor release, so we can't remove `@Context`.
2. If you deprecate something, it needs a replacement.

On point 2 what is not clear to me is how you can deprecate something in documentation and not have a replacement, but if you add the `@Deprecated` annotation you must provided a replacement.

That said, I really don't see an issue with requiring CDI in a container. I understand that will be an issue on the client side. However, that's still going to be an issue whether we deprecate the annotation or remove it.

mkarg:
The problem ist that software does exists that relies on the fact that 3.x is able to be run *without* CDI. It is not a good idea to introduce the enforcement for CDI in the 3.x line. Downstreams expect this enforcement in 4.x only.

jamezp:
I guess that's what I meant by "requiring CDI in a container". We have this issue in 4.x really as well with the client side and defining whether or not it should be required in `SeBootstrap`. Maybe I misunderstood, but I thought those two were still an open question.

The current spec, 3.1, already requires CDI if running in a CDI container. I do understand that doesn't address all cases like a servlet container without CDI or `SeBootstrap`. I guess, and I could be wrong, I don't see it as a far stretch to require it, but I also understand why there would be an objection to it.

That said, I still think it's a really useful and good idea to deprecate `@Context` via the `@Deprecated` annotation to make it very clear that it's going away.

jansupol:
> "requiring CDI in a container"

This basically means that in the app server, we need to support the CDI injection in the container, whereas the customer still wants to use `@Context`. So in the container, we:

[...]

PR #1340, opened 2026-03-30
Introduce CDI requirements for deprecated `@Context` injection 

jamezp:
This PR introduces the Jakarta Dependency and Context Injection requirements as a replacement for `@Context` injection. Note that most of this was taken from the original [JakartaRest40.pdf](https://github.com/user-attachments/files/26354537/JakartaRest40.pdf). There are a few things I did differently however. 

One is I changed the entity parameter annotation from `@Entity` to `@EntityParam`. One reason is there is already a `jakarta.ws.rs.client.Entity` type and that seemed confusing to me. The other is Jakarta Persistence already has a `@Entity` parameter. Using `@EntityParam` seemed the most natural with other parameter annotations. However, I'm open to suggestions like `@MessageBody` or anything else.

The second is I do think we should required the `@Inject` annotation on methods we want to use CDI injection on. The reason is implementations will need a way to know if we should be using CDI injection or the old way of parameter injection. For example consider a method like:
```java
@GET
@Path("example")
public Response performGet(final MyCdiBean bean) {
    return Response.ok(bean.get()).build();
}
```

We have no way to know if `MyCdiBean` is an entity or it should be handled by the CDI container.

I'm opening this as a draft for now for further discussions. There are more tests that are needed for the `@EntityParam` and we may need to copy some other tests to use CDI instead of `@Context` injection. I didn't want to spend too much time on that until we agreed on the rest of the PR though.

This PR relates to:

- #569 
- #1209 
- #1221 

We need to consider what needs to be done for #1214 and #1215 as well. I did not add any specification documentation about those as I wasn't sure where we wanted to go. 

Note I added a OpenRewrite recipe here for adding the `@EntityParam` annotation. However, this might be the wrong place for this. I'd initially added it here to do some testing. It needs some updating though as I changed the requirements a bit. I've left it in the draft for now, but we can remove it before we merge.

### DIFF
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -86,6 +86,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
--- a/jaxrs-api/pom.xml
+++ b/jaxrs-api/pom.xml
@@ -86,6 +86,14 @@
 
     <dependencies>
         <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.inject</groupId>
+            <artifactId>jakarta.inject-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/container/Suspended.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/container/Suspended.java
@@ -76,9 +76,11 @@ import java.util.concurrent.TimeUnit;
  *
  * @author Marek Potociar
  * @since 2.0
+ * @deprecated use Jakarta Context and Dependency injection
  */
 @Target({ ElementType.PARAMETER })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
+@Deprecated(forRemoval = true, since = "5.0")
 public @interface Suspended {
 }
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/Configurable.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/Configurable.java
@@ -261,7 +261,7 @@ public interface Configurable<C extends Configurable> {
      * As opposed to components registered via {@link #register(Class)} method, the lifecycle of providers registered using
      * this instance-based {@code register(...)} is not managed by JAX-RS runtime. The same registered component instance is
      * used during the whole lifespan of the configurable context. Fields and properties of all registered JAX-RS component
-     * instances are injected with their declared dependencies (see {@link Context}) by the JAX-RS runtime prior to use.
+     * instances are injected with their declared dependencies by CDI prior to use.
      * </p>
      *
      * @param component JAX-RS component instance to be configured in the scope of this configurable context.
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/Configurable.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/Configurable.java
@@ -261,7 +261,7 @@ public interface Configurable<C extends Configurable> {
      * As opposed to components registered via {@link #register(Class)} method, the lifecycle of providers registered using
      * this instance-based {@code register(...)} is not managed by JAX-RS runtime. The same registered component instance is
      * used during the whole lifespan of the configurable context. Fields and properties of all registered JAX-RS component
-     * instances are injected with their declared dependencies (see {@link Context}) by the JAX-RS runtime prior to use.
+     * instances are injected with their declared dependencies by the Jakarta REST runtime prior to use.
      * </p>
      *
      * @param component JAX-RS component instance to be configured in the scope of this configurable context.
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/core/HttpHeaders.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/core/HttpHeaders.java
@@ -28,7 +28,6 @@ import java.util.function.Predicate;
  *
  * @author Paul Sandoz
  * @author Marc Hadley
- * @see Context
  * @since 1.0
  */
 public interface HttpHeaders {
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/ext/Provider.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/ext/Provider.java
@@ -16,11 +16,16 @@
 
 package jakarta.ws.rs.ext;
 
+import java.io.Serial;
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Stereotype;
+import jakarta.enterprise.util.AnnotationLiteral;
 
 /**
  * Marks an implementation of an extension interface that should be discoverable by JAX-RS runtime during a provider
@@ -33,5 +38,25 @@ import java.lang.annotation.Target;
 @Target({ ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
+@ApplicationScoped
+@Stereotype
 public @interface Provider {
+
+    /**
+     * Supports inline instantiation of the {@link Provider} annotation.
+     *
+     * @since 5.0
+     */
+    final class Literal extends AnnotationLiteral<Provider> implements Provider {
+        @Serial
+        private static final long serialVersionUID = 1L;
+
+        /**
+         * Default Provider literal
+         */
+        public static final Provider INSTANCE = new Literal();
+
+        private Literal() {
+        }
+    }
 }
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-4.0-release.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-4.0-release.adoc
@@ -1,0 +1,35 @@
+////
+*******************************************************************
+* Copyright (c) 2026 Eclipse Foundation
+*
+* This specification document is made available under the terms
+* of the Eclipse Foundation Specification License v1.0, which is
+* available at https://www.eclipse.org/legal/efsl.php.
+*******************************************************************
+////
+
+[[changes-since-4.0-release]]
+=== Changes Since 4.0 Release
+
+* <<cdi>>: Required CDI 5.0 or later for proper integration of Jakarta REST and CDI injection (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi-injection-model>>: Documented comprehensive CDI injection model for resources and providers (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi>>: Updated `@ApplicationPath`, `@Path`, and `@Provider` annotations to be CDI stereotype
+annotations with default scopes (`@ApplicationScoped` for `@ApplicationPath` and `@Provider`,
+`@RequestScoped` for `@Path`) (https://github.com/jakartaee/rest/issues/556[#556], https://github.com/jakartaee/rest/issues/952[#952])
+* <<cdi-injection-model>>: Parameter extraction annotations (`@QueryParam`, `@PathParam`,
+`@HeaderParam`, `@MatrixParam`, `@CookieParam`, `@FormParam`) now act as CDI qualifiers (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi-injection-model>>: Fields in CDI-managed beans must use `@Inject` instead of the
+deprecated `@Context` annotation (https://github.com/jakartaee/rest/issues/569[#569], https://github.com/jakartaee/rest/issues/1209[#1209])
+* <<cdi-injection-model>>: Resource method parameters support only Jakarta REST context types,
+parameter extraction annotations, entity parameters, `AsyncResponse`, and `SseEventSink`;
+arbitrary CDI beans must be injected via fields or constructors (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi-injection-model>>: Clarified that `AsyncResponse` and `SseEventSink` can only be
+used as resource method parameters, not as injectable fields (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi-producers>>: Required CDI producers for Jakarta REST context types and parameter
+extraction annotations (https://github.com/jakartaee/rest/issues/569[#569])
+* Deprecated `@Context` annotation for removal in a future release (https://github.com/jakartaee/rest/issues/569[#569], https://github.com/jakartaee/rest/issues/1209[#1209])
+* Deprecated `@Suspended` annotation in 5.0 for removal in a future release; use type-alone approach
+for `AsyncResponse` in CDI environments (https://github.com/jakartaee/rest/issues/1209[#1209])
+* <<servlet_container>>: Deprecated `@Context` injection for Servlet-defined types; use CDI
+injection (`@Inject`) as specified by Jakarta Servlet and Jakarta EE Platform specifications (https://github.com/jakartaee/rest/issues/1214[#1214])
+* Removed Additional Requirements section; CDI specification defines bean lifecycle behaviors (https://github.com/jakartaee/rest/issues/1215[#1215])
--- a/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-4.0-release.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-4.0-release.adoc
@@ -12,3 +12,25 @@
 === Changes Since 4.0 Release
 
 * <<standard_entity_providers>>: Added entity provider for `java.nio.file.Path`
+* <<cdi>>: Required CDI 5.0 or later for proper integration of Jakarta REST and CDI injection (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi-injection-model>>: Documented comprehensive CDI injection model for resources and providers (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi>>: Updated `@ApplicationPath`, `@Path`, and `@Provider` annotations to be CDI stereotype
+annotations with default scopes (`@ApplicationScoped` for `@ApplicationPath` and `@Provider`,
+`@RequestScoped` for `@Path`) (https://github.com/jakartaee/rest/issues/556[#556], https://github.com/jakartaee/rest/issues/952[#952])
+* <<cdi-injection-model>>: Parameter extraction annotations (`@QueryParam`, `@PathParam`,
+`@HeaderParam`, `@MatrixParam`, `@CookieParam`, `@FormParam`) now act as CDI qualifiers (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi-injection-model>>: Fields in CDI-managed beans must use `@Inject` instead of the
+deprecated `@Context` annotation (https://github.com/jakartaee/rest/issues/569[#569], https://github.com/jakartaee/rest/issues/1209[#1209])
+* <<cdi-injection-model>>: Resource method parameters support only Jakarta REST context types,
+parameter extraction annotations, entity parameters, `AsyncResponse`, and `SseEventSink`;
+arbitrary CDI beans must be injected via fields or constructors (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi-injection-model>>: Clarified that `AsyncResponse` and `SseEventSink` can only be
+used as resource method parameters, not as injectable fields (https://github.com/jakartaee/rest/issues/569[#569])
+* <<cdi-producers>>: Required CDI producers for Jakarta REST context types and parameter
+extraction annotations (https://github.com/jakartaee/rest/issues/569[#569])
+* Deprecated `@Context` annotation for removal in a future release (https://github.com/jakartaee/rest/issues/569[#569], https://github.com/jakartaee/rest/issues/1209[#1209])
+* Deprecated `@Suspended` annotation in 5.0 for removal in a future release; use type-alone approach
+for `AsyncResponse` in CDI environments (https://github.com/jakartaee/rest/issues/1209[#1209])
+* <<servlet_container>>: Deprecated `@Context` injection for Servlet-defined types; use CDI
+injection (`@Inject`) as specified by Jakarta Servlet and Jakarta EE Platform specifications (https://github.com/jakartaee/rest/issues/1214[#1214])
+* Scoped Additional Requirements section to non-CDI environments; CDI specification defines bean lifecycle behaviors for CDI-managed beans (https://github.com/jakartaee/rest/issues/1215[#1215])
--- a/jaxrs-spec/src/main/asciidoc/chapters/context/_contexttypes.adoc
+++ b/jaxrs-spec/src/main/asciidoc/chapters/context/_contexttypes.adoc
@@ -17,12 +17,22 @@ subclasses (server only). Except for `Configuration` and `Providers`,
 which are injectable in both client and server-side providers, all the
 other types are server-side only.
 
+[NOTE]
+====
+In CDI-enabled environments, these context types MUST be made available
+for injection via CDI producers as specified in <<cdi-producers>>. The
+use of `@Context` annotation is deprecated in favor of CDI injection using
+`@Inject` (for fields and constructors) or type-based injection (for
+resource method parameters). See <<cdi-injection-model>> for details on
+the CDI injection model for Jakarta REST.
+====
+
 [[application]]
 ==== Application
 
 The instance of the application-supplied `Application` subclass can be
-injected into a class field or method parameter using the
-`@Context` annotation. Access to the `Application` subclass instance
+injected into a class field using the `@Inject` annotation or as a resource
+method parameter. Access to the `Application` subclass instance
 allows configuration information to be centralized in that class. Note
 that this cannot be injected into the `Application` subclass itself
 since this would create a circular dependency.
@@ -30,9 +40,9 @@ since this would create a circular dependency.
 [[uris-and-uri-templates]]
 ==== URIs and URI Templates
 
-An instance of `UriInfo` can be injected into a class field or method
-parameter using the `@Context` annotation. `UriInfo` provides both
-static and dynamic, per-request information, about the components of a
+An instance of `UriInfo` can be injected into a class field using the
+`@Inject` annotation or as a resource method parameter.
+`UriInfo` provides both static and dynamic, per-request information, about the components of a
 request URI. E.g. the following would return the names of any query
 parameters in a request:
 
@@ -40,11 +50,29 @@ parameters in a request:
 ----
 @GET
 @Produces("text/plain")
-public String listQueryParamNames(@Context UriInfo info) {
+public String listQueryParamNames(UriInfo info) {
     StringBuilder buf = new StringBuilder();
     for (String param: info.getQueryParameters().keySet()) {
         buf.append(param);
-        buf.append("\n");
+        buf.append('\n');
+    }
+    return buf.toString();
+}
+----
+
+[source,java]
+----
+
+@Inject
+UriInfo info;
+
+@GET
+@Produces("text/plain")
+public String listQueryParamNames() {
+    StringBuilder buf = new StringBuilder();
+    for (String param: info.getQueryParameters().keySet()) {
+        buf.append(param);
+        buf.append('\n');
     }
     return buf.toString();
 }
@@ -56,8 +84,8 @@ information following the pre-processing described in <<reqpreproc>>.
 [[httpheaders]]
 ==== Headers
 
-An instance of `HttpHeaders` can be injected into a class field or
-method parameter using the `@Context` annotation. `HttpHeaders` provides
+An instance of `HttpHeaders` can be injected into a class field using the
+`@Inject` annotation or as a resource method parameter. `HttpHeaders` provides
 access to request header information either in map form or via strongly
 typed convenience methods. E.g. the following would return the names of
 all the headers in a request:
@@ -66,11 +94,28 @@ all the headers in a request:
 ----
 @GET
 @Produces("text/plain")
-public String listHeaderNames(@Context HttpHeaders headers) {
+public String listHeaderNames(HttpHeaders headers) {
     StringBuilder buf = new StringBuilder();
     for (String header: headers.getRequestHeaders().keySet()) {
         buf.append(header);
-        buf.append("\n");
+        buf.append('\n');
+    }
+    return buf.toString();
+}
+----
+
+[source,java]
+----
+@Inject
+HttpHeaders headers;
+
+@GET
+@Produces("text/plain")
+public String listHeaderNames() {
+    StringBuilder buf = new StringBuilder();
+    for (String header: headers.getRequestHeaders().keySet()) {
+        buf.append(header);
+        buf.append('\n');
     }
     return buf.toString();
 }
@@ -86,8 +131,9 @@ Response headers may be provided using the `Response` class, see
 ==== Content Negotiation and Preconditions
 
 JAX-RS simplifies support for content negotiation and preconditions
-using the `Request` interface. An instance of `Request` can be injected
-into a class field or method parameter using the `@Context` annotation.
+using the `Request` interface. An instance of `Request` can be injected
+into a class field using the `@Inject` annotation or as a resource method
+parameter.
 The methods of `Request` allow a caller to determine the best matching
 representation variant and to evaluate whether the current state of the
 resource matches any preconditions in the request. Precondition support
@@ -99,7 +145,23 @@ the request before updating the resource:
 [source,java]
 ----
 @PUT
-public Response updateFoo(@Context Request request, Foo foo) {
+public Response updateFoo(Request request, Foo foo) {
+    EntityTag tag = getCurrentTag();
+    ResponseBuilder responseBuilder = request.evaluatePreconditions(tag);
+    if (responseBuilder != null)
+        return responseBuilder.build();
+    else
+        return doUpdate(foo);
+}
+----
+
+[source,java]
+----
+@Inject
+Request request;
+
+@PUT
+public Response updateFoo(Foo foo) {
     EntityTag tag = getCurrentTag();
     ResponseBuilder responseBuilder = request.evaluatePreconditions(tag);
     if (responseBuilder != null)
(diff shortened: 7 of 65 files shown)


FroMage:
I would like to suggest that the choice of calling this an _entity_ in the first place was not necessarily helpful.

It was called an _entity_ or _message body_ in the [first HTTP 1.1 RFC](https://datatracker.ietf.org/doc/html/rfc2616#section-7) but a _resource_ in [REST](https://en.wikipedia.org/wiki/REST) and in the [more recent HTTP RFC](https://datatracker.ietf.org/doc/html/rfc9110#name-terminology-and-core-concep) it is called a _resource_, a _representation data_ or a _payload_ or _message payload_ or _message body_.

A lot of other frameworks, libraries or programming languages also call it a _request body_ (as opposed to a response version).

I think we should consider a better name than _entity_ for this annotation, and not only because it clashes with JPA's `@Entity` annotation.

Do we expect this annotation to be used to designate the response _entity_ too?

jamezp:
> I would like to suggest that the choice of calling this an _entity_ in the first place was not necessarily helpful.
> 
> It was called an _entity_ or _message body_ in the [first HTTP 1.1 RFC](https://datatracker.ietf.org/doc/html/rfc2616#section-7) but a _resource_ in [REST](https://en.wikipedia.org/wiki/REST) and in the [more recent HTTP RFC](https://datatracker.ietf.org/doc/html/rfc9110#name-terminology-and-core-concep) it is called a _resource_, a _representation data_ or a _payload_ or _message payload_ or _message body_.
> 
> A lot of other frameworks, libraries or programming languages also call it a _request body_ (as opposed to a response version).

Yeah, `@RequestBody` was another one I thought of. I only went with `@EntityParam` because it kind of "fit" with the other `@*Param` annotations. However, this is different and I honestly don't love the name. I just wanted to push something up to start the conversion so thank you :)

> 
> I think we should consider a better name than _entity_ for this annotation, and not only because it clashes with JPA's `@Entity` annotation.
> 
> Do we expect this annotation to be used to designate the response _entity_ too?

No, just on a request which is why `@RequestyBody`, `@MessageBody`, something like that makes more sense to me personally.

FroMage:
I think, but this may be just me, I don't have data to back this up, but I think the _entity_ term is confusing for new users. In a way that _request body_ is not: I feel that most people understand what this is. I have some doubts about _message body_ because I suspect fewer people associate HTTP with messages.

FroMage:
But on the other hand, we already have `Entity` in several type names in the spec 😥

jamezp:
Agreed for sure. IMO `@Entity` so ambiguous and somewhat confusing. I think my personal preference would be `@RequestBody` because that is exactly what it is.

mkarg:
My personal preference would be `@Entity` because it is the word found in all the RFCs, so people dealing with specs will search for exactly that. It is not our job to criticize the IETF. It is our job to introduce CDI, solely.

WhiteCat22:
I am leaning towards `@Entity`, although I hate the potential collision with JPA.

My reasoning is that we already have `jakarta.ws.rs.client.Entity` in the `Client` and it might be confusing to users if the client and server don't match.

```
.post(Entity.entity(message, MediaType.TEXT_PLAIN));

...

public Response updateFoo(@Entity Foo foo) {
...
}
```

If we didn't already use `Entity` in the client I'd be more inclined towards `@RequestBody` with a "definition of terms" section in the REST spec to map `@RequestBody` to HTTP's `Entity`. However, since `Entity` is already well established in the REST spec I think we should continue to use it for consistency.

jamezp:
Another option is to simply not allow CDI for resource method parameters. Then we essentially have what we have now which is no annotation on the entity. The other types that require `@Context` can just use inference to be resolved. How the `@*Param` parameters are resolved becomes an implementation detail.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
```diff
@@ -66,8 +70,306 @@ public class CdiBeanResource {
...
+        this.service = service;
+        this.uriInfo = uriInfo;
+        this.category = category;
+    }
+}
+----
+
+NOTE: Jakarta REST parameter annotations (`@QueryParam`, `@PathParam`, etc.) in
+constructors depend on request context. Using these in application-scoped beans
+may result in undefined behavior since the constructor is invoked once at application
```
Ladicek, line 163: "May" here is an understatement. The `@*Param` beans are `@Dependent`, so the producer will be called immediately and will try to obtain the value from the REST request context, but there shall be none. (Except perhaps in subresource locators?) This _will_ result in undefined behavior. Well, it will most likely result in an exception :-)

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
```diff
@@ -66,8 +70,306 @@ public class CdiBeanResource {
...
+startup, not per-request.
+
+====== Resource Method Parameter Injection
+
+Resource methods which use parameter inject MUST be annotated with with `@Inject`.
```
Ladicek, line 168: Please _don't do this_. A method that is annotated `@Inject` is a so called _initializer method_, which is a fancy term for _setter injection_. See https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#initializer_methods

It feels like parameters of resource methods declared in resources that are CDI beans should always be looked up from the CDI container, so I don't think you need _any_ special designation. But if you feel like you do need one, it feels best to mandate a qualifier on the parameter. The qualifier is either the `@*Param` annotation, or some specific qualifier the user knows about, or `@Default` (which is a qualifier all beans automatically get when they don't have any other qualifier defined; see https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#builtin_qualifiers for precise definition).

FroMage:
> My personal preference would be `@Entity` because it is the word found in all the RFCs, so people dealing with specs will search for exactly that. It is not our job to criticize the IETF. It is our job to introduce CDI, solely.

As I said in my comment, the latest RFC removed _entity_ entirely, it's not called that anymore: https://datatracker.ietf.org/doc/html/rfc9110

For some reason, they did keep _entity tag_ but that's probably an oversight.

I am not criticising IETF, I'm pointing out that nobody ever called these entities outside of Jakarta REST, and the latest RFC appears to agree with this observation.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: Okay. I had it as "will" and thought maybe that was too strong :) Maybe I'll reword it in general that using those types as fields in a resource that is `@ApplicationScoped` will result in defined behavior and is not suggested.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: Ah, okay. That makes sense. I'd originally had it not required, but then I was trying to figure out how an implementation would support both the old way and the new way of processing resource methods and couldn't think of a good way.

IoW my concern was something like:
```java
public Response post(final Customer customer) {
}
```

Would now be required to be changed to:
```java
public Response post(@EntityParam final Customer customer) {
}
```

However, I definitely see what you mean and I'll have to think about how we want to handle the migration path.

jamezp:
It's seems like RFC9110 uses "content" so we could use `@Content` https://www.rfc-editor.org/rfc/rfc9110.html#name-content

> HTTP messages often transfer a complete or partial representation as the message "content": a stream of octets sent after the header section, as delineated by the message framing.

If we really want to stick with something around entity I'd prefer `@EntityParam` to not be confused with or require a FQCN reference, the `jakarta.ws.rs.client.Entity`.

FroMage:
> It's seems like RFC9110 uses "content" so we could use `@Content` https://www.rfc-editor.org/rfc/rfc9110.html#name-content

Huh, I did not notice _content_ when I listed all the other terms it uses. That one makes a lot of sense, given content types, content negociation…

Probably not as widespread as _request body_ for a term, but definitely a lot more obvious to anyone than _entity_.

mkarg:
> It's seems like RFC9110 uses "content" so we could use `@Content` https://www.rfc-editor.org/rfc/rfc9110.html#name-content
> 
> > HTTP messages often transfer a complete or partial representation as the message "content": a stream of octets sent after the header section, as delineated by the message framing.
> 
> If we really want to stick with something around entity I'd prefer `@EntityParam` to not be confused with or require a FQCN reference, the `jakarta.ws.rs.client.Entity`.

`@EntityParam` seems to be straightforward, as it builds on the "Entity" word used in other parts of Jakarta REST, and extends the `Param` series of injection markers. +1 from me.

mkarg:
> > It's seems like RFC9110 uses "content" so we could use `@Content` https://www.rfc-editor.org/rfc/rfc9110.html#name-content
> 
> Huh, I did not notice _content_ when I listed all the other terms it uses. That one makes a lot of sense, given content types, content negociation…
> 
> Probably not as widespread as _request body_ for a term, but definitely a lot more obvious to anyone than _entity_.

IMHO `@Content` is far too easy to confuse with `@Context`.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: We could donate a rule for OpenRewrite.

jamezp: Yeah, that makes sense to me as well. I don't know if they'll accept it without a release, but that seems okay as well. I'm definitely not convinced it needs to be part of the spec, I just didn't want to lose it :)

jamezp:
I did a short talk about these changes today in this PR on the Jakarta EE Futures call. @jhanders34 had a, very good IMO, idea to not allow arbitrary CDI beans to be injected into resource methods which means we would not need the `@EntityParam` annotation for 5.0. Then in 6.0 when we remove `@Context`, we can add the `@EntityParam` annotation and then allow CDI bean injection into resource methods.

Without this approach we could break compatibility and require users to add `@EntityParam` to request body content parameters.

mkarg:
@jamezp I need to confess I somewhat got lost in the discussion. Could you please elaborate how that distinction in 5.0 and 6.0 changes anything? I am no CDI expect, so please bear with me. What problem arises from our current plan and why is that "very good IMO" idea the soltuion? 🤔

jaxrs-api/src/main/java/jakarta/ws/rs/container/Suspended.java
```diff
@@ -76,9 +76,11 @@
  *
  * @author Marek Potociar
  * @since 2.0
+ * @deprecated use Jakarta Context and Dependency injection
```
mkarg, line 79: Maybe we should add `since 5.0` to the JavaDocs also, so people have a better understanding which version supports what technology?

jaxrs-api/src/main/java/jakarta/ws/rs/core/Configurable.java
```diff
@@ -261,7 +261,7 @@ public interface Configurable<C extends Configurable> {
      * As opposed to components registered via {@link #register(Class)} method, the lifecycle of providers registered using
      * this instance-based {@code register(...)} is not managed by JAX-RS runtime. The same registered component instance is
      * used during the whole lifespan of the configurable context. Fields and properties of all registered JAX-RS component
-     * instances are injected with their declared dependencies (see {@link Context}) by the JAX-RS runtime prior to use.
+     * instances are injected with their declared dependencies by the CDI prior to use.
```
mkarg: As CDI is an abstract idea but not a concrete product, shouldn't it be "...by CDI prio..." (without "the")?

jaxrs-api/src/main/java/jakarta/ws/rs/core/HttpHeaders.java
```diff
@@ -28,7 +28,6 @@
  *
  * @author Paul Sandoz
  * @author Marc Hadley
- * @see Context
```
mkarg, line 31: Maybe we need another cross reference mentioned, to make clear that these interfaces *are* injectable?

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Provider.java
```diff
@@ -33,5 +36,7 @@
 @Target({ ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
+@ApplicationScoped
```
mkarg, line 41: Actually I wonder if providers can only be application scoped, or if it was valid to make them request scoped before?

jaxrs-api/src/main/java/jakarta/ws/rs/EntityParam.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @author <a href="mailto:jperkins@ibm.com">James R. Perkins</a>
+ * @see jakarta.ws.rs.ext.MessageBodyReader
+ * @see jakarta.ws.rs.Consumes
+ * @since 5.0
+ */
+@Target({ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface EntityParam {
```
mkarg: 👍

jaxrs-api/pom.xml
```diff
@@ -85,6 +85,10 @@
     </build>
 
     <dependencies>
+        <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
+        </dependency>
```
mkarg, line 91: Shouldn't it be "provided"?

jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-4.0-release.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+[[changes-since-4.0-release]]
+=== Changes Since 4.0 Release
+
+* <<cdi>>: Required CDI 5.0 or later for proper integration of Jakarta REST and CDI injection
+* <<cdi-injection-model>>: Documented comprehensive CDI injection model for resources and providers
+* <<cdi>>: Updated `@ApplicationPath`, `@Path`, and `@Provider` annotations to be CDI stereotype
+annotations with default scopes (`@ApplicationScoped` for `@ApplicationPath` and `@Provider`,
+`@RequestScoped` for `@Path`)
+* <<cdi-injection-model>>: Parameter extraction annotations (`@QueryParam`, `@PathParam`,
```
mkarg, line 20: Do we need to list `@EntityParam` here?

jaxrs-spec/src/main/asciidoc/chapters/context/_contexttypes.adoc
```diff
@@ -151,8 +248,8 @@ modifications:
 ----
 @Path("widgets")
 public class WidgetsResource {
-    @Context
-    private ResourceContext rc;
+    @Inject
+    ResourceContext rc;
```
mkarg: Is @Inject not possible with `private` in CDI?

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
```diff
@@ -66,8 +70,316 @@ public class CdiBeanResource {
...
 provided in CDI by using another bean, of type `MyOtherCdiBean`, in
-order to return a resource representation. See <<additional_reqs>>
-for additional requirements on CDI-style Beans.
+order to return a resource representation. See <<cdi-injection-model>>
+for details on the CDI injection model.
+
+[[cdi-injection-model]]
+===== Injection Model for CDI-Managed Resources
+
+When Jakarta REST resources and providers are managed as CDI beans, a hybrid
```
mkarg, line 79: "When"? Isn't that mandatory now?

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
```diff
@@ -66,8 +70,316 @@ public class CdiBeanResource {
...
+----
+
+NOTE: Jakarta REST parameter annotations (`@QueryParam`, `@PathParam`, etc.) in
+constructors depend on request context. Using these in application-scoped beans
+will result in undefined behavior since the constructor is invoked once at application
+startup, not per-request.
+
+====== Resource Method Parameter Injection
+
+Resource methods allow for parameter injection via CDI. The method MUST not be annotated
```
mkarg: >The method MUST not be annotated with `@Inject`

Why should someone do that? Do we really have to explicitly say that?

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
```diff
@@ -66,8 +70,316 @@ public class CdiBeanResource {
...
+  `Sse`, and `UriInfo`. These types are identified by type alone. Implementations MUST
+  provide CDI producers as specified in <<cdi-producers>>.
+
+* **Application-provided CDI beans**: Any CDI-managed bean can be injected as
+  a resource method parameter by type. CDI handles injection of application beans
+  normally.
+
+* **Jakarta REST-managed special parameters**: The following parameters are NOT
+  CDI beans and are handled directly by the Jakarta REST implementation:
+  ** **Entity parameter** (`@EntityParam`): The request entity body, deserialized using
```
mkarg: Is there a technical reason why the entity should *not* be a CDI bean?

jamezp:
@mkarg That is a fair request and would help others too :) I'll work on something more detailed next week and work on updating this PR to reflect the plan.

## REST 5.0

Add CDI full CDI support for resources, providers and applications for field injection and constructor injection. For resource methods, CDI will only replace the usage of `@Context` injection. Meaning you cannot inject an arbitrary CDI bean into a resource method.

Deprecate `@Context`, `@Suspended` annotations and the `ContextResolver` interface. We may want to deprecate the `Providers.getContextResolver()` too since we're deprecating the `ContextResolver`. However, maybe we don't need to deprecate the latter two. I'm open to either option.

Full backward compatibility, meaning `@Context` injection will still work as will the other API's. There is no need to add the `@EntityParam` annotation as the request body parameter will still be the "unknown" parameter. 

## REST 6.0

Remove the deprecated API's and add the `@EntityParam` (or whatever we decide to call it) and require it on the message body parameter. This will allow us to support full CDI injection for resource methods.

mkarg:
Thank you, James. As in the end the result will be the same (everbody needs to overhaul his existing applications and put `@EntityParam` on each single resource method), I do not see the actual benefit of this two-step way. Developers will get confused by a long and ongoing change path. Why not simply doing that cut in 5.0 directly?

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
```diff
@@ -66,8 +70,316 @@ public class CdiBeanResource {
...
+
+These producers MUST be `@Dependent` scoped and could use the CDI `InjectionPoint`
+API to:
+
+1. Inspect the qualifier annotation to determine the parameter name
+2. Extract the value from the current request context
+3. Convert the value to the target type using the same conversion rules as
+   specified in <<resource_field>>
+
+NOTE: The `@EntityParam` annotation does NOT require a CDI producer. Entity parameters
```
mkarg: I do not understand this separation. Again, I am not a CDI expert, so bear with me. IMHO @EntityParam should be provided by a CDI producer. The runtime should be that CDI producer. To produce the instance, the runtime invokes MBRs. Wouldn't that be straightforward? 🤔

jaxrs-spec/src/main/asciidoc/chapters/providers/_contextprovider.adoc
```diff
@@ -40,3 +47,79 @@ providers according to the media types they declare support for. Sorting
 of media types follows the general rule: x/y latexmath:[$<$] x/*
 latexmath:[$<$] \*/*, i.e. a provider that explicitly lists a media type
 is sorted before a provider that lists \*/*.
+
+[[contextprovider-migration]]
+==== Migration to CDI Producers
+
+Applications using `ContextResolver` should migrate to using CDI producers
```
mkarg: I do not see why we need to deprecate "ContextResolver". Couldn't we instead say that runtime MUST provide CDI producers for these classes, and MUST internally invoke the original context resolver in turn?

jamezp: I've been wondering about this one as well. The `ContextResolver` likely doesn't need to be deprecated, so maybe we just don't deprecate which seems fine to me.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: The part of requiring `@Inject` for resource methods has been removed and we definitely shouldn't do that. My initial thought when I added it was that it would be easy for implementations to know if it was a CDI resource method or a legacy one.

jaxrs-api/pom.xml
jamezp: I can go either way. It's required at runtime and Maven doesn't provide the best scopes for this. However, I've always been of the opinion that an API should transitively bring in the dependencies it needs. Users can always override them.

jaxrs-spec/src/main/asciidoc/chapters/context/_contexttypes.adoc
jamezp: You know, I swear at one point it was not allowed, but Weld did allow it. However, re-reading the spec I think I'm wrong :) I'll change that back.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: We haven't defined what needs to happen with `SeBoostrap` which is why I added this.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Provider.java
jamezp: I don't see why they couldn't be request scoped. I probably wouldn't suggest it, but I don't see why it would be an issue :)

jaxrs-api/src/main/java/jakarta/ws/rs/container/Suspended.java
jamezp: We have the version in the `@Deprecated` annotation. We could add it here too I guess.

jaxrs-spec/src/main/asciidoc/chapters/appendix/_changes-since-4.0-release.adoc
jamezp: If we determine that `@EntityParam` really is a CDI qualifier, yes. I'm not sure it can be. I need to think through how this would work.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: I don't really see how it could be. You need the request to create the entity. If it was a CDI bean it would mean it could be injected into anything which doesn't have to be a Jakarta REST endpoint.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: I thought about this more and I don't think it can be. We need the request to create the entity with a reader. CDI doesn't really need a request, at least not an HTTP one. We can't make it a CDI bean and assume it can just be injected let's say into a scheduled EJB.

jamezp:
Thinking more about this, I'm not sure what advantage there is to allowing CDI injection into parameters. CDI beans can be injected at the instance level or in the constructor. I don't really see why we'd need them injected in resource methods.

If we don't allow CDI injection into resource methods, we can keep backwards compatibility much easier.

jamezp:
What I've done is gone with the approach of removing the `@EntityParam` annotation. Arbitrary CDI are not allowed to be injected in resource methods. You can inject them via instance fields or constructor parameters though.

The same rules apply for determining the entity in a resource method. If the parameter is not annotated with `@*Param`, is not a known required context type, e.g. `UriInfo`, `HttpHeaders`, etc., or is of type `AsyncResponse`, then it is assumed to be the entity parameter.

I've removed the deprecation of the `ContextResolver` as a CDI producer is not really a direct replacement.

I also removed the OpenRewrite recipe and will publish it elsewhere.

hantsy:
> I've removed the deprecation of the ContextResolver as a CDI producer is not really a direct replacement.

I have tried to register a Jsonb via `ContextResovler` in my sample project before 5.0. 

It seems the `ContextResovler` can be replaced by CDI bean discovering mechanism directly in 5.0. For example, a `Jsonb` bean is available, it could be found in the Jakarta REST directly now because CDI is the injection provider now, no need to register a `ContextResolver` here.

jamezp:
> > I've removed the deprecation of the ContextResolver as a CDI producer is not really a direct replacement.
> 
> I have tried to register a Jsonb via `ContextResovler` in my sample project before 5.0.
> 
> It seems the `ContextResovler` can be replaced by CDI bean discovering mechanism directly in 5.0. For example, a `Jsonb` bean is available, it could be found in the Jakarta REST directly now because CDI is the injection provider now, no need to register a `ContextResolver` here.

The key difference is the way a `ContextResolver` is looked up. The signature looks like:
```java
jakarta.ws.rs.ext.Providers.getContextResolver(Class<T> contextType, MediaType mediaType)
```

This allows the lookup to be filtered by the `MediaType`. It also allows for more than one provider to be defined for a type. You could use a CDI qualifier for this too, but I think for now keeping the `ContextResolver` makes sense. That said, if others think we should deprecate it, I'm also not opposed to that.

hantsy:
An about Jakarta REST component `Sse`, in the 4.x it is only can be injected by `@Context` in a resource class. 

In 5.0, I think it could be `ApplicationScoped` or `Singleton` CDI bean, and can be used in a CDI backing bean directly that serves the Jakarta REST resource bean.

jamezp:
> An about Jakarta REST component `Sse`, in the 4.x it is only can be injected by `@Context` in a resource class.
> 
> In 5.0, I think it could be `ApplicationScoped` or `Singleton` CDI bean, and can be used in a CDI backing bean directly that serves the Jakarta REST resource bean.

Anything that could be injected with `@Context` will be required to use CDI.

jamezp:
@jakartaee/ee4j-rest-project-leads This could really use a review as this is for Jakarta REST 5.0 and we really need to get moving if we want to be included in the EE 12 platform.

arjantijms (approved):

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Provider.java
mkarg: So how to proceed?

jaxrs-api/src/main/java/jakarta/ws/rs/EntityParam.java
mkarg: Still in favor of `@EntityParam`. 🙂

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: I do not see the actual problem (but maybe I am simply missing it): `SeBootstrap` can span a CDI environment and register itself as a CDI provider for these resources.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: I think we should concentrate on "replacing `@Context` by CDI" solely, instead of "enabling CDI everywhere", which are two different things.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: Yes, it could be injected everywhere, but why should that be a bad thing?

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: Again, I am not a CDI expert. I do not know what CDI can or can't do. I am just looking at JAX-RS with the eyes of an application vendor. For me, I just see "things that can get injected" (and I expect them to be CDI beans) and "things that accept injections" (which is everything that has `@Inject` in it). Having said that, I actually do not care about EJB at all, so please forgive me that I say that I don't care whether or whether not an entity can or cannot get injected into a scheduled EJB.

jaxrs-spec/src/main/asciidoc/chapters/providers/_contextprovider.adoc
mkarg: Great, as I am using it often. 😅

mkarg:
> Thinking more about this, I'm not sure what advantage there is to allowing CDI injection into parameters. CDI beans can be injected at the instance level or in the constructor. I don't really see why we'd need them injected in resource methods.
> 
> If we don't allow CDI injection into resource methods, we can keep backwards compatibility much easier.

As I said before: Let's concentrate on replacing `@Context` by `@Inject` and simply leave out whether or whether not *other* CDI beans could get inject at that point.

mkarg (commented):
Waiting with my vote until all discussions are finished. Please see my latest comments. Thanks.

jaxrs-api/src/main/java/jakarta/ws/rs/ext/Provider.java
jamezp: IMO it's fine to allow the scope to be changed. While it likely wouldn't make sense for a `MessageBodyReader` or a `MessageBodyWriter` to be `@RequestScoped`, I could see where someone *might* want a `ServerRequestFilter` or `ServerResponseFilter` to be `@RequestScoped`. 

IoW, I don't know that we should worry too much here, but maybe I'm wrong.

jaxrs-api/src/main/java/jakarta/ws/rs/EntityParam.java
jamezp: Do you think we need it though? We have a 0 step migration if we don't allow CDI injection into resource methods. Any CDI beans that a resource method needs can be injected via instance fields.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: Correct and I actually do that in RESTEasy for some of the implementations. I do think we could require it for `SeBootstrap` given the is an SE container for CDI as well.

Where the provider part might not work with CDI is in the client. While a client used within the container would work, one used outside a container, some standalone JBang command for example, there is no CDI container.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: I actually removed this part about allowing parameter injection via CDI for resource methods. I think it gives us better backwards compatibility and allows us not to require that entity parameters be annotated. CDI beans can be injected as fields and be used.

If we allow arbitrary CDI beans to be injected in resource methods, users *must* use the `@EntityParam` annotation for entities. From an implementation perspective, I don't see how could keep backwards compatibility here.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: It's more of an implementation concern for me. For example, think of a `@PreMatching` provider. How would we know what type the entity is supposed to be? If we treat it as a true CDI bean, the bean could be injected into the provider.

The only potential thing we could do is maybe create our own custom CDI scope, which I'm not sure is worth the effort or if it would even solve the problem.

jaxrs-spec/src/main/asciidoc/chapters/providers/_contextprovider.adoc
jamezp: +1, I removed the deprecation here as a CDI producer is close to the same thing, but not really the same.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: Whether we like it or not, it's a platform requirement to support EJB's. If we want a platform compliant implementation of Jakarta REST, we need to support EJB's. What I mean by this is, at the implementation level if we have to support CDI for any valid CDI bean. For a REST entity, we need an HTTP request we may not have for various reasons. I don't see how we could create a CDI producer without an HTTP request. 

Even then, it shifts how the readers and writers would be resolved. We'd need to resolve them before we could even produce the bean which would be well before the actual resource method is selected. I don't even know how we'd get the HTTP request either :) It can't be a Servlet request because not all requests come from a Servlet and there is no generic HTTP request defined in any of the Jakarta specs.

jamezp:
> > Thinking more about this, I'm not sure what advantage there is to allowing CDI injection into parameters. CDI beans can be injected at the instance level or in the constructor. I don't really see why we'd need them injected in resource methods.
> > If we don't allow CDI injection into resource methods, we can keep backwards compatibility much easier.
> 
> As I said before: Let's concentrate on replacing `@Context` by `@Inject` and simply leave out whether or whether not _other_ CDI beans could get inject at that point.

A bit +1 from me here. I think we effectively just allow CDI injection of `@Context` resources. From the implementation perspective, this is quite easy as there are known types that can be injected like that.

FWIW, RESTEasy already allows this. You can use `@Inject` for the known `@Context` injection types. In RESTEasy with resteasy-cdi as of 7.0.3.Final, and even 6.2.17.Final, even the resource constructors and `@*Param` annotations are treated as CDI resources that can be injected into resources.

jaxrs-api/src/main/java/jakarta/ws/rs/EntityParam.java
mkarg: On one hand I am fine with simply disallowing CDI injection into resource methods, and this would be totally in line with the current behavior.

On the other hand, if the CDI lovers desparately want to inject, I am also fine with `@EntityParam`.

So if you ask me, we should ask a larger community which of both ways to go.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: I assume it is my missing knowledge about CDI, but why can't a JAX-RS-Client simply spawn its own CDI container internally (https://jakarta.ee/learn/docs/jakartaee-tutorial/current/cdi/cdi-bootstrap-se8/cdi-bootstrap-se8.html)? In the end, the topic of this PR just is to replace `@Context` within JAX-RS, not "how to use CDI in JBang". 🤔

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: As I commented earlier, I am fine with both.

jaxrs-api/src/main/java/jakarta/ws/rs/EntityParam.java
jamezp: IMO we should leave it off for 5.0 and we can look at it for REST.next. I say that because from an implementation perspective, there is no good way to determine when a method is either the old style or the new style. What I mean by that is if we say we allow arbitrary CDI beans to be injected, then it would be required for anyone wanting to run on Jakarta REST 5.0 to modify their endpoints to use `@EntityParam` for entity parameter.

What I means is:
```java
@POST
public Response post(Foo foo) {}
```

Would have to be changed to 
```java
@POST
public Response post(@EntityParam Foo foo) {}
```

It goes back to the point of I don't think we can make entities actual CDI beans. We need the HTTP request to create them and I'm not sure how we would do that.

Every `@POST` and `@PUT` in the TCK would need to change too. It's not impossible, but it's not backwards compatible.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: Counter question: In the `@PreMatching` case, how to know the entity type as of today when using `@Context`?

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: There is no reason we can't do that for a client, it just seems like a lot to boot up for every `Client.newClient()`. You'd have to start a new container or some how manage a CDI container for every client initialization.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: Skipping everything regarding EJBs from my side as I am not a Jakarta guy, I am solely a JAX-RS guy, and use, teach and recommend JAX-RS for SE-purposes solely (to be at most clear: I do not care about *any* Jakarta profile). Hence, that is an issue to be discussed among those that are into Jakarta themselves.

Yes, I agree that if we go down the "fully CDI" route, this would mean to turn runtime implementations upside-down. But IIUC we kind of agreed that the solution is to get rid of `@EntityParam` anyways. 😉

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: Fair point. I'm just struggling to figure out it would work. As an example:

```java
public MyBean {

    @Inject
    MyEntity entity;
    
    public MyEntity entity() {
        return entity;
    }
}
```

```java
@Path("/test")
public class MyResource {

    @POST
    public Response post(MyBean bean) {}
}
```

I don't know how I'd create a producer for the `MyEntity` before the `MyBean` is requested.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: Yes, this makes newClient slow, but OTOH this is why I always teach people that Client instances should not be short-lived. Anyways, no strong feelings here.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
mkarg: IMHO this example is off-topic in this PR, IMHO it is not possible in JAX-RS 4.x to do `@Context MyEntity entity` within the entity class. But I might be wrong.

jamezp: No, you're right you can't do that in 4.x. Also, I agree that I would never encourage that, but it's valid CDI was my only point. It's 100% code-smell though.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: I agree here as well. TBH I don't even know, with the exception of tests, how much the client is even used outside of a container.

For the most part I added the "CDI environment" types of wording for backwards compatibility. It was more of a "it felt safer to say it this way". In production, I would expect very few cases where CDI is not being used.

jaxrs-spec/src/main/asciidoc/chapters/environment/_javaee.adoc
jamezp: Correct, we don't really need the `@EntityParam` if we don't allow CDI injection into resource method parameters. There are ways around that so it feels reasonable to me to skip it. We can always look at it again in a later release.

mkarg: +1

Issue #579, opened 2017-12-08
Default value for @PathParam


glassfishrobot:
`@PathParam` annotation requires the name of tha param to be specified. This attribute could be made optional and default to the name of the annotated element.

So instead of

```
@Path("hello")
public class HelloResource {

    @GET
    @Path("{name}/{surname}")
    public String hello(@PathParam("name") String name, @PathParam("surname") String surname) {
        return "Hello, " + name + " " + surname;
    }
}
```

We could just do:
```
@Path("hello")
public class HelloResource {

    @GET
    @Path("{name}/{surname}")
    public String hello(@PathParam String name, @PathParam String surname) {
        return "Hello, " + name + " " + surname;
    }
}
```

Method parameters in particular might be more problematic since they are not always available at runtime, but setter or field injection would easily work.

FroMage:
Is there any chance we can reopen this issue for consideration? We have added an equivalent annotation in RESTEasy that does just that and it's much better for user code.

chkal:
Just to make sure I understand correctly. Java 8 supports preserving method parameter names in the byte code, is that correct? But it is not enabled by default?

chkal:
For the maven-compile-plugin it is available as an opt-in. See the documentation for [&lt;parameter&gt;](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#parameters).

FroMage:
Yes, and the frameworks can detect that it has not been enabled and warn if required.

mkarg:
As it seems a solution might be possible I have reopened this issue.

hantsy:
Spring also supports this feature by compiler parameters enablement.

mkarg:
> Spring also supports this feature by compiler parameters enablement.

Maybe you like to file a PR against branch `release-5.0`?

anagha-paulson:
Checking

PR #1343, opened 2026-06-09
Add default value for parameter binding annotations ( rest-issue-#579)

anagha-paulson:
This PR is a proposed implementation for #579: It allows to simply write `@PathParam` without the need to specify the parameter name.

### DIFF
--- a/examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
+++ b/examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0, which is available at
--- a/examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
+++ b/examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
@@ -29,7 +29,7 @@ import jakarta.ws.rs.QueryParam;
 /**
  * Examples demonstrating optional parameter names in Jakarta REST annotations.
  * 
- * <p>Starting with Jakarta REST 4.0, parameter binding annotations support optional
+ * <p>Starting with Jakarta REST 5.0, parameter binding annotations support optional
  * parameter names. When the annotation value is omitted, the parameter name is
  * inferred from the annotated element.</p>
  * 
@@ -39,9 +39,8 @@ import jakarta.ws.rs.QueryParam;
  * without this flag.</p>
  * 
  * <h4>Maven Configuration:</h4>
- * <pre>{@code
+ * {@snippet :
  * <plugin>
- *     <groupId>org.apache.maven.plugins</groupId>
  *     <artifactId>maven-compiler-plugin</artifactId>
  *     <configuration>
  *         <compilerArgs>
@@ -49,9 +48,7 @@ import jakarta.ws.rs.QueryParam;
  *         </compilerArgs>
  *     </configuration>
  * </plugin>
- * }</pre>
- * 
- * @see <a href="https://github.com/jakartaee/rest/issues/579">Issue #579</a>
+ * }
  */
 @Path("examples")
 public class OptionalParamNameExample {
@@ -60,15 +57,15 @@ public class OptionalParamNameExample {
      * Example 1: Path parameters with optional names.
      * 
      * <p><strong>Before (explicit names required):</strong></p>
-     * <pre>{@code
-     * public String hello(@PathParam("name") String name, 
-     *                    @PathParam("surname") String surname)
-     * }</pre>
-     * 
+     * {@snippet :
+     * public String hello(@PathParam("name") String name,
+     *                     @PathParam("surname") String surname)
+     * }
+     *
      * <p><strong>After (optional names):</strong></p>
-     * <pre>{@code
+     * {@snippet :
      * public String hello(@PathParam String name, @PathParam String surname)
-     * }</pre>
+     * }
      */
     @GET
     @Path("hello/{name}/{surname}")
@@ -102,9 +99,9 @@ public class OptionalParamNameExample {
      */
     @GET
     @Path("matrix/{path}")
-    public String matrix(@PathParam String path, 
-                        @MatrixParam String color,
-                        @MatrixParam String size) {
+    public String matrix(@PathParam String path,
+                         @MatrixParam String color,
+                         @MatrixParam String size) {
         return "Path: " + path + ", Color: " + color + ", Size: " + size;
     }
 
@@ -135,8 +132,8 @@ public class OptionalParamNameExample {
     @GET
     @Path("mixed/{id}")
     public String mixed(@PathParam String id,
-                       @QueryParam("q") String query,
-                       @HeaderParam String userAgent) {
+                        @QueryParam("q") String query,
+                        @HeaderParam String userAgent) {
         return "ID: " + id + ", Query: " + query + ", User-Agent: " + userAgent;
     }
 
@@ -163,10 +160,9 @@ public class OptionalParamNameExample {
     @GET
     @Path("profile/{userId}")
     public String profile(@PathParam String userId,
-                         @QueryParam String tab,
-                         @QueryParam String section,
-                         @HeaderParam String accept) {
-        return String.format("User: %s, Tab: %s, Section: %s, Accept: %s",
-                           userId, tab, section, accept);
+                          @QueryParam String tab,
+                          @QueryParam String section,
+                          @HeaderParam String accept) {
+        return "User: %s, Tab: %s, Section: %s, Accept: %s".formatted(userId, tab, section, accept);
     }
 }
--- a/examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
+++ b/examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package jaxrs.examples.params;
+
+import jakarta.ws.rs.CookieParam;
+import jakarta.ws.rs.FormParam;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.HeaderParam;
+import jakarta.ws.rs.MatrixParam;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.QueryParam;
+
+/**
+ * Examples demonstrating optional parameter names in Jakarta REST annotations.
+ * 
+ * <p>Starting with Jakarta REST 4.0, parameter binding annotations support optional
+ * parameter names. When the annotation value is omitted, the parameter name is
+ * inferred from the annotated element.</p>
+ * 
+ * <h3>Compiler Configuration Required</h3>
+ * <p>For method parameters, the {@code -parameters} compiler flag must be enabled
+ * to preserve parameter names in bytecode. Field and property injection always works
+ * without this flag.</p>
+ * 
+ * <h4>Maven Configuration:</h4>
+ * <pre>{@code
+ * <plugin>
+ *     <groupId>org.apache.maven.plugins</groupId>
+ *     <artifactId>maven-compiler-plugin</artifactId>
+ *     <configuration>
+ *         <compilerArgs>
+ *             <arg>-parameters</arg>
+ *         </compilerArgs>
+ *     </configuration>
+ * </plugin>
+ * }</pre>
+ * 
+ * @see <a href="https://github.com/jakartaee/rest/issues/579">Issue #579</a>
+ */
+@Path("examples")
+public class OptionalParamNameExample {
+
+    /**
+     * Example 1: Path parameters with optional names.
+     * 
+     * <p><strong>Before (explicit names required):</strong></p>
+     * <pre>{@code
+     * public String hello(@PathParam("name") String name, 
+     *                    @PathParam("surname") String surname)
+     * }</pre>
+     * 
+     * <p><strong>After (optional names):</strong></p>
+     * <pre>{@code
+     * public String hello(@PathParam String name, @PathParam String surname)
+     * }</pre>
+     */
+    @GET
+    @Path("hello/{name}/{surname}")
+    public String hello(@PathParam String name, @PathParam String surname) {
+        return "Hello, " + name + " " + surname;
+    }
+
+    /**
+     * Example 2: Query parameters with optional names.
+     * The parameter names 'query' and 'filter' are inferred from method parameters.
+     */
+    @GET
+    @Path("search")
+    public String search(@QueryParam String query, @QueryParam String filter) {
+        return "Searching for: " + query + " with filter: " + filter;
+    }
+
+    /**
+     * Example 3: Header parameters with optional names.
+     * The parameter name 'authorization' is inferred from the method parameter.
+     */
+    @GET
+    @Path("secure")
+    public String secure(@HeaderParam String authorization) {
+        return "Authorization: " + authorization;
+    }
+
+    /**
+     * Example 4: Matrix parameters with optional names.
+     * Matrix parameters are semicolon-separated in the path.
+     */
+    @GET
+    @Path("matrix/{path}")
+    public String matrix(@PathParam String path, 
+                        @MatrixParam String color,
+                        @MatrixParam String size) {
+        return "Path: " + path + ", Color: " + color + ", Size: " + size;
+    }
+
+    /**
+     * Example 5: Cookie parameters with optional names.
+     * The parameter name 'sessionId' is inferred from the method parameter.
+     */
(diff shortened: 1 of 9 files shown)


examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
```
jamezp: This should be 2026.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/params/defaultnames/DefaultParamNamesIT.java
```diff
@@ -0,0 +1,210 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
```
jamezp: This should be 2026.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/params/defaultnames/DefaultParamNamesResource.java
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright (c) 2024 Contributors to the Eclipse Foundation
```
jamezp: Same note here about the year

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/params/defaultnames/DefaultParamNamesIT.java
anagha-paulson: done

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: done

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/params/defaultnames/DefaultParamNamesResource.java
anagha-paulson: done

mkarg:
@anagha-paulson Do you know of a compatible implementation which already supports this feature?

anagha-paulson:
> @anagha-paulson Do you know of a compatible implementation which already supports this feature?

Hi @mkarg,

Not for Jakarta REST annotations (`jakarta.ws.rs.*`) specifically. However, RESTEasy already supported this pattern for its own annotations (`org.jboss.resteasy.annotations.jaxrs.*`) through parameter name inference using the getParamName() pattern.

I've extended the same mechanism to support Jakarta REST annotations. There wasn't an existing implementation for Jakarta annotations, but the underlying infrastructure was already available in RESTEasy.

jamezp:
FWIW RESTeasy will be able to implement this and @anagha-paulson has already opened a RESTEasy PR for it, https://github.com/resteasy/resteasy/pull/4889.

I'll have to figure out a way to coordinate testing, but it should be doable.

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
...
+ * 
+ * <h3>Compiler Configuration Required</h3>
+ * <p>For method parameters, the {@code -parameters} compiler flag must be enabled
+ * to preserve parameter names in bytecode. Field and property injection always works
+ * without this flag.</p>
+ * 
+ * <h4>Maven Configuration:</h4>
+ * <pre>{@code
+ * <plugin>
+ *     <groupId>org.apache.maven.plugins</groupId>
```
mkarg: CoC: No need for `groupId` as `org.apache.maven.plugins` is the default.

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
...
+ *     <artifactId>maven-compiler-plugin</artifactId>
+ *     <configuration>
+ *         <compilerArgs>
+ *             <arg>-parameters</arg>
+ *         </compilerArgs>
+ *     </configuration>
+ * </plugin>
+ * }</pre>
+ * 
+ * @see <a href="https://github.com/jakartaee/rest/issues/579">Issue #579</a>
```
mkarg: Is it a good idea to link to issues? We never discussed whether these are an official archive / part of the project, but always looked at it as a temporary means of discussion solely. 🤔

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
...
+@Path("examples")
+public class OptionalParamNameExample {
+
+    /**
+     * Example 1: Path parameters with optional names.
+     * 
+     * <p><strong>Before (explicit names required):</strong></p>
+     * <pre>{@code
+     * public String hello(@PathParam("name") String name, 
+     *                    @PathParam("surname") String surname)
```
mkarg: Strange indentation policy IMHO: It is *neither* eight blanks (two levels) *nor* matching the previous `@PathParm`.

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
...
+     * }</pre>
+     * 
+     * <p><strong>After (optional names):</strong></p>
+     * <pre>{@code
```
mkarg: Does it make sense to mix `<pre>` and `{code}`? 🤔

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * Example 4: Matrix parameters with optional names.
+     * Matrix parameters are semicolon-separated in the path.
+     */
+    @GET
+    @Path("matrix/{path}")
+    public String matrix(@PathParam String path, 
+                        @MatrixParam String color,
```
mkarg: Indentation?

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * Example 9: Multiple parameters of the same type.
+     * Each parameter name is inferred independently.
+     */
+    @GET
+    @Path("profile/{userId}")
+    public String profile(@PathParam String userId,
+                         @QueryParam String tab,
```
mkarg: Indentation?

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
...
+                         @QueryParam String section,
+                         @HeaderParam String accept) {
+        return String.format("User: %s, Tab: %s, Section: %s, Accept: %s",
```
mkarg: My personal preference is "".formatted().

jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
```diff
@@ -61,7 +61,15 @@
      * Defines the name of the HTTP cookie whose value will be used to initialize the value of the annotated method
      * argument, class field or bean property.
      *
-     * @return HTTP cookie name.
+     * <p>
+     * If the value is not specified (empty string), the name of the annotated element (parameter, field, or property)
+     * will be used as the parameter name. This requires that parameter names are preserved in the bytecode, which can be
+     * enabled using the {@code -parameters} compiler flag in Java 8+. If parameter names are not available at runtime
+     * and no explicit value is provided, implementations MUST throw a deployment-time exception with a clear error message
```
mkarg: "deployment-time" maybe does not exist in some implementation. For example, I can simply bootstrap Jersey without upfront deployment. When is Jersey expected to throw that exception? IIRC we do not have defined "deployment-time" in the spec / API before, but I might be wrong.

jaxrs-api/src/main/java/jakarta/ws/rs/FormParam.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2026 Oracle and/or its affiliates. All rights reserved.
```
mkarg: Please with all Copyright headers: Do *not* update them, but *remove* the second year. Thanks.

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
jamezp: That's a good point. I think I'd prefer not having links to issues in the Javadoc. We never know how long this will last and it may go stale. Especially given this is part of the API that gets documented.

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
jamezp: TBH given we will require Java 21, we should probably use `{@snippet}`.

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
jamezp: I don't have a strong opinion, but I do agree that `"".formatted()` would be preferred by me as well.

jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
jamezp: I do see what you're saying here. Other documentation does mention "deployment time". For example from the `DynamicFeature`:
```java
 * A JAX-RS meta-provider for dynamic registration of <i>post-matching</i> providers during a JAX-RS application setup
 * at deployment time.
```

That's also in the spec:
> It is RECOMMENDED to process dynamic features at application deployment time.

That said, I do agree that in current times there might not be a deployment so to speak. However, I personally don't have a better term and I do feel like it's generally understand what "deployment time" means.

We could use something like:
> implementations MUST throw an exception during initialization with a clear error message

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
jamezp: I prefer leaving off the `groupId` too, but I don't have a strong opinion either way.

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
```diff
@@ -0,0 +1,172 @@
...
+import jakarta.ws.rs.MatrixParam;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.QueryParam;
+
+/**
+ * Examples demonstrating optional parameter names in Jakarta REST annotations.
+ * 
+ * <p>Starting with Jakarta REST 4.0, parameter binding annotations support optional
```
jamezp: This should be "5.0". not "4.0".

jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
```diff
@@ -61,7 +61,15 @@
...
+     * indicating that either an explicit parameter name must be provided or the {@code -parameters} compiler flag must be enabled.
+     * </p>
```
jamezp, line 70: I'll make the comment here, but it applies to all the Javadoc on each annotation.

We need to decide how this works when the annotation is on a setter. For example:
```java
@QueryParam()
public void setId(final String value) {
    this.id = id;
}
```

Would we use "id" by resolving it from the setter or "value" by resolving it from the parameter name?

I think I prefer the JavaBean style of resolving it from the setter name, e.g. "id" in the example above. Doing something like this wouldn't require the `-parameters` compiler argument to be enabled for this specific use-case.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/params/defaultnames/DefaultParamNamesIT.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @AfterEach
+    void logFinishTest(TestInfo testInfo) {
+        TestUtil.logMsg("FINISHED TEST : " + testInfo.getDisplayName());
+    }
+
+    /*
+     * @testName: pathParamDefaultNameTest
+     * 
+     * @assertion_ids: JAXRS:SPEC:XXX;
```
jamezp, line 59: I don't want to block this PR on this and I asked on a different PR if we still need the assertion_id's. If we do, we'll need to address these.

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: Done

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: done

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: updated

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: updated

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: updated

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: updated

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: updated

jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
anagha-paulson: Updated

jaxrs-api/src/main/java/jakarta/ws/rs/FormParam.java
anagha-paulson: Updated

examples/src/main/java/jaxrs/examples/params/OptionalParamNameExample.java
anagha-paulson: Updated

jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
anagha-paulson: @jamezp

Thanks, that makes sense. I was thinking of handling it as follows, does this align with what you had in mind, or do you see a better approach?

_**Method parameters**:_ Resolve the parameter name (requires the -parameters compiler flag).
_**Fields**:_ Resolve the field name (does not require -parameters).
_**Setter methods (bean properties)**:_ Resolve the JavaBean property name from the setter method (e.g. setId(String value) → "id"), rather than the parameter name ("value"). This avoids requiring the -parameters compiler flag for the setter injection use case and follows standard JavaBean conventions.

If you're happy with this approach, I'll update the Javadoc to explicitly describe this behavior for bean properties instead of the current generic "parameter, field, or property" wording. If you have a better approach in mind, I'm happy to go with that instead.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/params/defaultnames/DefaultParamNamesIT.java
anagha-paulson: Thanks. I used XXX as a placeholder since the corresponding assertion IDs haven't been assigned yet. Happy to update or remove these depending on the outcome of the discussion about whether @assertion_ids are still required.

jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
jamezp: That makes sense to me. One thing I kind of wonder is if we should allow for:
```java
@QueryParam
public void id(final String value) {
    this.id = id;
}
```

Basically check if it starts with `set`, trim it, if not assume the method name is the parameter name value. I don't have a strong opinion there though.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/rs/params/defaultnames/DefaultParamNamesIT.java
jamezp: Sounds good. I don't really understand how these are used yet, so I think XXX is fine for now.

mkarg: I don't know either.

jamezp: I'll have to do some digging. I know they're here defined here https://github.com/jakartaee/rest/blob/main/jaxrs-tck/docs/assertions/JAXRSJavadocAssertions_4.0.0.xml and there is a static HTML page too which I assume is generated from that. I kind of wonder if it's a legacy thing from the old JavaTest framework.

jaxrs-api/src/main/java/jakarta/ws/rs/CookieParam.java
mkarg: Do we really want to fall back to JavaBean style? This feels like an anachronism. My personal opinion os that we should only concentrate on the parameter name and ignore the method name.

mkarg: But wait. Maybe we should check how JSON-B and JAX-B do it?

PR #1344, opened 2026-06-12
Added a new listener for events related to Client

### LINKED ISSUE #1283: Add a `onClose()` method to the client to indicate the connection has been closed
It would be nice for users, and implementations, to have something like an `onClose()` method on the `Client` API. It could look something like:
```java
CompletionStage<Void> onClose();
```

The `CompletionStage` would be completed when the `close()` method is invoked. While in most cases the user is in control of when the `Client` gets closed, there are cases like CDI where the client may be injected and the user wants to know then the client has been closed.

Implementations could also use this if the connection has been reset and/or a reconnect is not possible. In this case the `CompletionStage` can end exceptionally.

neenapj:


jamezp:
I think the `ClientListener` should be an interface with default no-op methods. Something like:

```java
public interface ClientListener {
    
    default void connected(final Client client) {
         // no-op
    }
    
    default void closed(final Client client) {
         // no-op
    }

    default void connectionFailed(final Client client, final Throwable cause) {
         // no-op
    }

    default void reconnecting(final Client client, final Throwable cause) {
         // no-op
    }
}
```

We don't really want to use an enum because it can make it migrations more difficult. Using an interface with no-ops by default allows implementors to implement what they want and not fail if a new enum is not handled.

mkarg:
I do not see that these events have anything to do with REST, but more with the underlying HTTP or TCP layers (possibly what you want to reach is already feasible by simple registering lower-level event notification?). It might even be impossible for for some implementations to correctly forward these events as their underlying networking framework might possibly not publish these events (for example, Jersey provides an implementation based on JRE's HTTP client, and IMHO that one is *unable* to fire those events). Unless we are are sure that we actually want to have lower-level functionality mixed into this application-level API, and unless we are sure that all (or at least *most*) existing compliant implementations are technically *able* to forward all of these events, we should not adopt this proposal.

jamezp:
To be fair, the client itself is not really REST related at all. It's a simple HTTP/TCP layer. This solves #1283 which I'd originally created because of an issue I was having with knowing when a client was closed via a CDI producer. 

FWIW I'm okay with only having the close notification as that is what I recall needing. I wish I could remember why, but I'm not remembering off the top of my head :)

FWIW there are other things in the JRE's HTTP client that can't be supported either. For example setting a `HostnameVerifier` on the `ClientBuilder`.

neenapj:
Hi @jamezp as per the feedbacks above, shall I remove connected(), connectionFailed() and reconnecting() from ClientListener and keep only closed(Client client) which is universally implementable?

jamezp:
> Hi @jamezp as per the feedbacks above, shall I remove connected(), connectionFailed() and reconnecting() from ClientListener and keep only closed(Client client) which is universally implementable?

Hi @neenapj I think we keep it for now unless @mkarg thinks we should keep reduce the scope.

mkarg:
I would say remove everything not needed, as Jersey *cannot* implement the other events besides *closed*.

jamezp:
That's good to know, thank you! Really, a close event notification is all I really needed. The others we thought might be useful. However, if they're too difficult to implement we can skip them. We can always consider adding them later.

@neenapj I'd suggest we delete everything but close for now.

neenapj:
HI @jamezp deleted everything else except close

jamezp:
This looks good to me. My only preference would be that we squash the commits down to a single commit.

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
```diff
@@ -272,4 +272,17 @@ public ClientBuilder keyStore(final KeyStore keyStore, final String password) {
...
+    /**
+     * Register a {@link ClientListener} to receive client lifecycle events.
+     * <p>
+     * The listener will be notified of events such as connection establishment,
+     * closure, failures, and reconnection attempts.
+     * </p>
+     *
+     * @param listener the client listener to register
+     * @return an updated client builder instance
+     * @since 4.0
```
mkarg: 5.0?

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
```diff
@@ -272,4 +272,17 @@ public ClientBuilder keyStore(final KeyStore keyStore, final String password) {
      * @return a new client instance.
      */
     public abstract Client build();
+
+    /**
+     * Register a {@link ClientListener} to receive client lifecycle events.
+     * <p>
+     * The listener will be notified of events such as connection establishment,
+     * closure, failures, and reconnection attempts.
```
mkarg: Just closure

mkarg (commented):
TCK needed

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
neenapj: Done

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
neenapj: Done

neenapj:
Hi @mkarg Tck added and the comments are updated.

neenapj:
Hi @jamezp 

> This looks good to me. My only preference would be that we squash the commits down to a single commit.

Do you want me to close this PR and create a new with single commit?

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/callbacks/ClientListenerIT.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+    * Verify a {@link ClientListener} registered with
+    * {@link ClientBuilder#listener(ClientListener)} is notified via
+    * {@link ClientListener#closed(Client)} when the built {@link Client} is closed.
+    */
+    @Test
+    public void closedCallbackInvokedOnClientCloseTest() {
+        AtomicBoolean closedCalled = new AtomicBoolean(false);
+        Client[] closedClient = new Client[1];
```
jamezp, line 69: Any reason we're using an array here?

jamezp:
> Hi @jamezp
> 
> > This looks good to me. My only preference would be that we squash the commits down to a single commit.
> 
> Do you want me to close this PR and create a new with single commit?

I would suggest squashing to a single commit and force pushing to your branch. The PR has good comments I'd rather not lose on it.

jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/api/client/callbacks/ClientListenerIT.java
neenapj: Hi @jamezp  The array is just a small mutable holder so the callback can capture the Client instance and the test can assert it afterward. Since variables referenced from the anonymous ClientListener must be effectively final, Client[] is a simple way to store the callback argument for the later identity check.

jamezp: Ah, I see what you mean. That seems fine.

neenapj:
> > Hi @jamezp
> > > This looks good to me. My only preference would be that we squash the commits down to a single commit.
> > 
> > 
> > Do you want me to close this PR and create a new with single commit?
> 
> I would suggest squashing to a single commit and force pushing to your branch. The PR has good comments I'd rather not lose on it.

I have squashed them into a single commit. Please have a look

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026 Contributors to the Eclipse Foundation
```
jamezp: IIRC we're not supposed to update the copyright year and I don't think we can change the contributors. See https://www.eclipse.org/projects/handbook/#ip-copyright-headers.

jamezp (changes requested):
Thank you @neenapj. One minor change needed for the copyright and then I can give it my approval.

neenapj:
Hi @jamezp changes are done. Kindly review

mkarg:
@neenapj For any new Jakarta Component Specification Feature there must exist at least one compliant implementation before releasing the spec. Hence, at least one of the current Client implementations MUST proof compliance by passing your new IT. Would you like to provide a PR for the needed changes in Jersey to fulfil this rule?

jamezp:
I can setup a RESTEasy branch for this, but that shouldn't block merging a feature.

jamezp (dismissed):

mkarg:
If we merge it *now*, it blocks Jakarta REST 5.0 release completely *later* if no Compliant Implementation exists (due to EF's "Implementation First" policy), so it is easier if any product implements it *upfront*.

jamezp:
> If we merge it _now_, it blocks Jakarta REST 5.0 release completely _later_ if no Compliant Implementation exists (due to EF's "Implementation First" policy), so it is easier if any product implements it _upfront_.

IMO we're already blocked. The only thing that makes a 5.0 worth it is the CDI integration. Otherwise we may as well do a 4.1 if the platform even wants that.

mkarg:
Let's say, we're *delayed*, but not *blocked*. Nothing prevents us from publishing 5.0 even *without* CDI, and speaking of CDI, I am confident that I will find the time to review your changes in my summer break.

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientBuilder.java
```diff
@@ -272,4 +272,16 @@ public ClientBuilder keyStore(final KeyStore keyStore, final String password) {
...
+     * Register a {@link ClientListener} to receive client lifecycle events.
+     * <p>
+     * The listener will be notified of events such as connection closure
+     * </p>
+     *
+     * @param listener the client listener to register
+     * @return an updated client builder instance
+     * @since 5.0
+     */
+    public abstract ClientBuilder listener(ClientListener listener);
```
mkarg, line 279: Would you mind moving this *before* the terminal `build` method?

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientListener.java
```diff
@@ -0,0 +1,40 @@
...
+ * A listener interface for receiving client close notifications.
+ * <p>
+ * Implementations of this interface can be registered with a {@link ClientBuilder}
+ * to receive notification when a {@link Client} is closed.
+ * </p>
+ */
+public interface ClientListener {
+
+    /**
+     * Called when the client is closed.
```
mkarg, line 29: To be clear: What we want here is that calling this listener is the very last to happen *after* a client is finished shutdown, right?

mkarg (commented):
Seems we can do this in Jersey, see experimental code in https://github.com/eclipse-ee4j/jersey/pull/6121. Waiting for Team Jersey before voting +1 on this new API.

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientListener.java
neenapj: Yes @mkarg

jamezp:
The RESTEasy PR is up too https://github.com/resteasy/resteasy/pull/5006. I did suggest one more change here to change the signature to:
```java
void closed(Client client, Throwable t);
```

It seemed like a good idea to add the throwable argument for cases when close may have failed.

mkarg:
> The RESTEasy PR is up too [resteasy/resteasy#5006](https://github.com/resteasy/resteasy/pull/5006). I did suggest one more change here to change the signature to:
> 
> ```java
> void closed(Client client, Throwable t);
> ```
> 
> It seemed like a good idea to add the throwable argument for cases when close may have failed.

It feels a bit odd that an event called "closed" is invoked when actually the closing *failed*.

jamezp:
> > The RESTEasy PR is up too [resteasy/resteasy#5006](https://github.com/resteasy/resteasy/pull/5006). I did suggest one more change here to change the signature to:
> > ```java
> > void closed(Client client, Throwable t);
> > ```
> > It seemed like a good idea to add the throwable argument for cases when close may have failed.
> 
> It feels a bit odd that an event called "closed" is invoked when actually the closing _failed_.

I guess that is true. I was thinking a scenario where maybe you have a broken pipe, the error invokes `Client.close()` which also throws an error. Maybe we shouldn't accept add the `Throwable`, but have some wording that indicates `closed(Client)` will be invoked regardless if the `Client.close()` fails or not. Well the underlying client really. I definitely see your point though.

neenapj:
> > > The RESTEasy PR is up too [resteasy/resteasy#5006](https://github.com/resteasy/resteasy/pull/5006). I did suggest one more change here to change the signature to:
> > > ```java
> > > void closed(Client client, Throwable t);
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > It seemed like a good idea to add the throwable argument for cases when close may have failed.
> > 
> > 
> > It feels a bit odd that an event called "closed" is invoked when actually the closing _failed_.
> 
> I guess that is true. I was thinking a scenario where maybe you have a broken pipe, the error invokes `Client.close()` which also throws an error. Maybe we shouldn't accept add the `Throwable`, but have some wording that indicates `closed(Client)` will be invoked regardless if the `Client.close()` fails or not. Well the underlying client really. I definitely see your point though.

Hi @jamezp @mkarg shall we update Javadoc like this?

```
/**
 * Called when {@link Client#close()} is invoked, regardless of whether the
 * underlying resources were released successfully. If the close failed,
 * the client is still considered closed and this method will still be called.
 * Implementations should not attempt to use the client after this point.
 *
 * @param client the client that was closed
 */
default void closed(Client client) {
}
```

mkarg:
> I guess that is true. I was thinking a scenario where maybe you have a broken pipe, the error invokes `Client.close()` which also throws an error. Maybe we shouldn't accept add the `Throwable`, but have some wording that indicates `closed(Client)` will be invoked regardless if the `Client.close()` fails or not. Well the underlying client really. I definitely see your point though.

The question is: What benefit does somebody have to get information about close() being called if it actually failed? IMHO the "closed(client)" event should only get called if the close was *successful*. Hence, it tells the listener that client is actually closed, not that it was *tried* to close it or that the explicit method named "close()" was called.

neenapj:
> > I guess that is true. I was thinking a scenario where maybe you have a broken pipe, the error invokes `Client.close()` which also throws an error. Maybe we shouldn't accept add the `Throwable`, but have some wording that indicates `closed(Client)` will be invoked regardless if the `Client.close()` fails or not. Well the underlying client really. I definitely see your point though.
> 
> The question is: What benefit does somebody have to get information about close() being called if it actually failed? IMHO the "closed(client)" event should only get called if the close was _successful_. Hence, it tells the listener that client is actually closed, not that it was _tried_ to close it or that the explicit method named "close()" was called.

Hi @jamezp @mkarg Are we going with this? i.e. closed(client) gets called only when close was success?

jaxrs-api/src/main/java/jakarta/ws/rs/client/ClientListener.java
```diff
@@ -0,0 +1,40 @@
...
+     * <p>
+     * This method is invoked when {@link Client#close()} is called or when
+     * the client is closed by the container (e.g., during CDI shutdown).
+     * Implementations can use this to perform cleanup operations.
+     * </p>
+     *
+     * @param client the client that was closed
+     */
+    default void closed(Client client) {
+    }
```
jamezp, line 39: FWIW given now we're down to 1 method, we should likely not have it be a default.

jamezp:
One minor comment, then I'm okay with it @neenapj

mkarg:
Still don't like that `closed()` is called if the client is actually failed to close.

jamezp:
> Still don't like that `closed()` is called if the client is actually failed to close.

I'm fine with that as well. I think the current state of the PR does not state that it will always be called.

jamezp (approved):
FWIW I'm okay with it in it's current state. I don't know that we need the interface method to be default, but it also doesn't hurt in case we want to add anything in the future.

Issue #1345, opened 2026-06-19
Add First-Class Support for the HTTP QUERY Method (RFC 10008)


otaviojava:
This proposal introduces first-class support for the HTTP QUERY method, as defined by RFC 10008, into Jakarta REST.

## Server API

This would be based on a single annotation: 

```java

@Target(ElementType.METHOD)
@Retention(RetentionPolicy.RUNTIME)
@HttpMethod("QUERY")
public @interface QUERY {
}
```

Usage:

```java
@Path("/users")
public class UserResource {

    @QUERY
    @Consumes(MediaType.APPLICATION_JSON)
    @Produces(MediaType.APPLICATION_JSON)
    public List<User> search(UserFilter filter) {
        return service.search(filter);
    }
}
```

###  Client API

Provide convenience methods for invoking QUERY requests without requiring generic method invocation.

Example:

```java
client.target(uri)
      .request()
      .query(Entity.json(filter));
```
or equivalent API design determined by the specification committee.

## Specification Goal

Define the semantics of QUERY according to RFC 10008:

* Safe
* Idempotent
* Allows request content
* Intended for retrieval and query operations
* Not intended for resource modification

hantsy:
I have filed #1324

otaviojava:
> I have filed [#1324](https://github.com/jakartaee/rest/issues/1324)

Thanks, @hantsy, but I'll take a different approach: I'll provide more code details instead of a single link.

If the team agrees, I can move forward with a contribution.

hantsy:
It seems Servlet has already included HTTP QUERY method in 6.2, https://github.com/jakartaee/servlet/pull/1069/.

mkarg:
> It seems Servlet has already included HTTP QUERY method in 6.2, [jakartaee/servlet#1069](https://github.com/jakartaee/servlet/pull/1069).

IMHO this is irrelevant for Jakarta REST, as (a) Jakarta REST is not dependend of Servlet, (b) Jakarta REST is a REST API while Servlet is a HTTP API.

mkarg:
> If the team agrees, I can move forward with a contribution.

Actually there is no majority in team for adopting QUERY yet. This might change in future, but currently most committers did not vote, which does *not* imply agreement.

sandeshkuldharme1:
Hi @otaviojava ,

I would love to work on this! 

The proposal is clear. I can handle:
[...]

Issue #1324, opened 2025-07-30
Add HTTP Query Method Support


hantsy:
See: https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html#name-simple-query-with-a-direct-

gastaldi:
FWIW, that RFC draft has long expired, this is the last version: https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-11.html

hantsy:
Hello @gastaldi, more than ten years have passed since we parted in Boston in 2012. How are you?

Yeah, a lot of IETF drafts labelled as expired, not means it is not active. For example, HAL also labelled as expired, but still actively used in developer's community.

jeffque:
We got a RFC! https://www.rfc-editor.org/info/rfc10008/

jeffque:
As far as I could (for now) see, adding the new method seems trivial, but there is a catch: as QUERY is cacheable, the current cache in `CacheResponseFilter` needs to be smarter to allow for caching QUERY with body. For now it only caches based on URI (as cache key) for GET method, but now it needs to cache based on:

- URI
- method
- body (if method used is `QUERY` of course)

The sectio about [caching](https://www.rfc-editor.org/info/rfc10008/#caching) mentions many other stuff that I cannot foresee how to implement, like "MAY remove semantically insignificant differences from request content "

hantsy:
@gastaldi It is a published status now, https://datatracker.ietf.org/doc/rfc10008/history/

PR #1346, opened 2026-06-19
HTTP Query method

jeffque:
Basic implementation, no cache example for now

Tackling #1324 , #1345

Open for suggestions, there is a bunch that can be made in cache example

### DIFF
--- a/examples/src/main/java/jaxrs/examples/bootstrap/HelloWorld.java
+++ b/examples/src/main/java/jaxrs/examples/bootstrap/HelloWorld.java
@@ -20,8 +20,11 @@ import java.util.Collections;
 import java.util.Set;
 
 import jakarta.ws.rs.ApplicationPath;
+import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.QUERY;
 import jakarta.ws.rs.core.Application;
 
 /**
@@ -50,4 +53,10 @@ public class HelloWorld extends Application {
         return "Hello, World!";
     }
 
+    @QUERY
+    @Consumes("application/json")
+    @Produces("text/plain")
+    public String queryIt(final String message) {
+        return String.format("Got it from query! <%s>", message);
+    }
 }
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/HttpMethod.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/HttpMethod.java
@@ -73,6 +73,10 @@ public @interface HttpMethod {
      * HTTP OPTIONS method.
      */
     public static final String OPTIONS = "OPTIONS";
+    /**
+     * HTTP OPTIONS method.
+     */
+    public static final String QUERY = "QUERY";
 
     /**
      * Specifies the name of a HTTP method. E.g. "GET".
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/HttpMethod.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/HttpMethod.java
@@ -36,6 +36,7 @@ import java.lang.annotation.Target;
  * @see PATCH
  * @see HEAD
  * @see OPTIONS
+ * @see QUERY
  * @since 1.0
  */
 @Target({ ElementType.ANNOTATION_TYPE })
@@ -74,7 +75,7 @@ public @interface HttpMethod {
      */
     public static final String OPTIONS = "OPTIONS";
     /**
-     * HTTP OPTIONS method.
+     * HTTP QUERY method.
      */
     public static final String QUERY = "QUERY";
 
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/QUERY.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/QUERY.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+
+package jakarta.ws.rs;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates that the annotated method responds to HTTP QUERY requests.
+ *
+ * @see HttpMethod
+ * @since 5.0.0
+ */
+@Target({ ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@HttpMethod(HttpMethod.QUERY)
+@Documented
+public @interface QUERY {
+}
--- a/jaxrs-api/src/main/java/jakarta/ws/rs/client/AsyncInvoker.java
+++ b/jaxrs-api/src/main/java/jakarta/ws/rs/client/AsyncInvoker.java
@@ -96,6 +96,84 @@ public interface AsyncInvoker {
      */
     <T> Future<T> get(InvocationCallback<T> callback);
 
+    // QUERY
+
+    /**
+     * Invoke HTTP QUERY method for the current request asynchronously.
+     * <p>
+     * Note that calling the {@link java.util.concurrent.Future#get()} method on the returned {@code Future} instance may
+     * throw an {@link java.util.concurrent.ExecutionException} that wraps a {@link jakarta.ws.rs.ProcessingException} thrown
+     * in case of an invocation processing failure. In case a processing of a properly received response fails, the wrapped
+     * processing exception will be of {@link ResponseProcessingException} type and will contain the {@link Response}
+     * instance whose processing has failed.
+     *
+     * @param entity request entity, including its full {@link jakarta.ws.rs.core.Variant} information. Any variant-related
+     * HTTP headers previously set (namely {@code Content-Type}, {@code Content-Language} and {@code Content-Encoding}) will
+     * be overwritten using the entity variant information.
+     * @return invocation response {@link Future future}.
+     */
+    Future<Response> query(Entity<?> entity);
+
+    /**
+     * Invoke HTTP QUERY method for the current request asynchronously.
+     * <p>
+     * Note that calling the {@link java.util.concurrent.Future#get()} method on the returned {@code Future} instance may
+     * throw an {@link java.util.concurrent.ExecutionException} that wraps either a {@link jakarta.ws.rs.ProcessingException}
+     * thrown in case of an invocation processing failure or a {@link WebApplicationException} or one of its subclasses
+     * thrown in case the received response status code is not {@link jakarta.ws.rs.core.Response.Status.Family#SUCCESSFUL
+     * successful} and the specified response type is not {@link jakarta.ws.rs.core.Response}. In case a processing of a
+     * properly received response fails, the wrapped processing exception will be of {@link ResponseProcessingException}
+     * type and will contain the {@link Response} instance whose processing has failed.
+     *
+     * @param <T> response entity type.
+     * @param entity request entity, including its full {@link jakarta.ws.rs.core.Variant} information. Any variant-related
+     * HTTP headers previously set (namely {@code Content-Type}, {@code Content-Language} and {@code Content-Encoding}) will
+     * be overwritten using the entity variant information.
+     * @param responseType Java type the response entity will be converted to.
+     * @return invocation response {@link Future future}.
+     */
+    <T> Future<T> query(Entity<?> entity, Class<T> responseType);
+
+    /**
+     * Invoke HTTP QUERY method for the current request asynchronously.
+     * <p>
+     * Note that calling the {@link java.util.concurrent.Future#get()} method on the returned {@code Future} instance may
+     * throw an {@link java.util.concurrent.ExecutionException} that wraps either a {@link jakarta.ws.rs.ProcessingException}
+     * thrown in case of an invocation processing failure or a {@link WebApplicationException} or one of its subclasses
+     * thrown in case the received response status code is not {@link jakarta.ws.rs.core.Response.Status.Family#SUCCESSFUL
+     * successful} and the specified response type is not {@link jakarta.ws.rs.core.Response}. In case a processing of a
+     * properly received response fails, the wrapped processing exception will be of {@link ResponseProcessingException}
+     * type and will contain the {@link Response} instance whose processing has failed.
+     *
+     * @param <T> generic response entity type.
+     * @param entity request entity, including its full {@link jakarta.ws.rs.core.Variant} information. Any variant-related
+     * HTTP headers previously set (namely {@code Content-Type}, {@code Content-Language} and {@code Content-Encoding}) will
+     * be overwritten using the entity variant information.
+     * @param responseType representation of a generic Java type the response entity will be converted to.
+     * @return invocation response {@link Future future}.
+     */
+    <T> Future<T> query(Entity<?> entity, GenericType<T> responseType);
+
+    /**
+     * Invoke HTTP QUERY method for the current request asynchronously.
+     * <p>
+     * Note that calling the {@link java.util.concurrent.Future#get()} method on the returned {@code Future} instance may
+     * throw an {@link java.util.concurrent.ExecutionException} that wraps either a {@link jakarta.ws.rs.ProcessingException}
+     * thrown in case of an invocation processing failure or a {@link WebApplicationException} or one of its subclasses
+     * thrown in case the received response status code is not {@link jakarta.ws.rs.core.Response.Status.Family#SUCCESSFUL
+     * successful} and the generic type of the supplied response callback is not {@link jakarta.ws.rs.core.Response}. In case
+     * a processing of a properly received response fails, the wrapped processing exception will be of
+     * {@link ResponseProcessingException} type and will contain the {@link Response} instance whose processing has failed.
+     *
+     * @param <T> generic response entity type.
+     * @param entity request entity, including its full {@link jakarta.ws.rs.core.Variant} information. Any variant-related
+     * HTTP headers previously set (namely {@code Content-Type}, {@code Content-Language} and {@code Content-Encoding}) will
+     * be overwritten using the entity variant information.
+     * @param callback asynchronous invocation callback.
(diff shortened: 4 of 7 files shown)


jaxrs-api/src/main/java/jakarta/ws/rs/HttpMethod.java
```diff
@@ -73,6 +73,10 @@
      * HTTP OPTIONS method.
      */
     public static final String OPTIONS = "OPTIONS";
+    /**
+     * HTTP OPTIONS method.
```
hantsy: `OPTIONS` -> `QUERY`

jeffque: Solved

jamezp:
We would also need TCK tests for this. My request for that is we don't follow the old pattern and we use a better more modern test style. Something more like https://github.com/jakartaee/rest/blob/main/jaxrs-tck/src/main/java/ee/jakarta/tck/ws/rs/jaxrs31/ee/multipart/MultipartSupportIT.java. I'm happy to help write if needed too.

jeffque:
@jamezp , I appreciate it ❤️! I'll have little time in the next weeks, and I am not familiar with the TCK tests

mkarg:
As JAX-RS is a REST API (not a generic HTTP API), and as QUERY is not restful (it is RPC style), I will not vote +1 here.
OTOH I will not stand in the way if there is a majority of committers who are voting +1 for this (in my sense misplaced) addition.

FroMage:
> QUERY is not restful (it is RPC style)

Why makes you say that?

In my opinion, QUERY is just a GET with the same semantics, but an alternate way to provide query parameters that isn't limited to URI limitations.

If GET is restful, then QUERY is restful too.

mkarg:
> > QUERY is not restful (it is RPC style)
> 
> Why makes you say that?
> 
> In my opinion, QUERY is just a GET with the same semantics, but an alternate way to provide query parameters that isn't limited to URI limitations.
> 
> If GET is restful, then QUERY is restful too.

Fielding's dissertation. QUERY makes the result dependent of parameters. REST differs from RPC in being dependent of the object's identified (URI) solely.

hantsy:
> > > QUERY is not restful (it is RPC style)
> > 
> > 
> > Why makes you say that?
> > In my opinion, QUERY is just a GET with the same semantics, but an alternate way to provide query parameters that isn't limited to URI limitations.
> > If GET is restful, then QUERY is restful too.
> 
> Fielding's dissertation. QUERY makes the result dependent of parameters. REST differs from RPC in being dependent of the object's identified (URI) solely.

There is a situation in the real world REST API development, sometime there are too many parameters in `GET` URLs, eg. designing a complex search options, then maintaining becomes a little tedious. In before experience, personally I would like to use `POST` for it, assemble the query parameters into a body (respectively use a class to wrap them together), it is obviously not a good match with the RESTful style.

But currently it seems the `QUERY` method is a good match for this use case.

otaviojava:
@jamezp, does it make sense to review this one and, as the next step, have a second PR with the TCK?

I mean, this PR already has 7 files.

Also, please let me know how I can help.

examples/src/main/java/jaxrs/examples/bootstrap/HelloWorld.java
```diff
@@ -50,4 +53,10 @@ public String sayHello() {
         return "Hello, World!";
     }
 
+    @QUERY
+    @Consumes("application/json")
```
otaviojava, line 57: Should we use the `MediaType` instead?

FroMage:
> Fielding's dissertation. QUERY makes the result dependent of parameters. 

Which is already the case with GET, so I must again be missing a point.

> REST differs from RPC in being dependent of the object's identified (URI) solely

This is not true of POST or PUT, and only true of GET because the parameters are in the URI, but only because of convention and because at the time QUERY did not exist and POST was used for mutation. Where the parameters are located is irrelevant to the intent of GET, as long as the behaviour holds and can be respected across the route. Which should be the case with QUERY.

examples/src/main/java/jaxrs/examples/bootstrap/HelloWorld.java
jeffque: Indeed!

Perhaps even removing completely this annotation? As the service does not do anything worth with the received data.

otaviojava: We need a better scenario for it.

Maybe a record that simulates a filter.

mkarg:
> > Fielding's dissertation. QUERY makes the result dependent of parameters.
> 
> Which is already the case with GET, so I must again be missing a point.

This is *not* the case for GET if used *correctly*: One MUST NOT have query parameters, as these are not RESTful.
 
> > REST differs from RPC in being dependent of the object's identified (URI) solely
> This is not true of POST or PUT, and only true of GET because the parameters are in the URI, but only because of convention and because at the time QUERY did not exist and POST was used for mutation. Where the parameters are located is irrelevant to the intent of GET, as long as the behaviour holds and can be respected across the route. Which should be the case with QUERY.

Again, parameters are not RESTful, independent of what people do in reality, even with POST. A RESTful use is:
* GET /address
* POST /address
* PUT /address
* DELETE /address
Once the outcome is dependend on parameters (independent if these are query, matrix or header parameters) you do not have a RESTful API anymore, but solely a HTTP based one.

otaviojava:
> Again, parameters are not RESTful, independent of what people do in reality, even with POST. A RESTful use is:
> 
> * GET /address
> * POST /address
> * PUT /address
> * DELETE /address
>   Once the outcome is dependend on parameters (independent if these are query, matrix or header parameters) you do not have a RESTful API anymore, but solely a HTTP based one.

I do not agree that `QUERY` is inherently non-RESTful or strictly RPC-style. Much of the confusion stems from interpreting REST solely through the lens of the Richardson Maturity Model, which, while valuable, reflects an earlier context focused on `GET`, `POST`, `PUT`, and `DELETE` methods. `QUERY` was not considered because the HTTP QUERY method was not standardized at that time. Jakarta REST should not be limited to the HTTP methods commonly discussed in **2008**.


Fielding’s dissertation states that REST operates on resources identified by resource identifiers, which does not restrict usage to only the path or exclude parameters. In HTTP, the target URI may include both the path and the query component. For example, `GET /address?city=Lisbon` remains RESTful when it provides a filtered view of the address collection. The distinction between REST and RPC is not based on whether the response depends on input, as standard HTTP requests already rely on query parameters, headers, pagination, sorting, filtering, and content negotiation. The primary concern is command tunneling, such as `POST /address?action=delete`. RFC 10008 defines QUERY as a safe and idempotent method for requests with content, so `QUERY /address`  with search criteria still means “query this resource collection,” rather than “call this remote procedure.”

References:

* Fielding dissertation, REST: https://ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm
* Richardson Maturity Model / “Glory of REST”: https://martinfowler.com/articles/richardsonMaturityModel.html
* RFC 9110, HTTP Semantics, target resource: https://www.rfc-editor.org/rfc/rfc9110.html
* RFC 10008, The HTTP QUERY Method: https://www.rfc-editor.org/rfc/rfc10008.html

mkarg:
Let's agree to disagree.

ugautam-mdsol (approved):

jamezp:
> idempotent

My main concern there is in the past we've introduced API changes and the TCK tests have not made it in. We could consider filing a follow up issue as to not block this if it moves forward I suppose. 

I need to review all the comments and the spec to fully grok all the details :)

otaviojava:
> > idempotent
> 
> My main concern there is in the past we've introduced API changes and the TCK tests have not made it in. We could consider filing a follow up issue as to not block this if it moves forward I suppose.
> 
> I need to review all the comments and the spec to fully grok all the details :)

Maybe create an issue for each missing point, such as TCK, spec, etc.
Please let us know how we can help.

mkarg:
Can we please agree upon the best practice that any new API adopted in a PR should also have the corresponding TCK change in *the same* PR? That makes tracking way simpler.

mkarg:
Please also note that JAX-RS is only a *foundational* API, not a *framework* in its own. Jersey is a framework for example (not a server in its own). When we invented JAX-RS our vision was that *frameworks* (or at least third party libraries) build *ontop* of JAX-RS to provide more features and usability *external* of JAX-RS. For example, in 2008 or so I implemented "WebDAV Support for JAX-RS" completely *outside* of JAX-RS, but building *ontop*. It is an *extension* to JAX-RS which contained Filters, Methods, Providers, and so on. What I like to make you understand is this vision that the old committers had back then.

From this view, `QUERY` should go into such an *extension* as it is not essentially needed in the *foundational* core of JAX-RS: If someone needs `QUERY` he can simply extend JAX-RS in few lines of code. This is why we *made* JAX-RS extensible in first place.

Having said that, still I am not vetoing its adoption. I am just not that committer that will give the needed +1, because of exactly that vision.

mkarg:
> > > idempotent
> > 
> > 
> > My main concern there is in the past we've introduced API changes and the TCK tests have not made it in. We could consider filing a follow up issue as to not block this if it moves forward I suppose.
> > I need to review all the comments and the spec to fully grok all the details :)
> 
> Maybe create an issue for each missing point, such as TCK, spec, etc. Please let us know how we can help.

What would be appreciated would be implementing `QUERY` in the majority of existing JAX-RS implementations first, *before* requesting adoption to the API (and effectively *forcing* vendors to implement it, even if they don't want to).

jamezp:
> Can we please agree upon the best practice that any new API adopted in a PR should also have the corresponding TCK change in _the same_ PR? That makes tracking way simpler.

I'm a +1 for this idea. I think it also helps with understanding the authors intended usage and for others to form opinions based on that. Plus implementations have some to immediately test with.

jamezp:
I'm a +1 for this.

While I can appreciate the point that Jakarta REST aims to be strictly RESTful and that `QUERY` leans into RPC-style resource identification, we have to look at the pragmatic reality of modern API development. Fielding's dissertation was written 26 years ago, and the industry has evolved to handle use cases (like massive, complex, or sensitive search filters) that purely URI-based `GET` requests simply cannot accommodate.

I can also see the argument for third-party implementations for things like WebDAV. However, I would argue that `QUERY` is fundamentally different from a niche protocol extension.

Currently, developers are forced to use non-idempotent `POST` requests as a workaround for these complex searches. If we leave `QUERY` to third-party extensions, we risk fragmentation across the Java ecosystem where every framework writes their own `@QUERY` annotation. Adding it to the foundational API gives developers a standardized, idempotent alternative right out of the box that solves a very real developer pain point.

With the IETF finalizing RFC 10008 just last month (June 2026), and other major ecosystems like Go and .NET 10 already implementing it, I believe this is a highly reasonable compromise for modern times. Also, because the RFC was co-authored by engineers from major CDNs, supporting `QUERY` ensures Java developers can take advantage of edge-caching for these complex reads.

jaxrs-api/src/main/java/jakarta/ws/rs/QUERY.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2010, 2019 Oracle and/or its affiliates. All rights reserved.
```
jamezp, line 2: Given this is new, we need to update the copyright here. The date should be 2026 and this does not belong to Oracle, but to Eclipse. Something like:

```java
/*
 * Copyright (c) 2026 Contributors to the Eclipse Foundation
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0, which is available at
 * http://www.eclipse.org/legal/epl-2.0.
 *
 * This Source Code may also be made available under the following Secondary
 * Licenses when the conditions for such availability set forth in the
 * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
 * version 2 with the GNU Classpath Exception, which is available at
 * https://www.gnu.org/software/classpath/license.html.
 *
 * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 */
```

jaxrs-api/src/main/java/jakarta/ws/rs/client/SyncInvoker.java
```diff
@@ -69,6 +69,57 @@ public interface SyncInvoker {
...
+     * Invoke HTTP QUERY method for the current request synchronously.
+     *
+     * @param entity request entity, including its full {@link jakarta.ws.rs.core.Variant} information. Any variant-related
+     * HTTP headers previously set (namely {@code Content-Type}, {@code Content-Language} and {@code Content-Encoding}) will
+     * be overwritten using the entity variant information.
+     * @return invocation response.
+     * @throws ResponseProcessingException in case processing of a received HTTP response fails (e.g. in a filter or during
+     * conversion of the response entity data to an instance of a particular Java type).
+     * @throws ProcessingException in case the request processing or subsequent I/O operation fails.
+     */
```
jamezp, line 84: We should add `@since 5.0` on all of these new methods.

jaxrs-api/src/main/java/jakarta/ws/rs/client/AsyncInvoker.java
```diff
@@ -96,6 +96,84 @@ public interface AsyncInvoker {
...
+     * Note that calling the {@link java.util.concurrent.Future#get()} method on the returned {@code Future} instance may
+     * throw an {@link java.util.concurrent.ExecutionException} that wraps a {@link jakarta.ws.rs.ProcessingException} thrown
+     * in case of an invocation processing failure. In case a processing of a properly received response fails, the wrapped
+     * processing exception will be of {@link ResponseProcessingException} type and will contain the {@link Response}
+     * instance whose processing has failed.
+     *
+     * @param entity request entity, including its full {@link jakarta.ws.rs.core.Variant} information. Any variant-related
+     * HTTP headers previously set (namely {@code Content-Type}, {@code Content-Language} and {@code Content-Encoding}) will
+     * be overwritten using the entity variant information.
+     * @return invocation response {@link Future future}.
```
jamezp, line 113: We should add `@since 5.0` on all these methods.

jaxrs-api/src/main/java/jakarta/ws/rs/QUERY.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates that the annotated method responds to HTTP QUERY requests.
+ *
+ * @see HttpMethod
+ * @since 5.0.0
```
jamezp, line 29: Small nit, this should be `@since 5.0` and we should leave the patch version off.

jamezp (changes requested):

HeroxHeruxum:
This feature would really improve the readability of my Code Base. So, for what its worth, I'd give a +1 ;)

mkarg:
> This feature would really improve the readability of my Code Base. So, for what its worth, I'd give a +1 ;)

I doubt this, as it is only few lines. 😉 

Note that only committers can vote in this project.

hantsy:
>Note that only committers can vote in this project.

I hope the experts can listen the voice from the community.

hantsy:
> Currently, developers are forced to use non-idempotent `POST` requests as a workaround for these complex searches. If we leave `QUERY` to third-party extensions, we risk fragmentation across the Java ecosystem where every framework writes their own `@QUERY` annotation. Adding it to the foundational API gives developers a standardized, idempotent alternative right out of the box that solves a very real developer pain point.
> 


@jamezp I couldn't agree more on this. The complex search parameters were the real problem when using `GET` in the real world development. The new `QUERY` is a good match for the use case.

HeroxHeruxum:
> > This feature would really improve the readability of my Code Base. So, for what its worth, I'd give a +1 ;)
> 
> I doubt this, as it is only few lines. 😉
> 
> Note that only committers can vote in this project.

It would make my resources so much more readable. I often just missuse Post requests, for complex querries, but the Operation ist read only. And that just hides intend.

And as my Humble non-committer opinnion, I'd say that The HTTPMethods class, should contain all Methods from the HTTP Specs 🤷

But, y'all got this ;)

But could you maybe re-iterate why it IS Not a No brauner to add it?

mkarg:
> > > This feature would really improve the readability of my Code Base. So, for what its worth, I'd give a +1 ;)
> > 
> > 
> > I doubt this, as it is only few lines. 😉
> > Note that only committers can vote in this project.
> 
> It would make my resources so much more readable. I often just missuse Post requests, for complex querries, but the Operation ist read only. And that just hides intend.
> 
> And as my Humble non-committer opinnion, I'd say that The HTTPMethods class, should contain all Methods from the HTTP Specs 🤷
> 
> But, y'all got this ;)
> 
> But could you maybe re-iterate why it IS Not a No brauner to add it?

If it is such a help for you, why not simply adding `@QUERY` in your own code? Why waiting for JAX-RS to come up with that?

mkarg:
> > And as my Humble non-committer opinnion, I'd say that The HTTPMethods class, should contain all Methods from the HTTP Specs 🤷

This is technically impossible as there is no clear fence what "the HTTP specs" is and what is not. HTTP is an open framework, and there are lots of methods, like WebDAV.

HeroxHeruxum:
> > > And as my Humble non-committer opinnion, I'd say that The HTTPMethods class, should contain all Methods from the HTTP Specs 🤷
> 
> This is technically impossible as there is no clear fence what "the HTTP specs" is and what is not. HTTP is an open framework, and there are lots of methods, like WebDAV.

The above comment with the Link to the RFC suggested that it already is considered standard and thus "spec", but If you read this https://www.rfc-editor.org/info/rfc10008/ you'll see it is only proposed as standard.

And with that knowledge, i would not add it anymore tbh.
At least not Here.

jeffque:
> > > > And as my Humble non-committer opinnion, I'd say that The HTTPMethods class, should contain all Methods from the HTTP Specs 🤷
> 
> > 
> 
> > This is technically impossible as there is no clear fence what "the HTTP specs" is and what is not. HTTP is an open framework, and there are lots of methods, like WebDAV.
> 
> 
> 
> The above comment with the Link to the RFC suggested that it already is considered standard and thus "spec", but If you read this https://www.rfc-editor.org/info/rfc10008/ you'll see it is only proposed as standard.
> 
> 
> 
> And with that knowledge, i would not add it anymore tbh.
> 
> At least not Here.

@HeroxHeruxum , according to [RFC 9110, section 9](https://www.rfc-editor.org/info/rfc9110/#section-9.1-7) : 

> This specification defines a number of standardized methods that are commonly used in HTTP, as outlined by the following table.

And the table mentions among others: `TRACE`, which is explicitly not in the list of annotations among many WebDAV methods.

Also, in other words by my interpretation: those methods are listed here just because they are commonly used, for convenience, not for some prior standardization via some committee.

So I belief that this argument of "it is not a standard" is a weak argument.

That is a opinion of a mere non committee opinion that is no better than anyone here.

So I share the same view as @mkarg expressed just a few comments before.

desiderantes:
> > > > And as my Humble non-committer opinnion, I'd say that The HTTPMethods class, should contain all Methods from the HTTP Specs 🤷
> > 
> > 
> > This is technically impossible as there is no clear fence what "the HTTP specs" is and what is not. HTTP is an open framework, and there are lots of methods, like WebDAV.
> 
> The above comment with the Link to the RFC suggested that it already is considered standard and thus "spec", but If you read this https://www.rfc-editor.org/info/rfc10008/ you'll see it is only proposed as standard.
> 
> And with that knowledge, i would not add it anymore tbh. At least not Here.

Under this logic, [HTTP 2](https://datatracker.ietf.org/doc/html/rfc9113) is also a proposed standard, perhaps people should be cautious of implementing it :) But more seriously, proposed means something different from what you have in mind.

Issue #1351, opened 2026-08-26
[OSGi] Provide the JakartaRESTfulWebServices osgi.contract capability


maho7791:
# jakarta.ws.rs-api does not provide the JakartaRESTfulWebServices osgi.contract capability

## What an osgi.contract is for

OSGi resolves dependencies between bundles by matching requirements against capabilities. For Java packages the usual mechanism is `Import-Package` with a version range, which works as long as the package version follows OSGi semantic versioning - the exporter then guarantees that everything inside a range such as `[3.0,4.0)` is compatible.

Specification APIs like this one do not follow that scheme: the package version is the specification version. The `osgi.contract` namespace exists for exactly this case. The bundle that exports the API declares a single capability naming the specification and listing every version it satisfies:

```
Provide-Capability: osgi.contract;osgi.contract=JavaServlet;
  version:List<Version>="2.5,3.0,3.1";
  uses:="javax.servlet,javax.servlet.http"
```

A consumer then imports the packages without any version and requires the contract instead:

```
Require-Capability: osgi.contract;filter:="(&(osgi.contract=JavaServlet)(version=2.5))"
Import-Package: javax.servlet,javax.servlet.http
```

The `uses` directive is what makes a versionless import safe: it forces the consumer to be wired to the packages of the bundle that declared the contract and to nothing else, so the class space stays consistent. The point of the mechanism is that the API bundle - the only party that actually knows which specification versions it satisfies - says so once, instead of every consumer having to guess a range. 

## The actual problem

In OSGi, a bundle that uses the Jakarta RESTful Web Services API has two ways to express  that dependency: import the specification packages with a version range, or import them without a range and require the `osgi.contract` capability that names the specification.

Today only the first option exists for this API, and it does not work well, because the `jakarta.ws.rs` packages are not versioned according to OSGi semantic versioning - their version is simply the specification version. Every consumer therefore has to guess a range, and the guess that build tools generate by default is `[3.0,4.0)`.

The consequence became visible with the 4.0 release: every existing bundle built against  3.x declares `jakarta.ws.rs;version="[3.0,4.0)"` and does not resolve against `jakarta.ws.rs-api 4.0.0` or `jakarta.ws.rs-api 5.0.0` (soon) at all - although 4.0 is compatible with 3.1 for everything except two types (4.0 removed `Link.JaxbAdapter` and `Link.JaxbLink`). 

## Proposed change

[...]

PR #1352, opened 2026-08-26
Add OSGi contract metadata for API Manifest version 5

maho7791:
Implements what #1351 proposes, for `main`. A backport to the `4.0.0` branch follows as a separate PR, so #1351 can stay open until both are in. The motivation is in the issue; this description only covers what the change does to the build and to the manifest.

### What it adds

Two instructions in the `maven-bundle-plugin` block of the root `pom.xml`:

```xml
<_contract>!*</_contract>
<Provide-Capability>osgi.contract;osgi.contract=JakartaRESTfulWebServices;version:List&lt;Version&gt;="5.0,4.0,3.1,3.0";uses:="jakarta.ws.rs,jakarta.ws.rs.client,jakarta.ws.rs.container,jakarta.ws.rs.core,jakarta.ws.rs.ext,jakarta.ws.rs.sse"</Provide-Capability>
```

The capability lets OSGi consumers depend on "the Jakarta RESTful Web Services API, version X" instead of guessing a package version range. It is metadata only: no code, no additional dependency, and no effect on users who do not run in OSGi.


pom.xml
```diff
@@ -237,7 +237,9 @@
                             <specversion>${spec.version}</specversion>
                             <_versionpolicy>[$(version;==;$(@)),$(version;+;$(@)))</_versionpolicy>
                             <_nodefaultversion>false</_nodefaultversion>
+                            <_contract>!*</_contract>
                             <Require-Capability>osgi.ee;filter:="(&amp;(osgi.ee=JavaSE)(version=21))"</Require-Capability>
+                            <Provide-Capability>osgi.contract;osgi.contract=JakartaRESTfulWebServices;version:List&lt;Version&gt;="5.0,4.0,3.1,3.0";uses:="jakarta.ws.rs,jakarta.ws.rs.client,jakarta.ws.rs.container,jakarta.ws.rs.core,jakarta.ws.rs.ext,jakarta.ws.rs.sse"</Provide-Capability>
```
jamezp, line 242: Please excuse my ignorance on OSGi :) I read the comment on the version list from the issue:
> On the version list. 4.0 removed `Link.JaxbAdapter` and `Link.JaxbLink`, so it is not a strict superset of 3.x. They are still listed, because in OSGi those two types were never usable: the JAXB packages they need were only ever reachable through `DynamicImport-Package`. The list has to be extended with every new specification version.

I'm not sure I understand. Those types are no longer part of the API so I don't follow thy the older versions need to be referenced.

maho7791: Fair question — the list isn't about which types are in the jar, it's the set of contract/spec versions this jar can satisfy. A consumer built against 3.1 carries Require-Capability: osgi.contract;filter:="(&(osgi.contract=JakartaRESTfulWebServices)(version=3.1))", and the resolver matches that literally. Without 3.1 in the list it won't resolve against the 5.0 API, even though the 5.0 API is a fine drop-in.

On Link.JaxbAdapter/JaxbLink: right, so 5.0 isn't a strict superset on paper. But in OSGi those two were never usable anyway — they need jakarta.xml.bind.annotation.adapters, which the 3.0/3.1 API bundles never imported (their manifests only carry DynamicImport-Package: *), so touching them was a NoClassDefFoundError unless something else happened to export that package. No OSGi consumer can have relied on them, so the 3.0/3.1 entries can't break anyone.

If you'd rather be strict, "5.0,4.0" also works. I'd keep them, but it's a one-line change either way.

jamezp (approved):
I think this is okay. My only concern was the version range, but I'm not an OSGi expert, so I'll go with what the experts think :)

